PR #3, opened 2022-04-04
add self-node-remediation-operator label to all resources and selectors

shellykatz:


beekhof:
odd that the e2e tests are failing though

PR #6, opened 2022-04-05
update operator-sdk version

mshitrit:


Makefile
```diff
@@ -192,7 +192,7 @@ e2e-test:
 	go test ./e2e -ginkgo.v -ginkgo.progress -test.v -timeout 60m -count=1
 
 .PHONY: operator-sdk
-OPERATOR_SDK_VERSION = v1.17.0
+OPERATOR_SDK_VERSION = v1.18.0
```
razo7, line 195: Why in the Makefile we are using 1.18 version and in the bundle files we use 1.17 version?

mshitrit: That's just how it's generated - defiantly confusing (I'm pretty sure we discussed it at the previous sdk-operator upgrade as well :smile:  )
`operator-sdk-v1.17.0+` is equivalent to `v1.18.0`

razo7: That's true :) https://github.com/medik8s/poison-pill/pull/141/files

PR #7, opened 2022-04-07
Update ppr to snr

mshitrit:
changing ppr entries to snr.
This is split into 3 commits: 
- Entry changes in logs
- Entry changes in comments & test code naming
- Entry changes in variable and method names


slintes (changes requested):
too many spaces around 😉

PR #9, opened 2022-04-10
Webhook validation snrc time fields

shellykatz:
Webhook validation for validating **update** of the SelfNodeRemediationConfig CR. The validation checks that the time values in the time fields aren't smaller than a decided minimal value.


api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (r *SelfNodeRemediationConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
```
mshitrit, line 92: Nit: clean up TODOs

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,166 @@
...
+	"os"
+	"path/filepath"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"time"
+)
+
+const (
+	WebhookCertDir  = "/apiserver.local.config/certificates"
```
mshitrit, line 32: For all consts that are not used outside of package start with  lower case

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,166 @@
...
+func (r *SelfNodeRemediationConfig) ValidateDelete() error {
+	selfnoderemediationconfiglog.Info("validate delete", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object deletion.
+	return nil
+}
+
+// ValidateTimes validates each time field in the PoisonPillConfig CR doesn't go below the minimum time
+// that was defined to it
+func (r *SelfNodeRemediationConfig) ValidateTimes() error {
```
mshitrit, line 123: use lower case for method prefix

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,166 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func checkField(inputTime int64, minValidDur string, errMessage string) error {
+	minValidDurMS, err := toMS(minValidDur)
+	if err != nil {
```
mshitrit, line 147: Probably would be good to add log message either here or inside toMS

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,166 @@
...
+			break
+		}
+	}
+	if certsInjected {
+		server := mgr.GetWebhookServer()
+		server.CertDir = WebhookCertDir
+		server.CertName = WebhookCertName
+		server.KeyName = WebhookKeyName
+	} else {
+		selfnoderemediationconfiglog.Info("OLM injected certs for webhooks not found")
```
mshitrit, line 82: use camelCase for variable names

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -0,0 +1,141 @@
...
+	fieldValue    string
+	fieldMinValue string
+	errorMessage  string
+}
+
+var _ = Describe("SelfNodeRemediationConfig Validation", func() {
+
+	Describe("validate update time fields of SelfNodeRemediationConfig CR", func() {
+		shortTimeTestItems := []TestDataItem{
+			{"PeerApiServerTimeout", "1.2ms", MinDurPeerApiServerTimeout, ErrPeerApiServerTimeout},
```
mshitrit, line 21: Nit: I suggest using more diverse values instead of 1.2 for all the tests

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -0,0 +1,141 @@
...
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring(item.errorMessage))
+				})
+			})
+		}
+
+		Context("for valid CR", func() {
+			It("should not be rejected", func() {
+				snrcOld := createDefaultSelfNodeRemediationConfigCR()
+				snrc := createDefaultSelfNodeRemediationConfigCR()
```
mshitrit, line 70: Nit: I think it's better to use more diverse value set - at least some values should be different from original values

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	//set the field tested
+	setFieldValue(snrc, fieldName, value)
+
+	return snrc
+}
+
+func setFieldValue(snrc *SelfNodeRemediationConfig, fieldName string, value string) {
+	d, err := time.ParseDuration(value)
+	if err != nil {
```
mshitrit, line 122: replace with d, _ if you choose to ignore the error (I think it can be acceptable in some case of test  code).

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+					err := snrc.ValidateUpdate(snrcOld)
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring(item.errorMessage))
+				})
+			})
+		}
+
+		negativeTimeTestItems := []TestDataItem{
+			{"PeerApiServerTimeout", "-1ms", MinDurPeerApiServerTimeout, ErrPeerApiServerTimeout},
```
mshitrit, line 45: I recommend replacing all of those strings with constants - which can be used both here and in the production code

config/default/manager_webhook_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        ports:
+        - containerPort: 9443
+          name: webhook-server
+          protocol: TCP
+#        volumeMounts:
```
mshitrit, line 15: Probably cleanup if not needed

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
razo7: And maybe use a const variable instead of ` 1.2ms`.

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,171 @@
...
+
+//+kubebuilder:webhook:path=/validate-self-node-remediation-medik8s-io-v1alpha1-selfnoderemediationconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=self-node-remediation.medik8s.io,resources=selfnoderemediationconfigs,verbs=create;update,versions=v1alpha1,name=vselfnoderemediationconfig.kb.io,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Validator = &SelfNodeRemediationConfig{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *SelfNodeRemediationConfig) ValidateCreate() error {
+	selfNodeRemediationConfigLog.Info("validate create", "name", r.Name)
+
+	return nil
```
slintes, line 108: why do you check updates only, but no creates?

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,171 @@
...
+		err := checkField(f.fieldDurValue.Milliseconds(), f.fieldMinValueString, f.errorMessage)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func checkField(inputTime int64, minValidDur string, errMessage string) error {
```
slintes, line 149: You could make `checkField` a method of the `field` type, so you don't need to pass the arguments.
Maybe field.validate() would be a nice name then.

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -0,0 +1,171 @@
...
+	ApiServerTimeout     = "ApiServerTimeout"
+	PeerDialTimeout      = "PeerDialTimeout"
+	PeerRequestTimeout   = "PeerRequestTimeout"
+	ApiCheckInterval     = "ApiCheckInterval"
+	PeerUpdateInterval   = "PeerUpdateInterval"
+)
+
+//minimal time durations allowed string format
+const (
+	MinDurPeerApiServerTimeout = "10ms"
```
slintes, line 49: What about using `Duration` types here? Then you can skip the `toMS` conversion and directly compare with the config value (without `MilliSeconds()`)?

bundle/manifests/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -412,3 +416,24 @@ spec:
...
+    deploymentName: self-node-remediation-controller-manager
+    failurePolicy: Fail
+    generateName: vselfnoderemediationconfig.kb.io
+    rules:
+    - apiGroups:
+      - self-node-remediation.medik8s.io
+      apiVersions:
+      - v1alpha1
+      operations:
+      - UPDATE
```
slintes, line 433: we should have CREATE here as well?

config/webhook/kustomization.yaml
```diff
@@ -0,0 +1,24 @@
...
+      name: validating-webhook-configuration
+    webhooks:
+      - name: vselfnoderemediationconfig.kb.io
+        rules:
+        - apiGroups:
+          - self-node-remediation.medik8s.io
+          apiVersions:
+          - v1alpha1
+          operations:
+          - UPDATE
```
slintes, line 22: CREATE?

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,29 @@
...
+      path: /validate-self-node-remediation-medik8s-io-v1alpha1-selfnoderemediationconfig
+  failurePolicy: Fail
+  name: vselfnoderemediationconfig.kb.io
+  rules:
+  - apiGroups:
+    - self-node-remediation.medik8s.io
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
```
slintes, line 25: here it is :)

slintes (commented):
left some comments, didn't review the tests yet though

api/v1alpha1/selfnoderemediationconfig_webhook.go
shellykatz: I remember reading in the medik8s.io that users can't create the PPC CR, only edit it. So I figured only validating the update is a better way to prevent the problem that used to happen when the PPC object was created before the webhook server is started (now it's not a problem because the creation of the default PPC isn't validated). WDYT?

api/v1alpha1/selfnoderemediationconfig_webhook.go
shellykatz: I thought about it but I also use this const in logging the error. If I used `Duration` types I would still need these consts for the err message (since in duration type it's all in milliseconds and I not as clear for the user I guess).
So I wonder is it better to have more constants and not use the toMS OR use the toMS and have less constants? @slintes

api/v1alpha1/selfnoderemediationconfig_webhook.go
slintes: Users can delete and recreate the config at any time. And also validating our own default config has value IMHO.

api/v1alpha1/selfnoderemediationconfig_webhook.go
slintes: you can format to whatever you need for the error message, e.g.
```
	test := 2000 * time.Millisecond
	errMsg := fmt.Sprintf("time in s: %vs", test.Seconds())
	fmt.Println(errMsg)
```
prints `time in s: 2s`

main.go
```diff
@@ -19,12 +19,13 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
-	"github.com/medik8s/self-node-remediation/pkg/utils"
+	"github.com/medik8s/self-node-remediation/pkg/snrconfighelper"
```
slintes, line 22: please move this to the end of the imports

slintes, line 22: please move this to the end of the imports

main.go
```diff
@@ -19,12 +19,13 @@ package main
...
 	"os"
 	"strconv"
 	"time"
 
+	"github.com/medik8s/self-node-remediation/pkg/utils"
```
slintes, line 27: ditto

slintes, line 27: ditto

pkg/snrconfighelper/configinitializer.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/go-logr/logr"
+	selfnoderemediationv1alpha1 "github.com/medik8s/self-node-remediation/api/v1alpha1"
+	"github.com/medik8s/self-node-remediation/pkg/utils"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ConfigInitializer struct {
+	//defaultConfig poisonpillv1alpha1.PoisonPillConfig
```
slintes, line 14: please remove unused line

slintes, line 14: please remove unused line

slintes (commented):
Looks good! Just some smaller nits.
We need to check the e2e test later on, I understand the CI cluster is being upgraded atm and has some trouble, so atm I can't see why the test failed.

edit: I see my comments duplicated... but I don't dare to delete the duplicates, I'm afraid I will delete both 😉

PR #10, opened 2022-04-11
Update CSV - OCP Annotation & Description

razo7:
- Remove OCP annotation (disconnected)
- Adding description


bundle/manifests/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -411,4 +412,4 @@ spec:
   provider:
     name: medik8s
     url: https://www.medik8s.io/
-  version: 0.4.0
+  version: 0.4.0
```
slintes, line 415: where did the last line break go? 🤔

PR #13, opened 2022-04-17
Updating Github action versions + post PR functionality

razo7:
- Updating GitHub Action versions (similar to [NMO PR](https://github.com/medik8s/node-maintenance-operator/pull/24) and [NHC PR](https://github.com/medik8s/node-healthcheck-operator/pull/98)):
  - [actions/checkout](https://github.com/actions/checkout/tree/main) and  [actions/setup-go](https://github.com/actions/setup-go/tree/main) versions from 2 to 3.
  - [docker/login-action](https://github.com/docker/login-action) from version 1 to 2.
- Using `IMAGE_TAG` instead of `VERSION`.
- Adding _bundle_ and _catalog_ `build` and `push` and two new targets (`container-build` & `container-push`) for CI in Makefile.

### DIFF
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -12,17 +12,17 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: 1.18
 
       - name: Log in to Quay.io
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   push_to_registry:
-    name: Push a container image to quay.io/medik8s
+    name: Build and push images to quay.io/medik8s
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
@@ -30,9 +30,9 @@ jobs:
 
       - name: Build and push operator and bundle images, using version 0.0.1, for pushes to main
         if: ${{ github.ref_type != 'tag' }}
-        run: export VERSION=0.0.1 && make docker-build bundle bundle-build docker-push bundle-push
+        run: make container-build container-push
 
-      - name: Build and push versioned operator and bundle images, for tags
+      - name: Build and push versioned CSV and images, for tags
         if: ${{ github.ref_type == 'tag' }}
         # remove leading 'v' from tag!
-        run: export VERSION=$(echo $GITHUB_REF_NAME | sed 's/v//') && make docker-build bundle bundle-build docker-push bundle-push
+        run: export VERSION=$(echo $GITHUB_REF_NAME | sed 's/v//') && make container-build container-push
--- a/.github/workflows/pre-submit.yaml
+++ b/.github/workflows/pre-submit.yaml
@@ -13,12 +13,12 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
     - name: Set up Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: 1.18
 
--- a/.github/workflows/pre-submit.yaml
+++ b/.github/workflows/pre-submit.yaml
@@ -1,5 +1,5 @@
-name: Go
-on:
+name: Pre Submit # workflow name
+on: # on events
   push:
     branches:
       - main
@@ -22,8 +22,11 @@ jobs:
       with:
         go-version: 1.18
 
-    - name: Build and test
-      run: make manifests generate docker-build
+    - name: Run checks and unit tests
+      run: make test
+
+    - name: Test container build
+      run: make container-build
 
     - name: TestMutations
       run: make test-mutation-ci
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,9 @@ IMAGE_TAG_BASE ?= $(IMAGE_REGISTRY)/$(OPERATOR_NAME)
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-operator-bundle:$(IMAGE_TAG)
 
+# The image tag given to the resulting catalog image (e.g. make catalog-build CATALOG_IMG=example.com/operator-catalog:v0.2.0).
+CATALOG_IMG ?= $(IMAGE_TAG_BASE)-operator-catalog:$(IMAGE_TAG)
+
 # Image URL to use all building/pushing image targets
 export IMG ?= $(IMAGE_TAG_BASE)-operator:$(IMAGE_TAG)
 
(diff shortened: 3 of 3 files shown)


.github/workflows/post-submit.yaml
```diff
@@ -3,32 +3,31 @@ on:
...
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
           registry: quay.io
 
-      - name: Prepare Container Image
-        run: make docker-build
+      - name: Build and push CSV 0.0.1 + latest images, for pushes
+        if: ${{ github.ref_type != 'tag' }}
+        run: make docker-build bundle bundle-build docker-push bundle-push index-build index-push
```
slintes, line 28: bundle-push index-build index-push don't exist in the Makefile

.github/workflows/post-submit.yaml
```diff
@@ -3,32 +3,31 @@ on:
...
       - name: Log in to Quay.io
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
           registry: quay.io
 
-      - name: Prepare Container Image
-        run: make docker-build
+      - name: Build and push CSV 0.0.1 + latest images, for pushes
```
slintes, line 26: I don't see where that 0.0.1 will be set

slintes (changes requested):

.github/workflows/post-submit.yaml
razo7: IIUC it is now set in Makefile,

Makefile
```diff
@@ -228,4 +254,33 @@ ifeq (,$(wildcard $(OPERATOR_SDK)))
...
+
+.PHONY: catalog-push
+catalog-push: ## Push a catalog image.
+	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+
+##@ Targets used by CI
+
+.PHONY: container-build
+container-build: ## Build containers
+	make docker-build bundle-build
```
slintes, line 278: please add `bundle` target before `bundle-build`, so manifests get updated with $VERSION etc.

Makefile
```diff
@@ -228,4 +254,33 @@ ifeq (,$(wildcard $(OPERATOR_SDK)))
...
+
+.PHONY: container-push
+container-push: ## Push containers (NOTE: catalog can't be build before bundle was pushed)
+	docker-push bundle-push catalog-build catalog-push
```
slintes, line 282: missing `make`

slintes (changes requested):

slintes:
needs rebase because of https://github.com/medik8s/self-node-remediation/pull/29/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R40

slintes:
lgtm

@mshitrit do you want to have a look as well?

CI seems to be broken currently , probably because of quay.io maintenance

.github/workflows/post-submit.yaml
```diff
@@ -8,31 +8,31 @@ on:
 
 jobs:
   push_to_registry:
-    name: Push a container image to quay.io/medik8s
+    name: Build and push images to quay.io/medik8s
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
mshitrit, line 15: I see in couple of places upgrade of the version.
Can I take it the the github action itself are versioned ?

razo7: > Can I take it the the github action itself are versioned ?

Yes, see https://github.com/actions/checkout/releases

Makefile
```diff
@@ -110,9 +137,6 @@ test-mutation: verify-no-changes fetch-mutation ## Run mutation tests in manual
 	echo -e "## Verifying diff ## \n##Mutations tests actually changes the code while running - this is a safeguard in order to be able to easily revert mutation tests changes (in case mutation tests have not completed properly)##"
 	./hack/test-mutation.sh
 
-test-mutation-ci: fetch-mutation ## Run mutation tests as part of auto build process.
```
mshitrit, line 113: For future reference, if possible please try to keep the same location in the code - it makes the review process a lot easier

.github/workflows/post-submit.yaml
```diff
@@ -8,31 +8,31 @@ on:
...
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
           registry: quay.io
 
       - name: Build and push operator and bundle images, using version 0.0.1, for pushes to main
         if: ${{ github.ref_type != 'tag' }}
-        run: export VERSION=0.0.1 && make docker-build bundle bundle-build docker-push bundle-push
+        run: make container-build container-push
```
mshitrit, line 33: As far as I see this is not logically equivalent, can you explain the motivation ?

.github/workflows/post-submit.yaml
```diff
@@ -8,31 +8,31 @@ on:
...
 
-      - name: Build and push versioned operator and bundle images, for tags
+      - name: Build and push versioned CSV and images, for tags
         if: ${{ github.ref_type == 'tag' }}
         # remove leading 'v' from tag!
-        run: export VERSION=$(echo $GITHUB_REF_NAME | sed 's/v//') && make docker-build bundle bundle-build docker-push bundle-push
+        run: export VERSION=$(echo $GITHUB_REF_NAME | sed 's/v//') && make container-build container-push
```
mshitrit, line 38: Same question as before (also not logically equivelent)

.github/workflows/pre-submit.yaml
```diff
@@ -13,17 +13,20 @@ jobs:
...
     - name: Set up Go
-      uses: actions/setup-go@v2
+      uses: actions/setup-go@v3
       with:
         go-version: 1.18
 
-    - name: Build and test
-      run: make manifests generate docker-build
+    - name: Run checks and unit tests
+      run: make test
```
mshitrit, line 26: IIUC test is redundant here.
It also runs as part of _docker-build_ which is run as part of _container-build_

.github/workflows/post-submit.yaml
razo7: Previously we have built and push the operator and bundle containers, and with this change we also build and push the index/catalog image. Why? It can be used for testing afterwards, and it also gives us indication whether the creation index/catalog image is successful.

.github/workflows/pre-submit.yaml
```diff
@@ -13,17 +13,20 @@ jobs:
...
+
+    - name: Test container build
+      run: make container-build
```
mshitrit, line 29: This is not logically equivalent to previous code, can you explain the motivation ?

Makefile
```diff
@@ -1,17 +1,32 @@
...
 # VERSION defines the project version for the bundle. 
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-export VERSION ?= $(shell git describe --match="v*"| sed 's/v//')
-
-# use stable channel
-CHANNELS = stable
+VERSION ?= $(DEFAULT_VERSION)
```
mshitrit, line 28: Is there a use case, when VERSION is different than DEFAULT_VERSION ?

.github/workflows/pre-submit.yaml
razo7: In NHC and NMO we run `test` twice -  as a step before we try to build the containers and with _container-build_ , but with `test-no-verify`.
Why have we done that? IIUC since in container-build we have been using test-no-verify we wanted to run `test` beforehand while in SNR it is not the case as it doesn't have test-no-verify.
Therefore, it is indeed redundant

.github/workflows/pre-submit.yaml
razo7: We still have `make manifests generate docker-build` in the new code.
In container-build  there is `test`, which includes `manifests generate`, and afterwards we have `docker-build`.
The new thing is adding the creation of  the bundle directory (`bundle` target), and build of the bundle image before PR is merged for checking if building the bundle image was successful.

.github/workflows/post-submit.yaml
razo7: > As far as I see this is not logically equivalent,

That's correct, `export VERSION=0.0.1 &&` is not needed after the changes I am suggesting in Makefile

> container-push also runs catalog-build & catalog-push

I replied on the next comment.

razo7:
Oops, SNR doesn't have `opm` for building the catalog image, and PR #31 will try to fix that.

PR #14, opened 2022-05-02
Fix webhook validation error message

shellykatz:
Fixed the selfNodeRemediationConfig webhook validation to contain all fields with wrong values (instead of only the first one that was found wrong)

### DIFF
--- a/api/v1alpha1/selfnoderemediationconfig_webhook_test.go
+++ b/api/v1alpha1/selfnoderemediationconfig_webhook_test.go
@@ -103,7 +103,6 @@ func testMultipleInvalidFields(validationType string) {
 		setFieldValue(snrc, item.name, item.durationValue)
 		errorMsg += "\n" + item.name + " cannot be less than " + item.minDurationValue.String()
 	}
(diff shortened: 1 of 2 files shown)


api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -77,6 +94,33 @@ func testInvalidCRs(validationType string) {
...
+func testMultipleInvalidFields(validationType string) {
+	var errorMsg string
+	snrc := createDefaultSelfNodeRemediationConfigCR()
+
+	for _, item := range testItems2 {
+		item := item
+		setFieldValue(snrc, item.name, item.durationValue)
+		errorMsg += "\n" + item.name + " cannot be less than " + item.minDurationValue.String()
+	}
+	println(errorMsg)
```
razo7, line 106: Is that necessary? I assume it was merely for debugging.

shellykatz: good catch! thanks

PR #15, opened 2022-05-03
capture last error message from reconcile in snr CR status

shellykatz:
add LastError field to snr status. add function to update this field in the controller and return the error

### DIFF
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -62,6 +62,11 @@ type SelfNodeRemediationStatus struct {
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=status
 	Phase *string `json:"phase,omitempty"`
+
+	// LastError captures the last error that occurred during remediation.
+	// If no error occurred it would be nil
+	//+operator-sdk:csv:customresourcedefinitions:type=status
+	LastError string `json:"lastError,omitempty"`
 }
 
 //+kubebuilder:object:root=true
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -64,7 +64,7 @@ type SelfNodeRemediationStatus struct {
 	Phase *string `json:"phase,omitempty"`
 
 	// LastError captures the last error that occurred during remediation.
-	// If no error occurred it would be nil
+	// If no error occurred it would be empty
 	//+operator-sdk:csv:customresourcedefinitions:type=status
 	LastError string `json:"lastError,omitempty"`
 }
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -651,6 +651,7 @@ func (r *SelfNodeRemediationReconciler) updateSnrStatusLastError(snr *v1alpha1.S
 		updateErr := r.Client.Status().Patch(context.Background(), snr, patch)
 		if updateErr != nil {
 			r.logger.Error(updateErr, "Failed to update SelfNodeRemediation status")
+			return updateErr
 		}
 	}
 
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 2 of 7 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -53,6 +54,8 @@ var (
 
 	lastSeenSnrNamespace  string
 	wasLastSeenSnrMachine bool
+
+	untreatableError = errors.New("untreatable error, stop reconciling")
```
mshitrit, line 58: Nit: maybe unreconcilable instead of untreatable ?

controllers/selfnoderemediation_controller.go
```diff
@@ -624,3 +630,28 @@ func (r *SelfNodeRemediationReconciler) restoreNode(nodeToRestore *v1.Node) (ctr
...
 	return ctrl.Result{Requeue: true}, nil
 }
+
+func (r *SelfNodeRemediationReconciler) updateSnrStatusLastError(snr *v1alpha1.SelfNodeRemediation, err error) error {
+	var lastErrorVal string
+
+	if err != nil {
+		lastErrorVal = err.Error()
+	} else {
+		lastErrorVal = "successful reconcile"
```
mshitrit, line 640: I think setting `lastErrorVal` to "successful reconcile" is a bit confusing.
I'd suggest something like:
```
If err == nil {
  return clearSnrStatusLastError()
 }
```

controllers/selfnoderemediation_controller.go
slintes, line 58: I think I would prefer a custom error type, instead of relying on unwrapping exactly once to find this error message

controllers/selfnoderemediation_controller.go
slintes, line 640: According to the field description in the CRD it should be empty when no error occured. Which makes more sense to me than this message.

controllers/selfnoderemediation_controller.go
```diff
@@ -624,3 +630,28 @@ func (r *SelfNodeRemediationReconciler) restoreNode(nodeToRestore *v1.Node) (ctr
...
+	}
+
+	if snr.Status.LastError != lastErrorVal {
+		snr.Status.LastError = lastErrorVal
+		updateErr := r.Client.Status().Update(context.Background(), snr)
+		if updateErr != nil {
+			r.logger.Error(updateErr, "Failed to update SelfNodeRemediation status")
```
slintes, line 647: We don't want to retry when this fails?
Also, because every PP agent is potentially trying to update the CR, I think that there is a good chance that `Update` fails. I think `Patch` would be better here.

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -62,6 +62,11 @@ type SelfNodeRemediationStatus struct {
 	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=status
 	Phase *string `json:"phase,omitempty"`
+
+	// LastError captures the last error that occurred during remediation.
+	// If no error occurred it would be nil
```
slintes, line 67: s/nil/empty

mshitrit:
looks good!
I don't think any of my comments are blocking, so feel free to unhold (just want to give a chance for changes/more reviews).
/hold
/lgtm

slintes (changes requested):
Probably out of scoper for this PR, but we have multiple placs where we update the status. I think it should be patch instead of update, and at one place only, to save API calls and lower the chance of conflicts.
Some more comments inline.

controllers/selfnoderemediation_controller.go
shellykatz: will change to `Patch` :)
About retrying when this fails, I asked Michael and we decided not to because we thought we don't want to add another reason for the reconcile to fail (IIUC). Wdyt?  @mshitrit

mshitrit: IIRC we thought it does not make sense to return an error for fail  because it'll overshadow the original error.
However Marc brings up a good point - I think we can retry without returning an error that will overshadow the original error.
However since we are a bit short on time I suggest to proceed as is and just open a Jira ticket to capture this.

slintes: Sorry, saw the comments only just now.
I don't think we overshadow anything.
We just reconcile again. We run into the same error. Then we try to update the status again. Until it succeeds.

mshitrit: >I don't think we overshadow anything.

My intention is that now instead of returning the initial error that caused the malfunction, we will return an error that say that we couldn't updated the status.

Issue #17, opened 2022-05-06
[RFC] Support for out-of-service taint


k-keiichi-rh:
https://github.com/kubernetes/enhancements/pull/1116 is merged into k8s 1.24 as a beta feature.
And it will become a stable feature in k8s 1.26.

This KEP introduces a new "out-of-service" taint that ensures pods on the node will be forcefully
deleted and the volume detach operations on the node will happen immediately.
It means the deleted pods can recover quickly on different nodes.

This feature will have a good impact on the existing remediation mechanisms.
But there is a restriction that the user has to confirm the node is in 'NotReady' status
and the node is shutdown or in a non-recoverable state.

If the above conditions are not satisfied, pods on the unhealthy node are not deleted
and the volumeAttachment objects are not also removed.

There are several options to confirm the above conditions in SNR:

1. Fencing the node by using any management interface (e.g. IPMI etc.)
   => But SNR doesn't have the prerequisite according to https://www.medik8s.io/remediation/poison-pill/faq/#how-poison-pill-is-different-from-other-solutions
2. Intentional kernel panic
   => softdog supports the "soft_panic" option.
3. Soft power off
   => We can't trust soft power off when the node is unhealthy.
4. Other options?

Any comments are very welcome.
And if my understanding is wrong, please correct me.

For reference, the following flow is how NHC and SNR work:
```
1. When NHC detects a node failure, it creates a PoisonPillRemediation(PPR) object for the remediation.
2. SNR starts to remediate the node according to the PPR object.
3. SNR marks the unhealthy node as "SchedulingDisabled"
4. The unhealthy node reboots(*1)
   => We have three tiers(1. hardware watchdog, 2. softdog, 3. software reboot).
6. SNR waits for 180 seconds(default) before it starts deleting affected workloads like pods and volumeAttachment objects.
7. SNR takes a action according to the PPR object. We have "NodeDeletion" and "ResourceDeletion" as remediation
8. SNR deletes the Node Object(in case of the "NodeDeletion") or free the affected resources(in case of "ResourceDeletion").
9. After recovering the node, SNR unmarks the "SchedulingDisabled" flag from the unhealthy node.
   SNR restores the deleted Node Object(in case of the "NodeDeletion")

(*1)
SNR on a node will trigger reboot if
  1. it sees a remediation CR for itself
  2. or it can't connect to the API server AND
   2a) at least one peer SNR reports a remediation CR exists
[...]

PR #20, opened 2022-05-11
Add node Taint remediation

mshitrit:
This PR is the first attempt to implementing NEC requirments defined in  [this](https://issues.redhat.com/browse/ECOPROJECT-777) ticket.

**Implementation high level details**
 
 Added a new remediation strategy "Add Taint" (the following is relevant for this strategy):
1. Upon CR detection set NoExucte taint on the Node & add finalizer to the CR
2. This should trigger the actual remediation (implemented in the API Server)
3. When remediation is successful NHC will try to delete the CR
4. When CR is attempted to be deleted, remove the finalizer from the CR & the taint from the Node

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 	"sync"
 	"time"
 
@@ -49,6 +50,11 @@ var (
 	NodeUnschedulableTaint = &v1.Taint{
 		Key:    "node.kubernetes.io/unschedulable",
 		Effect: v1.TaintEffectNoSchedule,
+	}
+
+	NodeNoExecuteTaint = &v1.Taint{
+		Key:    "node.kubernetes.io/out-of-service",
+		Effect: v1.TaintEffectNoExecute,
 	}
 
 	lastSeenSnrNamespace  string
@@ -142,6 +148,8 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		result, err = r.remediateWithNodeDeletion(snr)
 	case v1alpha1.ResourceDeletionRemediationStrategy:
 		result, err = r.remediateWithResourceDeletion(snr)
+	case v1alpha1.AddTaintRemediationStrategy:
+		result, err = r.remediateByAddingTaint(snr)
 	default:
 		//this should never happen since we enforce valid values with kubebuilder
 		err := errors.New("unsupported remediation strategy")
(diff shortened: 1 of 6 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -664,3 +701,37 @@ func (r *SelfNodeRemediationReconciler) updateSnrStatusLastError(snr *v1alpha1.S
...
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *SelfNodeRemediationReconciler) removeOutOfServiceTaint(node *v1.Node) error {
+	patch := client.MergeFrom(node.DeepCopy())
+	var taintIndex int
+	taintFound := false
+	for index, taint := range node.Spec.Taints {
+		if reflect.DeepEqual(taint, *NodeNoExecuteTaint) {
```
slintes, line 720: maybe use k8s.io/api/core/v1/taint.go MatchTaint

slintes (commented):
- This feature should be documented somewhere, especially noting when it works and when not (k8s/ocp version, feature status, feature gates...)
- More important: check the note about the taint here: https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown
> Before adding the taint node.kubernetes.io/out-of-service , it should be verified that the node is already in shutdown or power off state (not in the middle of restarting).

At the moment shutdown is completely missing in the implementation. And "not in the middle of restarting" does not align with our reboot-and-guess-when-it's-done approach of other strategies. We need a safe way to shut the node down, or to detect that it aleady is. But then we can't switch on, which contradicts our goal of node recovery... 🤔 See also discussion in https://github.com/medik8s/self-node-remediation/issues/17

beekhof:
Closing the loop here, the expiry of the timeout implies shutdown - with a confidence that varies based on the users' environment and configuration (hardware watchdog, software watchdog, software reboot)

controllers/selfnoderemediation_controller.go
```diff
@@ -370,6 +378,35 @@ func (r *SelfNodeRemediationReconciler) remediateWithNodeDeletion(snr *v1alpha1.
...
+func (r *SelfNodeRemediationReconciler) remediateByAddingTaint(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
+	node, err := r.getNodeFromSnr(snr)
+	if err != nil {
+		r.logger.Error(err, "failed to get node", "node name", snr.Name)
+		return ctrl.Result{}, err
+	}
+
+	if !controllerutil.ContainsFinalizer(snr, SNRFinalizer) {
+		return r.addFinalizer(snr)
+	}
```
beekhof, line 390: Shouldn't that happen for all strategies before this function gets called?

beekhof: Even more concerning... 

```
        wasRebooted, timeLeft := r.wasNodeRebooted(snr)
	if !wasRebooted {
		return ctrl.Result{RequeueAfter: timeLeft}, nil
	}

```

**MUST** also be part of ***every*** strategy

mshitrit:
Closing this PR.
Implementation will be done in smaller segments (possibly by different teams), addressing the issues that were brought up.

PR #21, opened 2022-05-30
Add/Remove No Execute Taint

mshitrit:
- Adding a  `No Execute` taint in order to make sure after reboot the node is not running any workload - this is important so we can move this workload elsewhere even if the node does not have connectivity to the control plain.
- Once a node is healthy the taint is removed.

### DIFF
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -173,6 +173,8 @@ var _ = Describe("snr Controller", func() {
 
 				verifyFinalizerExists()
 
+				verifyNoExecuteTaintExist()
+
 				verifyNoWatchdogFood()
 
 				verifyNodeDeletedAndRestored(beforeSNR)
@@ -200,6 +202,8 @@ var _ = Describe("snr Controller", func() {
 				By("Verify that finalizer was removed and SNR can be deleted")
 				testNoFinalizer()
 
+				verifyNoExecuteTaintRemoved()
+
 			})
 		})
 
@@ -226,6 +230,8 @@ var _ = Describe("snr Controller", func() {
 
 				verifyFinalizerExists()
 
+				verifyNoExecuteTaintExist()
+
 				verifyNoWatchdogFood()
 
 				verifySelfNodeRemediationPodDoesntExist()
@@ -236,6 +242,8 @@ var _ = Describe("snr Controller", func() {
 
 				By("Verify that finalizer was removed and SNR can be deleted")
 				testNoFinalizer()
+
+				verifyNoExecuteTaintRemoved()
 
 			})
 		})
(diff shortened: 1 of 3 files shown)


beekhof:
Have you thought about putting it just before the snr.Spec.RemediationStrategy switch block?

slintes (changes requested):
- seems there is some best practice about naming taints, e.g. `node.kubernetes.io/out-of-service=nodeshutdown:NoExecute`. Maybe use `medik8s.io/remediation=nodefencing:NoExecute`
- can we have a e2e test for the taint please?

mshitrit:
> Maybe use `medik8s.io/remediation=nodefencing:NoExecute`

Sure :+1: 

> can we have a e2e test for the taint please?

Even though there is no specific e2e test for this use case my initial bug (not excluding the agent from the taint) was caught by our current e2e test so I'm not sure if adding another E2E is the way to go (I'd like to keep e2e tests to the necessary minimum).

In any case I'm planing to try to add a UT for this use case.
WDYT ?

slintes:
> > can we have a e2e test for the taint please?
> 
> Even though there is no specific e2e test for this use case my initial bug (not excluding the agent from the taint) was caught by our current e2e test

Sorry, but this is a bad argumeny IMHO. Current e2e catching this bug doesn't mean it catches other bugs, e.g. not setting the taint at all.   

> so I'm not sure if adding another E2E is the way to go (I'd like to keep e2e tests to the necessary minimum).
> 

IMHO this taint qualifies for "minimum". But if you think a unit test is good enough, I'm fine with it.

controllers/controller_test.go
```diff
@@ -354,6 +364,32 @@ func verifyFinalizerExists() {
...
+	By("Verify that node does not have NoExecute taint")
+	Eventually(isNoExecuteTaintExist(), 10*time.Second, 200*time.Millisecond).Should(BeFalse())
+}
+
+func verifyNoExecuteTaintExist() {
+	By("Verify that node has NoExecute taint")
+	Eventually(isNoExecuteTaintExist(), 10*time.Second, 200*time.Millisecond).Should(BeTrue())
+}
+
+func isNoExecuteTaintExist() func() bool {
```
slintes, line 377: this can be made easier: no need to return a new func(), as this already is a func. You just need to pass in this func just by name without () in the Eventually above:

```
... Eventually(isNoExecuteTaintExist,...

func isNoExecuteTaintExist() (bool, error) { // no new func here...}
```

slintes (changes requested):
just 2 comments about the test, otherwise lgtm

@beekhof do you agree with the taint name?

controllers/controller_test.go
beekhof: I might be inclined to have the value be the operator name, eg. `self-node-remediation`

slintes: so `medik8s.io/remediation=self-node-remediation:NoExecute` ?
Works for me 👍🏼

PR #30, opened 2022-06-17
Fix taint removal

slintes:
In manual tests and NHC CI the NoExecute taint isn't removed after remediation.

For debugging and fixing the issue this PR evolved over time and mainly contains these changes:
- add e2e test for the taint (didn't reproduce the issue)
- add taint after adding finalizer (didn't fix the issue)
- wait for removal of Unschedulable Taint after setting Unschedulable to false: this fixed the issue! It was a race between node-lifecycle-controller and us with removing taints

more changes:
- added logging
- fix running unit test when they break (prevent metrics port conflicts)
- use existing DeleteTaint func
- add TimeAdded to taint

### DIFF
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -278,8 +278,6 @@ var _ = Describe("snr Controller", func() {
 				lastFoodTime = newTime
 				return false
 			}, 10*peerUpdateInterval, timeout).Should(BeTrue())
-
-			verifyNoExecuteTaintExist()
 		})
 	})
 })
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -199,10 +199,12 @@ var _ = Describe("snr Controller", func() {
 				node.Status.Conditions[0].Reason = "foo"
 				Expect(k8sClient.Client.Status().Update(context.Background(), node)).To(Succeed())
 
-				By("Verify that finalizer was removed and SNR can be deleted")
-				testNoFinalizer()
+				removeUnschedulableTaint(node)
 
 				verifyNoExecuteTaintRemoved()
+
+				By("Verify that finalizer was removed and SNR can be deleted")
+				testNoFinalizer()
 
 			})
 		})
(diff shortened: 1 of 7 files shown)


slintes:
hm, here it looks good...

```
[1mSTEP[0m: checking if NoExecute taint was removed
2022-06-17T13:51:35.356Z	INFO		{"taints": [{"key":"medik8s.io/remediation","value":"self-node-remediation","effect":"NoExecute"}]}
2022-06-17T13:51:45.422Z	INFO		{"taints": null}
```

controllers/controller_test.go
```diff
@@ -199,11 +199,13 @@ var _ = Describe("snr Controller", func() {
 				node.Status.Conditions[0].Reason = "foo"
 				Expect(k8sClient.Client.Status().Update(context.Background(), node)).To(Succeed())
 
-				By("Verify that finalizer was removed and SNR can be deleted")
-				testNoFinalizer()
+				removeUnschedulableTaint(node)
```
mshitrit, line 202: Hmm I'm not clear about that.
I think it should be done in the actual code (not test code), otherwise `verifyNoExecuteTaintRemoved` doesn't make much sense

slintes: this is the NoSchedule taint, simulating the node lifecycle controller, just like with adding it a few line above 🙂

slintes:
> I think that when [restoring the Node Back Up](https://github.com/medik8s/self-node-remediation/pull/30/files#diff-dd6f1328b9432817228e47edffdc3fe1e4570bfa2e39099223c05131e89e9068L643) we need to remove the taint from the backup

it's done here: https://github.com/medik8s/self-node-remediation/blob/224e9472ec40fddf35c18394e30b210b2f353dc3/controllers/selfnoderemediation_controller.go#L342-L344
Otherwise this test would fail: https://github.com/medik8s/self-node-remediation/blob/224e9472ec40fddf35c18394e30b210b2f353dc3/controllers/controller_test.go#L204

PR #32, opened 2022-06-26
Remove taint and finalizer only when SNR is targeted to delete by NHC

mshitrit:
- Removing the no Execute taint only when SNR is targeted for deletion, in order to prevent a situation where the taint is removed but the node isn't ready for workload yet.

### DIFF
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -138,7 +138,7 @@ var _ = Describe("snr Controller", func() {
 	Context("Unhealthy node with api-server access", func() {
 		var beforeSNR time.Time
 		var remediationStrategy selfnoderemediationv1alpha1.RemediationStrategyType
-		var isDeleteSNR = true
+		var isSNRNeedsDeletion = true
 		JustBeforeEach(func() {
 			createSelfNodeRemediationPod()
 			updateIsRebootCapable("true")
@@ -150,10 +150,10 @@ var _ = Describe("snr Controller", func() {
 		})
 
 		AfterEach(func() {
-			if isDeleteSNR {
+			if isSNRNeedsDeletion {
 				deleteSNR(snr)
 			}
-			isDeleteSNR = true
+			isSNRNeedsDeletion = true
 		})
 
 		Context("NodeDeletion strategy", func() {
@@ -244,7 +244,7 @@ var _ = Describe("snr Controller", func() {
 				verifyNoExecuteTaintExist()
 
 				deleteSNR(snr)
-				isDeleteSNR = false
+				isSNRNeedsDeletion = false
 
 				verifyNodeIsSchedulable()
 
--- a/controllers/controller_test.go
+++ b/controllers/controller_test.go
@@ -138,7 +138,7 @@ var _ = Describe("snr Controller", func() {
 	Context("Unhealthy node with api-server access", func() {
 		var beforeSNR time.Time
 		var remediationStrategy selfnoderemediationv1alpha1.RemediationStrategyType
-
+		var isDeleteSNR = true
 		JustBeforeEach(func() {
 			createSelfNodeRemediationPod()
 			updateIsRebootCapable("true")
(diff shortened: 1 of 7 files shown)


controllers/controller_test.go
```diff
@@ -230,24 +233,26 @@ var _ = Describe("snr Controller", func() {
...
-
-				verifyNoExecuteTaintExist()
-
 				verifyNoWatchdogFood()
 
 				verifySelfNodeRemediationPodDoesntExist()
 
 				verifyVaDeleted(vaName)
 
+				verifyFinalizerExists()
```
slintes, line 242: why did you move these 2 verifies down here? Finalizer and taint are set early in the remdiaton process, so IMHO we should also test it early

controllers/controller_test.go
```diff
@@ -230,24 +233,26 @@ var _ = Describe("snr Controller", func() {
...
+
+				verifyNoExecuteTaintExist()
+
+				deleteSNR(snr)
+				isDeleteSNR = false
```
slintes, line 247: the name `isDeleteSNR` is confusing... 🤔

controllers/controller_test.go
mshitrit: Due to the code change, now they are deleted at a later stage (only when NHC triggers deletion) so I thought it make sense for the validation code to reflect that.

controllers/controller_test.go
mshitrit: >Maybe in this case that var isn't needed at all, just ignore isNotFound error when trying to delete the CR?

This will affect almost all the tests in this file making them less robust IMO, so I'd prefer to avoid that.

Regarding the var name personally I'm fine with that name, but I'm defiantly open to other suggestions :smile:

controllers/controller_test.go
slintes: I can't follow. This test ensures it exists, not that it's deleted... 🤔

controllers/controller_test.go
slintes: anything which is meaningful without reading the code would be fine.
for `isDeleteSNR` I needed to read the code...
maybe `isSNRDeleted` (would switch true / false meaning)
or `needsSNRDeletion` or `deleteSNRInAfterEach`
...

controllers/controller_test.go
mshitrit: You are correct, however as a rule of we should verify something exit at the latest possible point.

For example let's say that we have a bug in the code and the taint is deleted just after Watchdog is stopped (`verifyNoWatchdogFood`) in the previous order we could have easily missed that bug but in current order it will be caught.

controllers/controller_test.go
mshitrit: >anything which is meaningful without reading the code would be fine.

I guess that is subjective, for me `isDeleteSNR` fits this criteria perfectly,
This is how I read it:
is - an indication that this is a flag (or whether or not to preform the action)
Delete SNR - the action that should be done if the flag is true.

I'll switch to `isSNRneedsDeletion` I'm hoping it's something we can have consensus on  :crossed_fingers:

controllers/controller_test.go
slintes: >  we should verify something exist at the latest possible point

Sorry but I disagree. Setting the finalizer and taint is the first thig we need to do, so IMHO it should also be the first thing to test. 

> we have a bug in the code and the taint is deleted just after Watchdog is stopped

Then this is a 2nd test. Call verify again instead of moving it around 🙂

mshitrit: >Setting the finalizer and taint is the first thig we need to do, so IMHO it should also be the first thing to test.

IMO it makes better sense to verify according to removal order (i.e taint is removed last so verified it still exist last).
We can catch more bugs using that approach (per previous example), so the value in that approach is evident beyond personal taste on what is more readable or not (which is far more arguable).

>Then this is a 2nd test. Call verify again instead of moving it around :slightly_smiling_face: 

Since our tests are meant to deal with async code (i.e using `Eventually` to verify) I don't see any concrete value for that (i.e bugs that we will catch from having double taint verification vs just one at the end) please let me know if I'm missing something here.

slintes:
please rebase next time to get other changes in, there are unrelated commits now in this PR, which  makes reviews harder

/lgtm

PR #34, opened 2022-06-28
Use ubi-micro as base image in Dockerfile

razo7:
Updating the current base image, [ubi](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#con_understanding-the-ubi-micro-images_assembly_types-of-container-images), to a newer and even smaller image , [ubi-micro](https://www.redhat.com/en/blog/introduction-ubi-micro), for building the operator (in Dockerfile).

Using smaller image minimizes the attack surface of container images, and is suitable for minimal applications.

See similar PR in NMO - https://github.com/medik8s/node-maintenance-operator/pull/37

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,8 @@ COPY install/ install/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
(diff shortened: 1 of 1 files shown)


mshitrit:
> Pre Submit has failed with the same error that the last PR post submit action https://github.com/medik8s/self-node-remediation/runs/6978854320?check_suite_focus=true had, but in #33 it passed.
> 
> >

Looks like the test isn't stable because we are getting a conflict with another action that is changing the node (maybe taint removal).
I'll create a PR to make the test stable.

PR #35, opened 2022-06-29
Stabilizing tests update

mshitrit:
Replacing single update attempt with multiple attempts in order to stabilize the tests and avoid 409 conflict error.


controllers/controller_test.go
```diff
@@ -550,3 +551,20 @@ func testNoFinalizer() {
...
+func eventuallyUpdateNode(updateFunc func(*v1.Node), isStatusUpdate ...bool) {
+	By("Verify that node was updated successfully")
+
+	EventuallyWithOffset(1, func() error {
+		node := &v1.Node{}
+		if err := k8sClient.Client.Get(context.TODO(), unhealthyNodeNamespacedName, node); err != nil {
+			return err
+		}
+		updateFunc(node)
+		if len(isStatusUpdate) > 0 && isStatusUpdate[0] {
```
slintes, line 564: IMHO the code would be more simple and readable when `isStatusUpdate` is a normal bool instead of a variadic, especially because having more than one of them will never make sense

slintes (changes requested):

PR #39, opened 2022-07-18
Add new icon dynamically

slintes:
- Added new red and blue icons
- Use blue one by default when creting bundle


slintes:
/cherrypick release-0.4

let's get the new icon in to the next bugfix release

PR #40, opened 2022-07-28
Allow watchdog grace period to reboot before "kicking" with software reboot

mshitrit:
This PR suppose to solve [this](https://issues.redhat.com/browse/ECOPROJECT-879) ticket.

- At the moment a software reboot occurs immediately after watchdog is stopped so watchdog does not get a chance to reboot the node.
- We'd like on the one hand to give the watchdog some grace period to reboot (one motivation is to allow enough time to generate kernel dump ) and on the other hand to have software reboot as a backup in case watchdog reboot isn't working.

### DIFF
--- a/pkg/reboot/rebooter.go
+++ b/pkg/reboot/rebooter.go
@@ -2,6 +2,8 @@ package reboot
 
 import (
 	"os/exec"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/medik8s/self-node-remediation/pkg/watchdog"
@@ -9,7 +11,7 @@ import (
 
 type Rebooter interface {
 	// Reboot triggers a node reboot
-	Reboot() error
+	Reboot() (ctrl.Result, error)
 }
 
 var _ Rebooter = &WatchdogRebooter{}
@@ -27,27 +29,47 @@ func NewWatchdogRebooter(wd watchdog.Watchdog, log logr.Logger) Rebooter {
 	}
 }
 
-func (r *WatchdogRebooter) Reboot() error {
+func (r *WatchdogRebooter) Reboot() (ctrl.Result, error) {
+	if r.isWatchdogRebootStuck() {
+		return r.softwareReboot()
+	}
+	//Watch dog is rebooting, wait to make sure watchdog is rebooting properly otherwise intervene with software reboot
+	if r.wd.IsRebooting() {
+		r.log.Info("watchdog is waiting for reboot to commence, waiting for 10 seconds")
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
+	}
+
 	if r.wd == nil || !r.wd.IsStarted() {
 		r.log.Info("no watchdog is present on this host, trying software reboot")
 		//we couldn't init a watchdog so far but requested to be rebooted. we issue a software reboot
-		if err := r.softwareReboot(); err != nil {
-			r.log.Error(err, "failed to run reboot command")
-			// TODO retry because of this?
-			//return err
-			return nil
-		}
-		return nil
+		return r.softwareReboot()
 	}
 	// we stop feeding the watchdog for a reboot
 	r.wd.Stop()
 	r.log.Info("watchdog feeding has stopped, waiting for reboot to commence")
-	return nil
+	return ctrl.Result{RequeueAfter: time.Second * 10}, nil
 }
 
 // softwareReboot performs software reboot by running systemctl reboot
-func (r *WatchdogRebooter) softwareReboot() error {
+func (r *WatchdogRebooter) softwareReboot() (ctrl.Result, error) {
 	// hostPID: true and privileged:true required to run this
 	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot", "--force", "--force")
-	return rebootCmd.Run()
+
+	if err := rebootCmd.Run(); err != nil {
+		r.log.Error(err, "failed to run reboot command")
+		// TODO retry because of this?
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *WatchdogRebooter) isWatchdogRebootStuck() bool {
+	timeToAssumeRebootHasStarted := time.Second * 30
+	lastFoodTime := r.wd.LastFoodTime()
+	var isStuck bool
+	now := time.Now()
+	if isStuck = now.After(lastFoodTime.Add(timeToAssumeRebootHasStarted)); isStuck {
+		r.log.Info("watchdog reboot is stuck, about to try software reboot")
+	}
+
+	return isStuck
 }
--- a/pkg/watchdog/interface.go
+++ b/pkg/watchdog/interface.go
@@ -11,6 +11,8 @@ type Watchdog interface {
 	Start(ctx context.Context) error
 	// IsStarted return if the watchdog is running
 	IsStarted() bool
+	// IsRebooting return if the watchdog has already stopped feeding and currently about to reboot the node
+	IsRebooting() bool
 	// Stop stops feeding the watchdog, which results in a reboot of the node
 	Stop()
 	// GetTimeout returns the watchdog timeout when it reboots the node without feeding
--- a/pkg/watchdog/synchronized.go
+++ b/pkg/watchdog/synchronized.go
@@ -15,13 +15,13 @@ var _ Watchdog = &synchronizedWatchdog{}
 
 // synchronizedWatchdog implements the Watchdog interface with synchronized calls of the implementation specific methods
 type synchronizedWatchdog struct {
-	impl         watchdogImpl
-	timeout      time.Duration
-	isStarted    bool
-	stop         context.CancelFunc
-	mutex        sync.Mutex
-	lastFoodTime time.Time
-	log          logr.Logger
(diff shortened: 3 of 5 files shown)


pkg/reboot/rebooter.go
```diff
@@ -27,27 +29,47 @@ func NewWatchdogRebooter(wd watchdog.Watchdog, log logr.Logger) Rebooter {
 	}
 }
 
-func (r *WatchdogRebooter) Reboot() error {
+func (r *WatchdogRebooter) Reboot() (ctrl.Result, error) {
+	if r.isWatchdogRebootStuck() {
```
slintes, line 33: you need to check if we have a watchdog at all as first thing (wd != nil)

pkg/reboot/rebooter.go
```diff
@@ -27,27 +29,47 @@ func NewWatchdogRebooter(wd watchdog.Watchdog, log logr.Logger) Rebooter {
...
+
+	if err := rebootCmd.Run(); err != nil {
+		r.log.Error(err, "failed to run reboot command")
+		// TODO retry because of this?
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *WatchdogRebooter) isWatchdogRebootStuck() bool {
+	timeToAssumeRebootHasStarted := time.Second * 30
```
slintes, line 66: please make this a const on top of the file

pkg/reboot/rebooter.go
```diff
@@ -27,27 +29,47 @@ func NewWatchdogRebooter(wd watchdog.Watchdog, log logr.Logger) Rebooter {
...
+	lastFoodTime := r.wd.LastFoodTime()
+	var isStuck bool
+	now := time.Now()
+	if isStuck = now.After(lastFoodTime.Add(timeToAssumeRebootHasStarted)); isStuck {
+		r.log.Info("watchdog reboot is stuck, about to try software reboot")
```
slintes, line 71: this method should not know what we do when we return it's stuck, the caller should log this IMHO

slintes: that, and introducing the const, and skipping unneeded new vars can make this a one liner:
`return time.Now().After(r.wd.LastFoodTime().Add(timeToAssumeRebootHasStarted))`

pkg/watchdog/interface.go
```diff
@@ -11,6 +11,8 @@ type Watchdog interface {
 	Start(ctx context.Context) error
 	// IsStarted return if the watchdog is running
 	IsStarted() bool
+	// IsRebooting return if the watchdog has already stopped feeding and currently about to reboot the node
```
slintes, line 14: `IsStopped` would be a better name IMHO, because watchdogs can do more than just reboots

pkg/watchdog/synchronized.go
```diff
@@ -15,13 +15,13 @@ var _ Watchdog = &synchronizedWatchdog{}
...
-	impl         watchdogImpl
-	timeout      time.Duration
-	isStarted    bool
-	stop         context.CancelFunc
-	mutex        sync.Mutex
-	lastFoodTime time.Time
-	log          logr.Logger
+	impl                   watchdogImpl
+	timeout                time.Duration
+	isStarted, isRebooting bool
```
slintes, line 20: while this is possible of course, I think it's best practice to use a new line for each field?

pkg/reboot/rebooter.go
```diff
@@ -2,14 +2,16 @@ package reboot
...
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/medik8s/self-node-remediation/pkg/watchdog"
 )
 
 type Rebooter interface {
 	// Reboot triggers a node reboot
-	Reboot() error
+	Reboot() (ctrl.Result, error)
```
slintes, line 14: I don't think we need the `Result` here, and the logic for setting `Requeue` if needed.
The reconciler is already calling `Reboot()` until it detected the reboot. I think just the `Requeue` over there needs to be changed to a shorter interval than now.

slintes (changes requested):
AFAIK it never was intentional that we fallback to soft reboot when creating the watchdog was successful, but disarming it didn't cause a reboot 🤔 This comment kind of approves that assumption: `we couldn't init a watchdog so far but requested to be rebooted. we issue a software reboot`. It does not mention a failed reboot by watchdog, but just failed initialization.
The bugfix then would be: do not set `isStarted` to false when stopping the watchdog, but introduce `isStopped`.

However, doing a soft reboot in this case makes sense. Left some comments...

slintes:
> This PR suppose to solve [this](https://issues.redhat.com/browse/ECOPROJECT-879) ticket.

Automation does not pick this up like this. I always just write the issue name like ECOPROJECT-xyz without link, and automation makes it a link, and adds PR to the issue

pkg/watchdog/interface.go
mshitrit: My problem with `IsStopped` is that it can be interpreted in two different ways:
1. A previously running watchdog was stopped, so now it `isStopped`
2. An existing watchdog was never started therefore it `isStopped`

I'm looking for a meaningful names that will only match the first option.

pkg/watchdog/synchronized.go
mshitrit: As a rule of thumb I think you are right, however in this case those variables are closely related so I think this grouping makes sense.

pkg/reboot/rebooter.go
mshitrit: >I think just the Requeue over there needs to be changed to a shorter interval than now.

I think I'm missing something, I assume you refer to requeue [here](https://github.com/medik8s/self-node-remediation/blob/0eb8d6ca7aa1e65371bcc4e6f89fde3758341455/controllers/selfnoderemediation_controller.go#L418) instead of in the reboot method (at the moment there is no timeout). 

>The reconciler is already calling Reboot() until it detected the reboot

This is true, however our use-case is different, we need to allow the watchdog some time to reboot before applying a safeguard software reboot, at the moment the fields  which are needed (i.e mainly watchdog inner params) for this logical decision (i.e applying or not the software safeguard reboot)  are not exposed at the higher scope.

The code can be refactored in a such a way that those fields will be available in a higher level scope, but IMO it'll be wrong.

mshitrit:
> AFAIK it never was intentional that we fallback to soft reboot when creating the watchdog was successful, but disarming it didn't cause a reboot

Initially I was under the same impression but it turns out that both of us were somewhat wrong.
The software reboot right after watchdog disarm was by [design](https://issues.redhat.com/browse/ECOPROJECT-879?focusedCommentId=20518937&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20518937).
It was intended in either case (i.e whether watchdog disarming caused reboot or not), IIRC per Andrew explanation this was just another safeguard to make sure the reboot occurs (kernel dumps were not considered at the time).

pkg/reboot/rebooter.go
mshitrit: >this method should not know what we do when we return it's stuck, the caller should log this IMHO

I agree, I've changed the logging

>that, and introducing the const, and skipping unneeded new vars can make this a one liner:

I've kept some (different) logging so I don't think it's relevant.

mshitrit:
> Automation does not pick this up like this. I always just write the issue name like ECOPROJECT-xyz without link, and automation makes it a link, and adds PR to the issue

Thanks  :+1: , I've changed the description.

pkg/watchdog/interface.go
beekhof: `isTriggered` or `isTerminating` perhaps?

beekhof:
> The software reboot right after watchdog disarm was by [design](https://issues.redhat.com/browse/ECOPROJECT-879?focusedCommentId=20518937&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20518937).

It could be however, that the original design was bad, or that different tradeoffs are now more appropriate.
If you've thought through the implication and feel that removing the reboot is correct, then 👍 from me.

pkg/watchdog/interface.go
mshitrit: >isTriggered or isTerminating perhaps?

Thanks !
both SGTM, @slintes do you have any preference ?

mshitrit: Yep, I like this idea.
Done.

pkg/reboot/rebooter.go
mshitrit: Indeed, Done :+1:

pkg/reboot/rebooter.go
```diff
@@ -28,26 +32,55 @@ func NewWatchdogRebooter(wd watchdog.Watchdog, log logr.Logger) Rebooter {
...
-		if err := r.softwareReboot(); err != nil {
-			r.log.Error(err, "failed to run reboot command")
-			// TODO retry because of this?
-			//return err
-			return nil
-		}
+		return r.softwareReboot()
+	}
+
+	if r.isWatchdogRebootStuck() {
```
slintes, line 41: can we move this into the `Triggered` case below?

slintes (commented):
Yes this looks much better to me 👍🏼 
Only one little nit inside.

pkg/reboot/rebooter.go
mshitrit: Yep, I agree :+1:

slintes:
retriggering test after CI config update

/test all

PR #41, opened 2022-07-28
Updating SNR readme - noting it replaces Poison Pill

mshitrit:


slintes:
Do we really need / want this? We want to get rid of the old name, this adds it at a new place 🤷🏼‍♂️ IMHO the deprecation notice on PP is good enough.

mshitrit:
> Do we really need / want this? We want to get rid of the old name, this adds it at a new place 🤷🏼‍♂️ IMHO the deprecation notice on PP is good enough.

No objection on my part.

@razo7 any objection that I'll close this PR ?

razo7:
> > Do we really need / want this? We want to get rid of the old name, this adds it at a new place 🤷🏼‍♂️ IMHO the deprecation notice on PP is good enough.
> 
> No objection on my part.
> 
> @razo7 any objection that I'll close this PR ?

None. The deprecation notice is indeed good enough.

mshitrit:
Per [discussion](https://github.com/medik8s/self-node-remediation/pull/41#issuecomment-1200408879) this PR is unneeded.

PR #47, opened 2022-09-12
Remove Node Deletion remediation

mshitrit:
Removing Node Deletion remediation for snr version 0.5.0, as part of ECOPROJECT-1011 ticket.


epenchev:
Hi, why is Node Deletion remediation been removed? Unfortunately I can't access the ticket and was hoping for more details in the PR. 
Thanks.

mshitrit:
> Hi, why is Node Deletion remediation been removed? Unfortunately I can't access the ticket and was hoping for more details in the PR. Thanks.

Hi @epenchev Thanks for reaching out !
In a nutshell we think that `Node Deletion remediation` is somewhat flaky, we prefer to use `Resource Deletion remediation` since it's more stable and has no disatvanges (that we could find) compared to `Node Deletion remediation`.

As far as the road-map and current situation: 
- Current SNR version (0.4.X) support both, however `Resource Deletion remediation` is the default remediation.
- In the next version (0.5.X which is planned for December 2022) `Node Deletion remediation` is planned to be removed completely.

epenchev:
Hi @mshitrit, thanks for the explanation. 
Now I see both, `Node Deletion remediation` and `Resource Deletion remediation` are pretty much alike. And both do reboot the Node and markNodeAsUnschedulable as well.

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -34,11 +33,11 @@ type RemediationStrategyType string
 // SelfNodeRemediationSpec defines the desired state of SelfNodeRemediation
 type SelfNodeRemediationSpec struct {
 	//RemediationStrategy is the remediation method for unhealthy nodes
-	//could be either "NodeDeletion" or "ResourceDeletion"
+	//currently only option is "ResourceDeletion"
 	//the first will delete the node to signal to the cluster that the node was fenced
```
slintes, line 37: the first.. the latter doesn't make sense anymore

api/v1alpha1/selfnoderemediationtemplate_types.go
```diff
@@ -85,17 +84,5 @@ func NewRemediationTemplates() []*SelfNodeRemediationTemplate {
 				},
 			},
 		},
-		{
-			ObjectMeta: metav1.ObjectMeta{
```
slintes, line 89: IMHO we need to also check if this one exists in the cluster, and if so delete it?

controllers/controller_test.go
```diff
@@ -93,56 +83,12 @@ var _ = Describe("snr Controller", func() {
 
 	})
 
-	Context("Unhealthy node with self-node-remediation pod but unable to reboot", func() {
```
slintes, line 96: isn't this a valid test case for resource deletion strategy? 🤔

controllers/controller_test.go
```diff
@@ -258,37 +148,6 @@ var _ = Describe("snr Controller", func() {
 		})
 	})
 
-	Context("Unhealthy node without api-server access", func() {
```
slintes, line 261: isn't this a valid test case for resource deletion strategy? :thinking:

e2e/self_node_remediation_test.go
```diff
@@ -168,41 +155,6 @@ var _ = Describe("Self Node Remediation E2E", func() {
 			})
 		})
 
-		Context("Unhealthy node (with SNR)", func() {
```
slintes, line 171: Isn't this a valid test case for resource deletion strategy? :thinking:
You need to adapt the test though, don't check for node deletion/recreation, but resouce deletions?
Same for the test above, old line 163 doesn't make sense anymore (can't comment there)

slintes (changes requested):

api/v1alpha1/selfnoderemediationtemplate_types.go
mshitrit: IIUC your aiming for the use case of cleaning leftovers from a previous installation of SNR.
I don't disagree but I don't think it's a blocker, IMO the proper place to do it would be along with other cleanup procedures (old daemonsets, template, config ect).
If you agree I can open a ticket for that.

slintes: I agree we have more things to clean up after uninstall of SNR, but for this one (the deleted template) I'd prefer to deal with it in this PR. NHC makes a lot more noise when a template isn't found (log, condition, event) than SNR does for an unknown strategy (just a log). I want to make sure users have a good chance to notice that their current config doesn't work anymore. By deleting the default node deletion template we increase that chance significantly IMHO.

slintes:
> Re-adding Node Deletion Strategy Type. Even though it won't be used it may exist in a legacy SNR Template, in order for that template to be automatically deleted the strategy need to exist.

Why does the strategy need to exist for deleting the tempate? 🤔 
By keeping it in the CRD we also allow to create new templates and CRs with that type.

mshitrit:
> > Re-adding Node Deletion Strategy Type. Even though it won't be used it may exist in a legacy SNR Template, in order for that template to be automatically deleted the strategy need to exist.
> 
> Why does the strategy need to exist for deleting the tempate? thinking By keeping it in the CRD we also allow to create new templates and CRs with that type.

I've experienced some issues (couldn't get the operator to start) when I've removed node deletion strategy from the CRD and had a legacy template containing it it didn't happen when the legacy template was deleted so  I've just assumed that this is the cause.

slintes (commented):
I think we need a plan to get rid of the deprecated NodeDeletion enum value. I'm afraid that with just ignoring it in the CRs and always doing ResourceDeletion users will ignore logs and events and never change their templates (and our?) templates...

slintes:
summarizing some discussions around this topic:

- we can't delete the NodeDeletion enum value from the CRD, because that would break upgrades on clusters with existing templates with NodeDeletion strategy, like one of our own default templates
- we are exploring options to at least prevent creating or updating CRs with NodeDeletion (switching to validating webhook, new API version, ...)
- the latter will be implemented in a follow up

/lgtm

slintes (approved):

PR #52, opened 2022-11-06
Set SNR version to 0.5.0

mshitrit:
Signed-off-by: Michael Shitrit <mshitrit@redhat.com>

### DIFF
--- a/config/manifests/bases/self-node-remediation.clusterserviceversion.yaml
+++ b/config/manifests/bases/self-node-remediation.clusterserviceversion.yaml
@@ -6,10 +6,10 @@ metadata:
     capabilities: Basic Install
     categories: OpenShift Optional
     containerImage: quay.io/medik8s/self-node-remediation-operator:latest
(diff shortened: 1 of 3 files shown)


config/manifests/bases/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -6,10 +6,10 @@ metadata:
     capabilities: Basic Install
     categories: OpenShift Optional
     containerImage: quay.io/medik8s/self-node-remediation-operator:latest
-    createdAt: "2022-09-11 12:36:32"
+    createdAt: "2022-11-06 08:57:52"
     description: Self Node Remediation Operator for remediate itself in case of a
       failure.
-    olm.skipRange: '>=0.1.0 <0.4.0'
+    olm.skipRange: '>=0.1.0 <0.5.0'
```
razo7, line 12: Does SNR have v0.4.0 or lower? I think only PP, thus we can potentially use `v0.4.0` as the minimum version for upgrades.

PR #53, opened 2022-11-06
[WIP] setup psa

mshitrit:
This PR is related to [ECOPROJECT-812](https://issues.redhat.com//browse/ECOPROJECT-812) 

When SNR starts it'll set `"pod-security.kubernetes.io/enforce"=privileged` and `"security.openshift.io/scc.podSecurityLabelSync"=false` labels on the namespace.

- `pod-security.kubernetes.io/enforce` enables the creation of the snr agent pod (which requires root permissions).
- `security.openshift.io/scc.podSecurityLabelSync` disables the label syncer on that namespace in order to make sure the previous label isn't removed. 

I've done the following testes
- I've made sure snr can't run the agents pods without root permissions (tested on OCP)
- I've made sure that a pod with root permissions can be installed when the above labels are set on the ns (tested on 1.25 k8s  cluster created with Kind)
-  I've made sure that a pod with root permissions can't be installed without `pod-security.kubernetes.io/enforce` label, or with restricted value(tested on 1.25 k8s  cluster created with Kind)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,6 +224,12 @@ bundle-build: ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	docker push $(BUNDLE_IMG)
 
+# Generate bundle manifests and metadata, then validate generated files.
+.PHONY: bundle-k8s
+bundle-k8s: bundle
+	$(KUSTOMIZE) build config/manifests-k8s | $(OPERATOR_SDK) generate --verbose bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	$(OPERATOR_SDK) bundle validate ./bundle
+
 .PHONY: protoc ## Download protoc (protocol buffers tool needed for gRPC)
 PROTOC = $(shell pwd)/bin/proto/bin/protoc
 protoc: protoc-gen-go protoc-gen-go-grpc
--- a/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
+++ b/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
@@ -365,9 +365,12 @@ spec:
                     memory: 110Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
               priorityClassName: system-cluster-critical
               securityContext:
-                runAsNonRoot: false
+                runAsNonRoot: true
               serviceAccountName: self-node-remediation-controller-manager
               terminationGracePeriodSeconds: 10
       permissions:
--- a/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
+++ b/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
@@ -370,7 +370,7 @@ spec:
                     - ALL
               priorityClassName: system-cluster-critical
               securityContext:
-                runAsNonRoot: true
+                runAsNonRoot: false
               serviceAccountName: self-node-remediation-controller-manager
               terminationGracePeriodSeconds: 10
       permissions:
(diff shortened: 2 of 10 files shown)


slintes:
I think we need to  differentiate more between upstream k8s and OCP:

k8s:
- I don't think we should update the pod security label programmatically, but document instead we need privileged, and admins need to update the namespace if needed.
- The sync label should not be applied at all, it's OCP specific

OCP:
- Also here we should not change namespace labels, but document that we need changes (if so)
- But actually I'm surprised that we need to change labels at all. My expection is that the label syncer updates the pod security label to privileged once we try to deploy the daemonset? (didn't try myself though yet, but e2e still runs successfully, not?)

Last but not least, we might want to update the daemonset to be more specific what we need? (e.g. also add "drop all caps" and just re-add what we need)

beekhof:
Just as a point of reference, CNV definitely messes with namespace labels, so there is some precedent.

I don't yet grok the work-around(s) needed to make SNR useful out-of-the-box, but in general it is fair to claim that the admin installing the operator is sufficient signal that they agree to it's security requirements (which I agree we should document, also in the install-time metadata for NHC, and generally be very transparent about).

To my mind, the most likely consequence of adding a manual post-installation step is that people will just assume SNR is broken.  Or worse, assume it's working until they most need it. So that would need to be weighed against the undesirability of the work-around(s).

slintes:
I did some tests on latest WIP OCP 4.12, with the latest NHC + SNR version from main branch (so without this PR):

- deploying in openshift-operators: no issue, it has the privileged label
- deploying in a new namespace with openshift prefix: no issue, namespace was automatically labeled by label syncer
- deploying in a new namespace without openshift prefix: no issue, namespace was automatically labeled (even though there is no sync label on the ns... 🤔 but the security label is updated when uninstalling / installing)
- deploying in default namespace: failed because of missing label sync (expected)

So at least for OCP 4.12 we are fine, and should not touch namespaces.
We need to investigate what's best practice on k8s.

beekhof:
I hate to ask, but what is the situation on 4.11?

slintes:
nothing changed on OCP 4.11, did it? 🤔

beekhof:
True, but it wasn't clear to me that the latest version of the operator had been tested on 4.11 too.  So I thought I'd ask :)

mshitrit:
>So at least for OCP 4.12 we are fine, and should not touch namespaces.

That's quite good news (I've only tested on k8s ).

I'd like to suggest only adding `pod-security.kubernetes.io/enforce` label and document this behavior  (mostly upstream for k8s sake)
- For OCP it is indeed redundat since it'll be added in any case by the label syncer, but it causes no harm  (it even might be useful in case label syncer is not functioning).
- For k8s it'll allow our operator to run - I think it's better than the alternative of letting the user set the label after the installation.

slintes:
> I'd like to suggest only adding `pod-security.kubernetes.io/enforce` label and document this behavior (mostly upstream for k8s sake)
> 
> * For OCP it is indeed redundat since it'll be added in any case by the label syncer, but it causes no harm  (it even might be useful in case label syncer is not functioning).

I disagree
- the label syncer does not add the label "in any case", see e.g.  `default` namespace, and there probably is a good reason for it
- when the label syncer is not functioning there is  bigger issue in the cluster than not being able to deploy SNR
- who reverts the label when SNR is undeployed?
- how to argue to customers that we need "update namespace" permissions? 

> * For k8s it'll allow our operator to run - I think it's better than the alternative of letting the user set the label after the installation.

As stated earlier, IMHO we should find out what's recommended on k8s instead of guessing.

mshitrit:
Following some more discussion (slack [link](https://coreos.slack.com/archives/CE4UC6AVB/p1669284599921669)) , agreed that best practice would be:
- ocp set up by label syncer (tested and working)
- k8s let the user set the label (I've created a task to update docs once k8s community version will be available)

PR #55, opened 2022-11-22
Fix test file name

slintes:
nothing to add to the title 😎


beekhof:
I don't see how renaming the test file could cause the failures which seem to be coming from must-gather. Is it appropriate to merge anyway?

slintes:
> I don't see how renaming the test file could cause the failures which seem to be coming from must-gather. Is it appropriate to merge anyway?

yep, just double checked that the actual e2e tests run successfully

/override ci/prow/e2e-test

PR #56, opened 2022-11-29
updating skipRange in generated csv

mshitrit:
updating skipRange in the generated csv which was missed in the original PR.

### DIFF
--- a/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
+++ b/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
@@ -40,7 +40,7 @@ metadata:
     createdAt: "2022-11-06 08:57:52"
     description: Self Node Remediation Operator for remediate itself in case of a
       failure.
-    olm.skipRange: '>=0.1.0 <0.5.0'
+    olm.skipRange: '>=0.4.0 <0.5.0'
     operators.operatorframework.io/builder: operator-sdk-v1.18.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/medik8s/self-node-remediation


bundle/manifests/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -40,7 +40,7 @@ metadata:
     createdAt: "2022-11-06 08:57:52"
     description: Self Node Remediation Operator for remediate itself in case of a
       failure.
-    olm.skipRange: '>=0.1.0 <0.5.0'
+    olm.skipRange: '>=0.4.0 <0.5.0'
```
razo7, line 43: It is something that I believe our CI should catch (and [fail in the E2E test](https://github.com/medik8s/self-node-remediation/blob/main/Makefile#L136-L138) of the last PR)  as unstage changes

slintes, line 43: Since 0.4.0 was the first release, this change does not make a difference in the end. But it also doesn't hurt.

slintes (approved):

bundle/manifests/self-node-remediation.clusterserviceversion.yaml
mshitrit: Agreed, we've done it awhile back (this PR is just a leftover because I didn't generate csv properly) just for good measure.

mshitrit: IIUC this is not an unstaged change, this is just a value which does not correspond to the csv it was generated from, where an unstaged change would be a file with uncommitted changes.

razo7: In other operators we have test called `test` which runs the `bundle` target (and update it's files. Including the CSV), and then there should be a verify check of unstaged changes as in [NHC](https://github.com/medik8s/node-healthcheck-operator/blob/main/Makefile#L97-L98).

PR #57, opened 2022-11-29
Better doc for safe time to assume node rebooted seconds

mshitrit:
Related to [ECOPROJECT-1077](https://issues.redhat.com//browse/ECOPROJECT-1077)
Improving config note on `SafeTimeToAssumeNodeRebootedSeconds` in order to make it more user friendly.


api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -40,9 +40,10 @@ type SelfNodeRemediationConfigSpec struct {
 	WatchdogFilePath string `json:"watchdogFilePath,omitempty"`
 
 	// SafeTimeToAssumeNodeRebootedSeconds is the time after which the healthy self node remediation
-	// agents will assume the unhealthy node has been rebooted and it is safe to remove the node
-	// from the cluster. This is extremely important. Deleting a node while the workload is still
+	// agents will assume the unhealthy node has been rebooted, and it is safe to remediate the node.
+	// This is extremely important. Remediating a node while the workload is still
 	// running there might lead to data corruption and violation of run-once semantic.
+	// Note that remediation time is affected by other factors as well, therefor modifying this field will not necessarily affect it.
```
razo7, line 46: missing **e** - `therefor` -> `therefore`

slintes (approved):

slintes:
/hold 

let's wait if @razo7 or @beekhof have feedback on this as well

api/v1alpha1/selfnoderemediationconfig_types.go
mshitrit: Thanks, fixed

config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediationconfigs.yaml
```diff
@@ -98,10 +98,12 @@ spec:
...
                   which the healthy self node remediation agents will assume the unhealthy
-                  node has been rebooted and it is safe to remove the node from the
-                  cluster. This is extremely important. Deleting a node while the
-                  workload is still running there might lead to data corruption and
-                  violation of run-once semantic.
+                  node has been rebooted, and it is safe to remediate the node. This
+                  is extremely important. Remediating a node while the workload is
+                  still running there might lead to data corruption and violation
+                  of run-once semantic. Note that remediation time is affected by
+                  other factors as well, therefore modifying this field will not necessarily
```
beekhof, line 105: This is true and more accurate, but it's a terrible message for users.
"Set this, it might do something, it might not"

The CRD is not a substitute for exhaustive documentation, but the fact we can't concisely explain what this field does is surely a sign something needs to be improved.

mshitrit: >The CRD is not a substitute for exhaustive documentation

Agreed, I assume that when/after we'll merge this CR will consider best approach for ocp docs

>This is true and more accurate, but it's a terrible message for users.
"Set this, it might do something, it might not"

How about something like this ? ``Note that SafeTimeToAssumeNodeRebootedSeconds field is not meant for modifying overall remediation time ,but to verify remediation takes place once fencing is completed - therefore modifying this field will not necessarily affect the overall remediation time.``

beekhof: That's not really true though.  It *is* meant to control the overall remediation time, but not below a level calculated from other variables.

beekhof: In what way would they be different goals?

There are two times:
a) the maximum amount of time between a failure and node committing sepuku
b) the amount of time surviving peers wait before assuming the node is in a safe state (ie. dead)

They are related because while 'b' can be as long as the user wants, it must never be smaller than 'a'.

mshitrit: I've modified the comment per our conversion, feel free to edit more

PR #59, opened 2022-12-01
Bug Fix ECOPROJECT-1028

mshitrit:
[ECOPROJECT-1028](https://issues.redhat.com//browse/ECOPROJECT-1028) bug fix.
Adding a grace time before deciding that a worker peer is unresponsive. 
This way node will not assume it's isolated due to to short network/communication issue.

I've managed to recreate the bug and test the fix on 4.12 OCP

### DIFF
--- a/pkg/apicheck/check.go
+++ b/pkg/apicheck/check.go
@@ -24,8 +24,6 @@ import (
 	"github.com/medik8s/self-node-remediation/pkg/reboot"
 )
 
-const maxTimeForNoPeersResponse = 60 * time.Second
-
 type ApiConnectivityCheck struct {
 	client.Reader
 	config                 *ApiConnectivityCheckConfig
@@ -37,18 +35,19 @@ type ApiConnectivityCheck struct {
 }
 
 type ApiConnectivityCheckConfig struct {
-	Log                logr.Logger
-	MyNodeName         string
-	CheckInterval      time.Duration
-	MaxErrorsThreshold int
-	Peers              *peers.Peers
-	Rebooter           reboot.Rebooter
-	Cfg                *rest.Config
-	CertReader         certificates.CertStorageReader
-	ApiServerTimeout   time.Duration
-	PeerDialTimeout    time.Duration
-	PeerRequestTimeout time.Duration
-	PeerHealthPort     int
+	Log                       logr.Logger
+	MyNodeName                string
+	CheckInterval             time.Duration
+	MaxErrorsThreshold        int
+	Peers                     *peers.Peers
+	Rebooter                  reboot.Rebooter
+	Cfg                       *rest.Config
+	CertReader                certificates.CertStorageReader
+	ApiServerTimeout          time.Duration
+	PeerDialTimeout           time.Duration
+	PeerRequestTimeout        time.Duration
+	PeerHealthPort            int
+	MaxTimeForNoPeersResponse time.Duration
 }
 
 func New(config *ApiConnectivityCheckConfig, controlPlaneManager *controlplane.Manager) *ApiConnectivityCheck {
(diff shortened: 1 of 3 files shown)


pkg/apicheck/check.go
```diff
@@ -24,13 +24,16 @@ import (
 	"github.com/medik8s/self-node-remediation/pkg/reboot"
 )
 
+const maxTimeForNoPeersResponse = 60 * time.Second
```
razo7, line 27: Can you elaborate why one minute?  
Please add explanation as a comment next to it for future reference whether it's a rule of thumb or it has more meaningful explanation.

mshitrit: Well in the test I've done (when recreating the bug) I saw the the max time I'm reaching is about 20 seconds.
60 sec seems on the one hand as a reasonable buffer and on the other hand not too long a duration to delay remediation use case of isolated node when it's actually needed

slintes: - what about adding this to ApiConnectivityCheckConfig?
- doesn't this need to go into the minTimeToAssumeNodeRebooted calculation?

slintes (changes requested):

pkg/apicheck/check.go
slintes: >  the max time I'm reaching is about 20 seconds

What exactly is happening or not happening during these 20 seconds?
I'm asking because we already wait a minute before asking peers + rebooting (`MaxErrorsThreshold` * (`CheckInterval` + `ApiServerTimeout`). This change adds another minute, while everyone is asking us to be quicker.

mshitrit: >what about adding this to ApiConnectivityCheckConfig?

Agreed :+1: , however I still prefer not to expose this to the user.

>doesn't this need to go into the minTimeToAssumeNodeRebooted calculation?

This is a very good question which indeed I haven't initially considered.
If we want to stay on the safe side we probably should, however this use-case is relevant only for a worker node which is completely isolated (which I assume is pretty rare) so maybe we can discuss other alternatives. 

>What exactly is happening or not happening during these 20 seconds?

For now the only way I've managed to recreate this bug is by taking down 2 out of 3 of the control-plane nodes, once I take 2 control-plane nodes out I witness a short time where worker nodes can't communicate to each other (TBH I don't really know why).

epenchev: Is this somehow aligned with the `--node-monitor-grace-period` duration option for the kube-controller-manager?
If those 2 timeouts are misaligned can we have rebooting worker nodes that are still not considered unhealthy from K8s perspective.
I don't know how often cluster admins tend to change this option and in what direction is the change.

slintes: > which I assume is pretty rare

even when it's rare, IMHO we shouldn't risk that fencing didn't finish yet when we start deleting pods, it can break workloads

mshitrit: >Is this somehow aligned with the --node-monitor-grace-period duration option for the kube-controller-manager?

I don't think so,  IIUC `node-monitor-grace-period` would affect the time it takes for the node to be marked as unhealthy ,where this discussion is about what happens after that.

mshitrit: >even when it's rare, IMHO we shouldn't risk that fencing didn't finish yet when we start deleting pods, it can break workloads

Agreed.
How about using a 30 second timeout (instead of 1 minute) and adding it to the min time calculation ?
This way maybe we has less certainty of always resolving the bug but at least we are still making it less likely to happen and we are not hurting the overall time as much. 

I can think of other alternatives of calculating the minTime during runtime according to the type of remediation (i.e different calculation for the use-case of isolated node) but it seems rather complex and IMO would be an overkill for this.

PR #60, opened 2022-12-04
Remove NodeBackup

mshitrit:
Remove NodeBackup from status, since it's only relevant in the Deprecated NodeDeletion strategy.


slintes (changes requested):
As discussed in chat, this is a breaking API change. While the chance might be low that it actually causes issues, we should prevent that risk. It's better to keep the field and mark it as deprecated until we introduce a new API version.

We can still remove the leftovers in the actual code.

PR #72, opened 2022-12-20
Verify bundle

mshitrit:
Verifying bundle & manifest are updated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,13 +23,13 @@ export IMAGE_REGISTRY
 
 # When no version is set, use latest as image tags
 DEFAULT_VERSION := 0.0.1
-ifeq ($(origin VERSION), undefined)
-IMAGE_TAG = latest
-else ifeq ($(VERSION), $(DEFAULT_VERSION))
+
+ifeq ($(VERSION), $(DEFAULT_VERSION))
 IMAGE_TAG = latest
 else
 IMAGE_TAG = v$(VERSION)
 endif
+
 export IMAGE_TAG
 CHANNELS = stable
 export CHANNELS
--- a/Makefile
+++ b/Makefile
@@ -328,5 +328,10 @@ vendor:
 tidy:
 	go mod tidy
 
+##Verifies vendor and tidy didn't cause changes
 .PHONY:verify-vendor
 verify-vendor:tidy vendor verify-no-changes
+
+##Verifies bundle and manifests didn't cause changes
+.PHONY:verify-bundle
+verify-bundle: manifests bundle verify-no-changes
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -222,14 +222,24 @@ export ICON_BASE64 ?= ${DEFAULT_ICON_BASE64}
...
+	# set skipRange
+	sed -r -i "s|olm.skipRange: .*|olm.skipRange: '>=0.4.0 <${VERSION}'|;" ./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
+	# set icon (not version or build date related, but just to not having this huge data permanently in the CSV)
+	sed -r -i "s|base64data:.*|base64data: ${ICON_BASE64}|;" ./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
+	$(OPERATOR_SDK) bundle validate ./bundle
+
+## Build the bundle image.
 .PHONY: bundle-build
-bundle-build: ## Build the bundle image.
+bundle-build: bundle bundle-update
```
razo7, line 242: NIT: Can we use comments for targets in the same line of the target?

Suggestion

> bundle-build: ## Build the bundle image.
 $(MAKE)  bundle bundle-update
docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .

With this change running `make help` will present the comment next to the target (e.g. `bundle-build: Build the bundle image.`)

Makefile
```diff
@@ -310,8 +320,18 @@ container-build: ## Build containers
 container-push: ## Push containers (NOTE: catalog can't be build before bundle was pushed)
 	make docker-push bundle-push catalog-build catalog-push
 
-.PHONY:verify-vendor
-verify-vendor:
-	go mod tidy
+.PHONY:vendor
+vendor:
```
razo7, line 324: NIT: some targets are even missing any comment/description

bundle/manifests/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -36,7 +36,7 @@ metadata:
       ]
     capabilities: Basic Install
     categories: OpenShift Optional
-    containerImage: quay.io/medik8s/self-node-remediation-operator:0.5.0
+    containerImage: quay.io/medik8s/self-node-remediation-operator:latest
```
razo7, line 39: If we set v0.0.1 in the images below then why we use `latest` in here?

mshitrit: good catch, it had to do with order of declaration of the vars, fixed.

razo7: Isn't it because of https://github.com/medik8s/self-node-remediation/pull/72/commits/aaa80da2b9472f4eb0b7ff3d4ce3e61f3cb0ac67#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L27?

PR #73, opened 2022-12-21
Release 0.5 updates

mshitrit:
- adding eus channel
- updating bundle


mshitrit:
eus only relevant in downstream

PR #77, opened 2023-01-01
Update controller runtime

mshitrit:
Update controller runtime from 0.9.6 to 0.13.1, also automatically updated other packages.


config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediations.yaml
```diff
@@ -82,9 +82,10 @@ spec:
                     description: Spec defines the behavior of a node. https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
                     properties:
                       configSource:
-                        description: If specified, the source to get node configuration
-                          from The DynamicKubeletConfig feature gate must be enabled
-                          for the Kubelet to use this field
+                        description: 'Deprecated: Previously used to specify the source
+                          of the node''s configuration for the DynamicKubeletConfig
```
razo7, line 86: NIT: Why double apostrophe(`''`)? 
I assume it was supposed to be `node's` rather than `node''s`

config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediations.yaml
```diff
@@ -82,9 +82,10 @@ spec:
                     description: Spec defines the behavior of a node. https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
                     properties:
                       configSource:
-                        description: If specified, the source to get node configuration
-                          from The DynamicKubeletConfig feature gate must be enabled
-                          for the Kubelet to use this field
+                        description: 'Deprecated: Previously used to specify the source
```
razo7, line 85: Is there any reason why the whole sentence/block has an apostrophe in the beginning and in the end?
Normally we don't use it

config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediations.yaml
mshitrit: We are not responsible for the text change, we just see it because of the core updates

config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediations.yaml
mshitrit: Same as comment above

PR #78, opened 2023-01-01
Watchdog init logic

mshitrit:
Fix for [ECOPROJECT-483](https://issues.redhat.com//browse/ECOPROJECT-483) .

Previously the use case where WD failed to start was ignored - which might have ultimately caused a failure to fence/remediate a failed node later on.

Now, in case software reboot **isn't enabled** SNR will fail to initialize.
In case watchdog failed to start and software reboot **is enabled** we will failover to software reboot.

### DIFF
--- a/pkg/reboot/rebooter_test.go
+++ b/pkg/reboot/rebooter_test.go
@@ -1,0 +1,64 @@
+package reboot
+
+import (
+	"context"
+	"os"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/medik8s/self-node-remediation/pkg/utils"
+	"github.com/medik8s/self-node-remediation/pkg/watchdog"
+)
+
+var isSoftwareRebootCalled bool
+
+var _ = Describe("Watchdog tests", func() {
+	var rebooter *watchdogRebooter
+
+	Context("Crash on start", func() {
+		BeforeEach(func() {
+			wd, _ := watchdog.NewFake(false)
+			rebooter = &watchdogRebooter{wd, ctrl.Log.WithName("fake rebooter"), fakeSoftwareReboot}
+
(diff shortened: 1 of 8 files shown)


pkg/utils/annotations.go
```diff
@@ -49,3 +51,12 @@ func UpdateNodeWithIsRebootCapableAnnotation(watchdogInitiated bool, nodeName st
 
 	return nil
 }
+
+func IsSoftwareRebootEnabled() (bool, error) {
+	softwareRebootEnabledEnv := os.Getenv(isSoftwareRebootEnabledEnvVar)
+	softwareRebootEnabled, err := strconv.ParseBool(softwareRebootEnabledEnv)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to convert IS_SOFTWARE_REBOOT_ENABLED env valueto boolean. value is: %s", softwareRebootEnabledEnv)
```
clobrano, line 58: just a typo here: valueto -> value to

clobrano (commented):
I noticed only a typo. For what I can tell, the rest looks good to me

pkg/reboot/rebooter_test.go
```diff
@@ -0,0 +1,64 @@
...
+				err := wd.Start(context.TODO())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(wd.Status()).To(Equal(watchdog.Malfunction))
+				//Verify reboot goes as expected
+				Expect(rebooter.Reboot()).ToNot(HaveOccurred())
+				Expect(isSoftwareRebootCalled).To(BeTrue())
+			})
+		})
+
+	})
```
razo7, line 57: NIT: I am not so much of an expert in UT but is it ok to use `Context` inside `Context`? By looking at other examples, like in https://onsi.github.io/ginkgo/, I don't see Context block inside another Context block. I think `Describe` might be better suited.
`Context("Crash ...` -> `Describe("Crash ...`

PR #80, opened 2023-01-08
Enable etcd healthcheck

mshitrit:
As part of [ECOPROJECT-1073](https://issues.redhat.com//browse/ECOPROJECT-1073) adding etcd health check as part of control-plane node diagnostics.
Tested on local cluster.

### DIFF
--- a/main.go
+++ b/main.go
@@ -278,7 +278,7 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 
 	// but the reboot time needs be at least the time we know we need for determining a node issue and trigger the reboot!
 	// 1. time for determine node issue
-	minTimeToAssumeNodeRebooted := (apiCheckInterval + apiServerTimeout) * time.Duration(maxErrorThreshold) + maxTimeForNoPeersResponse
+	minTimeToAssumeNodeRebooted := (apiCheckInterval+apiServerTimeout)*time.Duration(maxErrorThreshold) + maxTimeForNoPeersResponse
 	// 2. time for asking peers (10% batches + 1st smaller batch)
 	minTimeToAssumeNodeRebooted += (10 + 1) * (peerDialTimeout + peerRequestTimeout)
 	// 3. watchdog timeout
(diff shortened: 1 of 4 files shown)


pkg/controlplane/manager.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	kubeletPort = "10250"
+	kubeletPort    = "10250"
+	etcdHealthPort = "9980"
```
razo7, line 26: Why did you choose this port? Is it a well known port for etcd health?

mshitrit: yes, it's taken from [here](https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L205-L209)

main.go
```diff
@@ -278,7 +278,7 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 
 	// but the reboot time needs be at least the time we know we need for determining a node issue and trigger the reboot!
 	// 1. time for determine node issue
-	minTimeToAssumeNodeRebooted := (apiCheckInterval + apiServerTimeout) * time.Duration(maxErrorThreshold) + maxTimeForNoPeersResponse
+	minTimeToAssumeNodeRebooted := (apiCheckInterval+apiServerTimeout)*time.Duration(maxErrorThreshold) + maxTimeForNoPeersResponse
```
slintes, line 281: why this change?

pkg/controlplane/manager.go
```diff
@@ -82,13 +83,15 @@ func (manager *Manager) IsControlPlaneHealthy(workerPeerResponse peers.Response,
 }
 
 func (manager *Manager) isDiagnosticsPassed() bool {
+	manager.log.Info("Starting control-plane node diagnostics")
 	if manager.isEndpointAccessLost() {
 		return false
 	} else if !manager.isKubeletServiceRunning() {
```
slintes, line 89: no need for `else`

main.go
mshitrit: Not something I've done manually, I assume it's due to update in the linter

pkg/controlplane/manager.go
mshitrit: Thanks, I'll update in another PR :+1:

PR #81, opened 2023-01-09
[WIP] remove hostPID

mshitrit:
because of ECOPROJECT-1120 removing hostPID which is part of the PodSecurityPolicy.
PodSecurityPolicy was deprecated at 1.25 k8s (replaced by PSA)


config/manifests/bases/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -117,7 +117,7 @@ spec:
   - email: medik8s@googlegroups.com
     name: Medik8s Team
   maturity: alpha
-  minKubeVersion: 1.20.0
+  minKubeVersion: 1.25.0
```
slintes, line 120: this will break OCP 4.10/4.11, not?

mshitrit: I believe so, that's part of the reason it's WIP.
I wanted to have some discussion on what's the best way to go about it.

PR #84, opened 2023-01-22
use the Dockerized go version for testing docker-build

mshitrit:


razo7:
/lgtm

/hold

Makefile
```diff
@@ -93,6 +96,19 @@ ifeq (,$(shell which kubectl))
...
 endif
 
+# Run go in a container
+# --rm                                                          = remove container when stopped
+# -v $$(pwd):/home/go/src/github.com/medik8s/self-node-remediation-operator = bind mount current dir in container
+# -u $$(id -u)                                                  = use current user (else new / modified files will be owned by root)
+# -w /home/go/src/github.com/medik8s/self-node-remediation-operator         = working dir
+# -e ...                                                        = some env vars, especially set cache to a user writable dir
+# --entrypoint /bin bash ... -c                                 = run bash -c on start; that means the actual command(s) need be wrapped in double quotes, see e.g. check target which will run: bash -c "make test"
+export DOCKER_GO=docker run --rm -v $$(pwd):/home/go/src/github.com/medik8s/$(OPERATOR_NAME)-operator \
```
clobrano, line 106: Could we use $GOPATH here (if set)?

clobrano (changes requested):

Makefile
mshitrit: I don't quite follow.
IIUC  $GOPATH would be the path golang is installed locally, in this use case we don't want to use locally installed golang but rather created a docker with a specific golang version (which matches the golang version in our deployed operator) in order to make sure that even if our locally installed golang is different (or doesn't exist) than that of our operator our tests are still valid.

clobrano: right, it was a wrong observation

clobrano (approved):

PR #86, opened 2023-02-02
Add out of service taint remediation

k-keiichi-rh:
This PR is adding a new remediation strategy based on https://github.com/kubernetes/enhancements/pull/1116
In k8s 1.26, https://github.com/kubernetes/enhancements/pull/1116 is a Beta feature. This feature will be graduated to GA in k8s 1.27.

This PR is implemented based on the discussion in https://github.com/medik8s/self-node-remediation/issues/17.
The following is the new remediation strategy for the out-of-service taint:

1. When detecting node failure by NHC, SNR sets NoExecute taint on the failed Node
2. The failed node is rebooted by watchdog
3. SNR waits for a fixed time to complete rebooting the failed node
4. SNR sets the out-of-service taint
=> This taint expects that the node is in shutdown or power off state (not in the middle of restarting) according to [the enhancement](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/2268-non-graceful-shutdown#proposal.)
      However, SNR works well with this taint as we discussed in https://github.com/medik8s/self-node-remediation/issues/17#issuecomment-1155947689.
5. If the failed node is rebooted and there is no terminating pods and volumeattachements, SNR deletes the out-of-service taint.
=> SNR waits for deleting workloads on the failed node with a timer. If the timer is expired, the reconcile method is triggered. 
6. The fencing complete is set
7. After setting the fencing complete, it starts cleaning up and then the NoExecute taint is deleted and the failed node becomes schedulable again
8. The remediation is completed.

### DIFF
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	ResourceDeletionRemediationStrategy = RemediationStrategyType("ResourceDeletion")
+	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
+	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
 	// SnrConditionProcessing is the condition type used to signal NHC the remediation status
 	SnrConditionProcessing = "Processing"
 )
@@ -33,11 +34,13 @@ type RemediationStrategyType string
 
 // SelfNodeRemediationSpec defines the desired state of SelfNodeRemediation
 type SelfNodeRemediationSpec struct {
-	//RemediationStrategy is the remediation method for unhealthy nodes
-	//currently "NodeDeletion" is deprecated and "ResourceDeletion" will always happen, regardless of which strategy is selected
-	//it will iterate over all pods and volume attachments related to the unhealthy node and delete them
+	//RemediationStrategy is the remediation method for unhealthy nodes.
+	//Currently "NodeDeletion" is deprecated and it could be either "ResourceDeletion" or "OutOfServiceTaint".
+	//The first will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
+	//The latter will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
+	//that enables automatic deletion of pv-attached pods on failed nodes.
 	// +kubebuilder:default:="ResourceDeletion"
-	// +kubebuilder:validation:Enum=ResourceDeletion
+	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
 	RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
 }
 
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -225,6 +225,75 @@ func (r *SelfNodeRemediationReconciler) isFencingCompleted(snr *v1alpha1.SelfNod
 }
 
 func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
+	return r.remediateWithResourceRemoval(snr, r.deleteResourcesWrapper)
+}
+
+// deleteResourcesWrapper returns a 'zero' time and nil if it completes to delete node resources successfully
+// if not, it will return a 'zero' time and non-nil error, which means exponential backoff is triggered
+// snr is only used in order to match method signature required by remediateWithResourceRemoval
+func (r *SelfNodeRemediationReconciler) deleteResourcesWrapper(node *v1.Node, snr *v1alpha1.SelfNodeRemediation) (time.Duration, error) {
+	return 0, r.deleteResources(node)
+}
+
+func (r *SelfNodeRemediationReconciler) deleteResources(node *v1.Node) error {
+	//fence
+	zero := int64(0)
+	backgroundDeletePolicy := metav1.DeletePropagationBackground
+
+	deleteOptions := &client.DeleteAllOfOptions{
+		ListOptions: client.ListOptions{
+			FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": node.Name}),
+			Namespace:     "",
+			Limit:         0,
+		},
+		DeleteOptions: client.DeleteOptions{
+			GracePeriodSeconds: &zero,
+			PropagationPolicy:  &backgroundDeletePolicy,
+		},
+	}
+
+	namespaces := v1.NamespaceList{}
+	if err := r.Client.List(context.Background(), &namespaces); err != nil {
+		r.logger.Error(err, "failed to list namespaces", err)
+		return err
+	}
+
+	r.logger.Info("starting to delete node resources", "node name", node.Name)
+
+	pod := &v1.Pod{}
+	for _, ns := range namespaces.Items {
+		deleteOptions.Namespace = ns.Name
+		err := r.Client.DeleteAllOf(context.Background(), pod, deleteOptions)
+		if err != nil {
+			r.logger.Error(err, "failed to delete pods of unhealthy node", "namespace", ns.Name)
+			return err
+		}
+	}
+
+	volumeAttachments := &storagev1.VolumeAttachmentList{}
+	if err := r.Client.List(context.Background(), volumeAttachments); err != nil {
+		r.logger.Error(err, "failed to get volumeAttachments list")
+		return err
+	}
+	forceDeleteOption := &client.DeleteOptions{
+		GracePeriodSeconds: &zero,
+	}
+	for _, va := range volumeAttachments.Items {
+		if va.Spec.NodeName == node.Name {
+			err := r.Client.Delete(context.Background(), &va, forceDeleteOption)
+			if err != nil {
+				r.logger.Error(err, "failed to delete volumeAttachment", "name", va.Name)
+				return err
+			}
+		}
+	}
+
+	r.logger.Info("done deleting node resources", "node name", node.Name)
+
+	return nil
+}
+
+func (r *SelfNodeRemediationReconciler) remediateWithResourceRemoval(snr *v1alpha1.SelfNodeRemediation, removeNodeResources func(*v1.Node, *v1alpha1.SelfNodeRemediation) (time.Duration, error)) (ctrl.Result, error) {
 	node, err := r.getNodeFromSnr(snr)
 	if err != nil {
 		r.logger.Error(err, "failed to get node", "node name", snr.Name)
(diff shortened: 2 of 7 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -311,6 +321,116 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
...
+		}
+
+		if controllerutil.ContainsFinalizer(snr, SNRFinalizer) {
+			if err := r.removeFinalizer(snr); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+
+		return ctrl.Result{}, nil
+	}
```
mshitrit, line 378: I think this code is 95% duplicated, I suggest refactor it into a method
Something like:
```
if r.isFencingCompleted(snr) {
  return postFencingLogic(node, r.isResourceDeletionCompleted)
}
  ```

File: controllers/selfnoderemediation_controller.go [hunk 2]
```diff
@@ -311,6 +321,116 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
...
+		// we have a problem on this node, reboot!
+		return r.rebootIfNeeded(snr)
+	}
+
+	wasRebooted, timeLeft := r.wasNodeRebooted(snr)
+	if !wasRebooted {
+		return ctrl.Result{RequeueAfter: timeLeft}, nil
+	}
+
+	r.logger.Info("TimeAssumedRebooted is old. The unhealthy node assumed to been rebooted", "node name", node.Name)
```
mshitrit, line 414: I think this part  (380-414) is duplicated, I  suggest refactor into a method.

controllers/selfnoderemediation_controller.go
```diff
@@ -311,6 +321,116 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
...
+
+	if err = r.addOutOfServiceTaint(node); err != nil {
+		return ctrl.Result{}, err
+	}
+
```
mshitrit, line 419: Hmm I think we can consider moving the Verification of the resource deletion here (before setting the fencing complete).
I think it'll serve 2 purposes:
1. it'll be more consistent with how we decide fencing is complete (same as Resource deletion - after resource are deleted)
2. It'll be much easier to refactor the code: `remediateWithResourceDeletion` and `remediateOutOfServiceTaint` they can share the same code expect for the removal method:
```
func (r *SelfNodeRemediationReconciler) remediateOutOfServiceTaint(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
  return remediateWithResourceRemoval(snr, removeByOutOfServiceTaint)
}

func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
   return remediateWithResourceRemoval(snr, removeByResourceDeletion)
}
```

controllers/selfnoderemediation_controller.go
k-keiichi-rh: I did. I completely agree to refactor it to commonize.
And it's moving the verification of the resource deletion before setting the fencing complete and then I refactor it.

controllers/selfnoderemediation_controller.go
k-keiichi-rh: I did it.

controllers/selfnoderemediation_controller.go
k-keiichi-rh: I agree the 2 purpose. Now it's moved here. I updated the current flow in the description.

controllers/selfnoderemediation_controller.go
```diff
@@ -169,7 +177,7 @@ func (r *SelfNodeRemediationReconciler) isFencingCompleted(snr *v1alpha1.SelfNod
 	return snr.Status.Phase != nil && *snr.Status.Phase == fencingCompletedPhase && snr.DeletionTimestamp != nil
 }
 
-func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
+func (r *SelfNodeRemediationReconciler) remediateWithResourceRemoval(snr *v1alpha1.SelfNodeRemediation, removeResource func(*v1.Node, *v1alpha1.SelfNodeRemediation) (time.Duration, error)) (ctrl.Result, error) {
```
mshitrit, line 180: Nit: I think that the name `removeResource` might be a bit confusing.
Maybe something like `removeNodeResources` or even `removeResources` ?

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -34,9 +35,11 @@ type RemediationStrategyType string
 // SelfNodeRemediationSpec defines the desired state of SelfNodeRemediation
 type SelfNodeRemediationSpec struct {
 	//RemediationStrategy is the remediation method for unhealthy nodes
-	//currently "NodeDeletion" is deprecated and "ResourceDeletion" will always happen, regardless of which strategy is selected
-	//it will iterate over all pods and volume attachments related to the unhealthy node and delete them
-	// +kubebuilder:default:="ResourceDeletion"
+	//currently "NodeDeletion" is deprecated and it could be either "ResourceDeletion" or "OutOfServiceTaint"
+	//the first will iterate over all pos and volumeattachments related to the unhealthy node and delete them
```
razo7, line 39: Typos:

- `pos` -> `pods`
- `volumeattachments` -> `VolumeAttachment`

controllers/selfnoderemediation_controller.go
```diff
@@ -243,6 +251,31 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
...
+		}
+		r.logger.Error(err, "failed to mark SNR as fencing completed")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
+	return r.remediateWithResourceRemoval(snr, r.removeResourceByResourceDeletion)
```
mshitrit, line 275: IIUC the method `remediateWithResourceRemoval` only accepts `snr` and returns  a `Duration` because it needs to fulfill the same method signature of `removeResourceByAddingOutOfServiceTaint`.
I think a nicer way to keep that method "cleaner" would be to wrap it with another method.
Either anonymous::
```
func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
	return r.remediateWithResourceRemoval(snr, func(node *v1.Node, snr *v1alpha1.SelfNodeRemediation) (time.Duration, error) {
		return 0, r.removeResourceByResourceDeletion(node)
	})
}
```

Or explicit:
```
func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
	return r.remediateWithResourceRemoval(snr, r.removeResourceByResourceDeletionWrapper)
}

//removeResourceByResourceDeletionWrapper snr is only used in order to match method signature required by remediateWithResourceRemoval
func (r *SelfNodeRemediationReconciler) removeResourceByResourceDeletionWrapper(node *v1.Node, snr *v1alpha1.SelfNodeRemediation) (time.Duration, error) {
	return 0, r.removeResourceByResourceDeletion(node)
}
```

controllers/selfnoderemediation_controller.go
k-keiichi-rh: Changed to removeNodeResources. Thanks.

api/v1alpha1/selfnoderemediation_types.go
k-keiichi-rh: Fixed typos. Thanks.

controllers/selfnoderemediation_controller.go
k-keiichi-rh: Thanks. I agree with you. I took defining a new method.

slintes (changes requested):
I'm having a hard time to follow the remdiation flow. I'd suggest to reorder methods, and maybe rename some of them in order to get rid of too many methods with similar names (remediate/remove/deleteResources)?

e.g.:

Reconcile

remediateWithResourceDeletion
removeNodeResourcesByResourceDeletionWrapper -> deleteResourcesWrapper (not nice that we need this wrapper, but I have no quick better idea)
removeNodeResourcesByResourceDeletion -> deleteResources (much easier and better to understand IMHO)

remediateByAddingOutOfServiceTaint -> remediateWithOutOfServiceTaint (align name with remediateWithResourceDeletion)
removeNodeResourcesByAddingOutOfServiceTaint -> useOutOfServiceTaint  (much easier and better to understand IMHO)

remediateWithResourceRemoval

controllers/selfnoderemediation_controller.go
```diff
@@ -646,3 +766,76 @@ func (r *SelfNodeRemediationReconciler) removeNoExecuteTaint(node *v1.Node) erro
...
+func (r *SelfNodeRemediationReconciler) isResourceDeletionCompleted(node *v1.Node) bool {
+
+	pods := &v1.PodList{}
+	if err := r.Client.List(context.Background(), pods); err != nil {
+		r.logger.Error(err, "failed to get volumeAttachments list")
+		return false
+	}
+	for _, pod := range pods.Items {
+		if pod.Spec.NodeName == node.Name {
+			return false
```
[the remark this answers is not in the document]
k-keiichi-rh: As far as I checked, there are the following cases where pods won't be evicted by adding a common NoExecute taint:

1. The Pod has a matching toleration for the taint.
2. The Pod is managed by a DaemonSet.
3. The Pod has an init container that is in an error or crash loop state.
4. The Pod is a critical Pod.
5. The Pod is a static Pod.
6. The Pod has a deletion timestamp and is in its grace period.

And the out-of-service taint is a kind of a NoExecute taint. The only difference between them is whether the case of 6 is covered or not. The out-of-service taint will do that.
If the taint is added and the node is in shutdown or power off state, the taint triggers forcing delete terminating pods.

In this part, I think it's enough to check if the terminating pods and volumeattachements are deleted.

controllers/selfnoderemediation_controller.go
```diff
@@ -646,3 +766,76 @@ func (r *SelfNodeRemediationReconciler) removeNoExecuteTaint(node *v1.Node) erro
...
+		return err
+	}
+	return nil
+}
+
+func (r *SelfNodeRemediationReconciler) isResourceDeletionCompleted(node *v1.Node) bool {
+
+	pods := &v1.PodList{}
+	if err := r.Client.List(context.Background(), pods); err != nil {
+		r.logger.Error(err, "failed to get volumeAttachments list")
```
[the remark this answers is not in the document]
k-keiichi-rh: I will fix it. Thanks.

slintes:
thanks a lot for renaming / reordering, much better readable now IMHO 🙂 

/ok-to-test

controllers/selfnoderemediation_controller.go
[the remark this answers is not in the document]
slintes: as discussed in the meeting, looking for terminating pods only will be ok 👍🏼

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -32,10 +33,12 @@ type RemediationStrategyType string
 // SelfNodeRemediationSpec defines the desired state of SelfNodeRemediation
 type SelfNodeRemediationSpec struct {
 	//RemediationStrategy is the remediation method for unhealthy nodes
-	//currently "NodeDeletion" is deprecated and "ResourceDeletion" will always happen, regardless of which strategy is selected
-	//it will iterate over all pods and volume attachments related to the unhealthy node and delete them
+	//currently "NodeDeletion" is deprecated and it could be either "ResourceDeletion" or "OutOfServiceTaint"
```
slintes, line 36: Please add fullstops and start new sentences uppercase, because new lines are stripped, which make this hard to read in the yaml.

slintes (changes requested):
One nit inside, otherwise lgtm. Needs rebase though.

api/v1alpha1/selfnoderemediation_types.go
k-keiichi-rh: Thanks. I fix it.

slintes:
Looks good to me, nice work!
Leaving the final review to @mshitrit 

/retest

slintes (commented):
Looks good to me, nice work!
Leaving the final review to @mshitrit

slintes:
couldn't get rid of the "change requested" state in another way 🤷🏼‍♂️ 

/lgtm cancel

PR #88, opened 2023-02-02
Remove NodeDeletion Strategy

mshitrit:
[ECOPROJECT-1074](https://issues.redhat.com//browse/ECOPROJECT-1074) Last Phase of Removing Node Deletion Strategy, making sure it does not exist by by OLM enum validation


clobrano (commented):
/lgtm
/hold

slintes (changes requested):
please remove the webhook reference for the templates from the PROJECT file as well

PROJECT
```diff
@@ -36,7 +33,4 @@ resources:
   kind: SelfNodeRemediationConfig
   path: github.com/medik8s/self-node-remediation/api/v1alpha1
   version: v1alpha1
-  webhooks:
```
slintes, line 39: the config still has a webhook...

slintes (changes requested):

PROJECT
mshitrit: Nope, it's marked in red to show it's removed.

slintes: let me rephrase: we still have a webhook for the SelfNodeRemediationConfig CRD, and this is its webhook config, which should not be deleted :)

PR #89, opened 2023-02-07
Stop remediation if nhc timeout annotation is set

mshitrit:


clobrano:
/lgtm
/hold

PR #90, opened 2023-02-09
Add "protoc" as dependency for "check" target

slintes:
Download the protoc binary first, because
the golang image misses the needed "unzip" tool

This blocks NHC CI

Reproducer without this PR: delete ./bin, run `make check`


clobrano:
/lgtm
/hold

PR #91, opened 2023-02-12
Reduce timeout for compact (or otherwise small) clusters

mshitrit:
[ECOPROJECT-1197](https://issues.redhat.com//browse/ECOPROJECT-1197)
ATM when we calculate SNR timeout we are sampling all the worker nodes in batches of 10% of the nodes in each batch.
Therefore we calculate the timeout assuming 10 batches.
In case a cluster has fewer than 10 worker nodes this calculation can be optimized.

### DIFF
--- a/main.go
+++ b/main.go
@@ -241,7 +241,7 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 	peerRequestTimeout := getDurEnvVarOrDie("PEER_REQUEST_TIMEOUT")   //timeout for each peer request
 	timeToAssumeNodeRebooted := getDurEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
 
-	safeRebootCalc := reboot.New(mgr.GetClient(), wd, maxErrorThreshold, apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout, timeToAssumeNodeRebooted)
+	safeRebootCalc := reboot.NewSafeTimeCalculator(mgr.GetClient(), wd, maxErrorThreshold, apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout, timeToAssumeNodeRebooted)
 	if err = mgr.Add(safeRebootCalc); err != nil {
 		setupLog.Error(err, "failed to add safe reboot time calculator to the manager")
 		os.Exit(1)
--- a/main.go
+++ b/main.go
@@ -54,9 +54,8 @@ import (
 )
 
 const (
-	nodeNameEnvVar            = "MY_NODE_NAME"
-	peerHealthDefaultPort     = 30001
-	maxTimeForNoPeersResponse = 30 * time.Second
+	nodeNameEnvVar        = "MY_NODE_NAME"
+	peerHealthDefaultPort = 30001
 )
 
 var (
@@ -224,9 +223,6 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 		os.Exit(1)
 	}
 
-	// it's fine when the watchdog is nil!
-	rebooter := reboot.NewWatchdogRebooter(wd, ctrl.Log.WithName("rebooter"))
-
 	// TODO make the interval configurable
 	peerUpdateInterval := getDurEnvVarOrDie("PEER_UPDATE_INTERVAL")
 	peerApiServerTimeout := getDurEnvVarOrDie("PEER_API_SERVER_TIMEOUT")
@@ -243,6 +239,16 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 	apiServerTimeout := getDurEnvVarOrDie("API_SERVER_TIMEOUT")       //timeout for each api-connectivity check
 	peerDialTimeout := getDurEnvVarOrDie("PEER_DIAL_TIMEOUT")         //timeout for establishing connection to peer
 	peerRequestTimeout := getDurEnvVarOrDie("PEER_REQUEST_TIMEOUT")   //timeout for each peer request
+	timeToAssumeNodeRebooted := getDurEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
+
+	safeRebootCalc := reboot.New(mgr.GetClient(), wd, maxErrorThreshold, apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout, timeToAssumeNodeRebooted)
+	if err = mgr.Add(safeRebootCalc); err != nil {
+		setupLog.Error(err, "failed to add safe reboot time calculator to the manager")
+		os.Exit(1)
+	}
+
+	// it's fine when the watchdog is nil!
+	rebooter := reboot.NewWatchdogRebooter(wd, ctrl.Log.WithName("rebooter"), safeRebootCalc)
 
 	// init certificate reader
 	certReader := certificates.NewSecretCertStorage(mgr.GetClient(), ctrl.Log.WithName("SecretCertStorage"), ns)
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -325,3 +325,15 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 		os.Exit(1)
 	}
 }
+
+func calcNumOfBatches(myPeers *peers.Peers) time.Duration {
+	// 2. time for asking peers (10% batches + 1st smaller batch)
+	numberOfBatches := 10
+	numOfWorkerPeers := len(myPeers.GetPeersAddresses(peers.Worker))
```
slintes, line 332: this will always return 0, because Start() wasn't called yet

slintes (changes requested):

main.go
mshitrit: Thanks, Great catch

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,101 @@
...
+type safeTimeCalculator struct {
+	timeToAssumeNodeRebooted                                                time.Duration
+	wd                                                                      watchdog.Watchdog
+	maxErrorThreshold                                                       int
+	apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout time.Duration
+	log                                                                     logr.Logger
+	k8sClient                                                               client.Client
+}
+
+func New(k8sClient client.Client, wd watchdog.Watchdog, maxErrorThreshold int, apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout, timeToAssumeNodeRebooted time.Duration) SafeTimeCalculator {
```
razo7, line 38: IMO `New` is to generic. Maybe `NewCalculater`?

slintes: agree. Just `New` is fine if the thing which is created matches the package name, which isn't the case here. I would even go with `NewSafeTimeCalculator`

slintes (changes requested):
Good idea to put the calculation in its own component. Doing the calculation on start only doesn't work though, because the number of nodes in the cluster can change any time.

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+//goland:noinspection GoUnusedParameter
+func (s *safeTimeCalculator) Start(ctx context.Context) error {
+	s.calcMinTimeAssumeRebooted()
+	return nil
+}
+
+func (s *safeTimeCalculator) calcMinTimeAssumeRebooted() {
+	// but the reboot time needs be at least the time we know we need for determining a node issue and trigger the reboot!
```
clobrano, line 68: Is the initial "but" intentional?
How about something like "Reboot time needs to be >= to the time needed to determine a node issue and trigger the reboot" ?

mshitrit: I'm not sure  IIRC this comment is a copy paste from main, I agree phrasing can be better

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+	// time for asking peers (10% batches + 1st smaller batch)
+	numberOfBatches := 10
+	if err := s.k8sClient.List(context.Background(), nodes, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		s.log.Error(err, "couldn't fetch worker nodes")
+		return time.Duration(numberOfBatches)
+	}
+
+	//in case there are less than 10 worker peers we can cover all of them with less than batches
+	if workerNodesCount := len(nodes.Items); workerNodesCount > 0 && workerNodesCount < numberOfBatches {
+		numberOfBatches = workerNodesCount
```
slintes, line 99: this isn't correct. The first batch always takes 3 nodes already (check.go line 145)

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	numberOfBatches = numberOfBatches + 1
```
slintes, line 102: what's this about? 🤔

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+	if s.wd != nil {
+		minTime += s.wd.GetTimeout()
+	}
+	// 4. some buffer
+	minTime += 15 * time.Second
+	s.log.Info("calculated minTimeToAssumeNodeRebooted is:", "minTimeToAssumeNodeRebooted", minTime)
+	s.minTimeToAssumeNodeRebooted = minTime
+}
+
+func (s *safeTimeCalculator) calcNumOfBatches() time.Duration {
```
slintes, line 83: why is this returning a Duration? NumOfBatches should be a number, not? 🤔

slintes (changes requested):
Left some comments.

I'm also wondering if we can't speed up remediation even better by improving the batch sizes in check.go for a small clusters? 10% of e.g. 20 nodes just is 2 nodes per batch. Shouldn't be a problem to do more, so we need less than 11 batches also for clusters up to e.g. 50 nodes

pkg/reboot/calculator.go
mshitrit: Adding the [extra one ](https://github.com/medik8s/self-node-remediation/pull/91#discussion_r1119892685)after the calculation

pkg/reboot/calculator.go
mshitrit: Just wanted to save the casting that would need to be done higher up (i.e 
instead of currently: 
```
minTime += s.calcNumOfBatches() * (s.peerDialTimeout + s.peerRequestTimeout)
```

it would be:
```
minTime += time.Duration( s.calcNumOfBatches()) * (s.peerDialTimeout + s.peerRequestTimeout)
```

I don't really like either, let me know if you have a better idea

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (s *safeTimeCalculator) GetTimeToAssumeNodeRebooted() time.Duration {
+	s.calcMinTimeAssumeRebooted()
+	if s.timeToAssumeNodeRebooted < s.minTimeToAssumeNodeRebooted {
+		return s.minTimeToAssumeNodeRebooted
+	}
+	return s.timeToAssumeNodeRebooted
+}
+
+//goland:noinspection GoUnusedParameter
```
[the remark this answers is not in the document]
mshitrit: Golang is perfectly happy with `_ context.Context` :smile:

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,108 @@
...
+func (s *safeTimeCalculator) Start(ctx context.Context) error {
+	s.calcMinTimeAssumeRebooted()
```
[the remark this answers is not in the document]
mshitrit: I feel that it's a bit safer to use that,  I return the highest value of batch Number ever calculated.
Doing this calculation early on it's very likely that all the worker nodes will be working properly.

mshitrit:
>I'm also wondering if we can't speed up remediation even better by improving the batch sizes in check.go for a small clusters? 10% of e.g. 20 nodes just is 2 nodes per batch. Shouldn't be a problem to do more, so we need less than 11 batches also for clusters up to e.g. 50 nodes, and for very small clusters just 1 or 2 batches.

Probably worth looking into, but I think I'll leave it for another PR

pkg/reboot/calculator.go
slintes: I think it makes more sense to do the cast where it's needed, here it's confusing. Thanks for changing it 🙂

pkg/reboot/calculator.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := s.k8sClient.List(context.Background(), nodes, client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		s.log.Error(err, "couldn't fetch worker nodes")
+		return maxNumberOfBatches
+	}
+	workerNodesCount := len(nodes.Items)
+
+	var numberOfBatches int
+	switch {
+	// 3 workers of first batch + 10% for each batch
+	case workerNodesCount >= 13:
```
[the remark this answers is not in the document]
mshitrit: Yep, Good idea

slintes (approved):

PR #92, opened 2023-02-17
Avoid SelfNodeRemediation object being deleted before remediation is complete.

epenchev:
This tries to fix an existing race condition when a SelfNodeRemediation object is created and immediately scheduled for 'Foreground' deletion from K8s.

When a SelfNodeRemediation object is created the object has the 'ownerReferences' field set, and this immediately schedules it for deletion.
If a 'finalizer' is not set immediately the object gets deleted and no remediation is triggered for the Node.
The idea for the fix is the object to have a 'finalizer' set right after a reconcile request is triggered.
Although this might not be the best solution for the problem!

This happens only once when the cluster is initially started.
Every subsequent attempt for 'remediation' is successful afterwards.
After a reconcile request is received `isNodeRebootCapable()` is checked, which itself lists all the Pods from the DS and at only first run controller-runtime library cache is still empty.
During the time that Pod information is fetched from the K8s API server and not from the client's cache, the 'SelfNodeRemediation' object is already deleted without the protection of the 'finalizer' set right afterwards.

Typical logs with timestamps when the issue is hit.
```
2023-02-17T09:34:16.515Z INFO controller-runtime.metrics metrics server is starting to listen {"addr": ":8080"}
2023-02-17T09:34:16.517Z INFO setup Starting as a self node remediation agent that should run as part of the daemonset
2023-02-17T09:34:16.543Z INFO setup Time to assume that unhealthy node has been rebooted {"time": "3m5s"}
2023-02-17T09:34:16.544Z INFO setup init grpc server
2023-02-17T09:34:16.544Z INFO setup starting manager
2023-02-17T09:34:16.544Z INFO controller-runtime.manager.controller.selfnoderemediation Starting EventSource {"reconciler group": "self-node-remediation.medik8s.io", "reconciler kind": "SelfNodeRemediation", "source": "kind source: /, Kind="}
2023-02-17T09:34:16.544Z INFO controller-runtime.manager.controller.selfnoderemediation Starting Controller {"reconciler group": "self-node-remediation.medik8s.io", "reconciler kind": "SelfNodeRemediation"}
2023-02-17T09:34:16.544Z INFO watchdog watchdog started
2023-02-17T09:34:16.545Z INFO api-check api connectivity check started
2023-02-17T09:34:16.544Z INFO controller-runtime.manager starting metrics server {"path": "/metrics"}
2023-02-17T09:34:16.645Z INFO controller-runtime.manager.controller.selfnoderemediation Starting workers {"reconciler group": "self-node-remediation.medik8s.io", "reconciler kind": "SelfNodeRemediation", "worker count": 1}
2023-02-17T09:34:16.645Z INFO peers peers started
2023-02-17T09:34:16.645Z INFO peerhealth.server peer health server started
2023-02-17T09:36:36.649Z INFO controllers.SelfNodeRemediation fencing not completed yet, continuing remediation {"selfnoderemediation": "self-node-remediation/k8s-node6"}
2023-02-17T09:36:36.756Z INFO controllers.SelfNodeRemediation SNR already deleted {"selfnoderemediation": "self-node-remediation/k8s-node6"}
2023-02-17T09:36:37.759Z INFO controllers.SelfNodeRemediation SNR already deleted {"selfnoderemediation": "self-node-remediation/k8s-node6"} 
```
And of course, no remediation hasn't been triggered at all.

A way to reproduce it.
- Reload all the self-node-remediation operator DS Pods so the cache is lost.
- Execute 
```
cat <<EOF | kubectl apply -f -
apiVersion: self-node-remediation.medik8s.io/v1alpha1
kind: SelfNodeRemediation
metadata:
  name: k8s-node6
  namespace: self-node-remediation
  ownerReferences:
  - apiVersion: remediation.medik8s.io/v1alpha1
    controller: false
    kind: NodeHealthCheck
    name: nodehealthcheck-sample
    uid: fe293bb4-ce1e-4b01-b6f4-7f6f08ec6958
spec:
  remediationStrategy: ResourceDeletion
EOF
```

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -213,15 +213,15 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
 		return ctrl.Result{}, nil
 	}
 
+	if !controllerutil.ContainsFinalizer(snr, SNRFinalizer) {
+		return r.addFinalizer(snr)
+	}
+
 	r.logger.Info("fencing not completed yet, continuing remediation")
 
 	if !r.isNodeRebootCapable(node) {
 		//use err to trigger exponential backoff
 		return ctrl.Result{}, errors.New("Node is not capable to reboot itself")
-	}
-
-	if !controllerutil.ContainsFinalizer(snr, SNRFinalizer) {
-		return r.addFinalizer(snr)
 	}
 
 	if err = r.addNoExecuteTaint(node); err != nil {


slintes:
Ah, from the comments on the other PR I understand you use SNR standalone, without NHC? And then you manually create the remediation CR? In that case, just don't set the owner reference 🙂

epenchev:
> Ah, from the comments on the other PR I understand you use SNR standalone, without NHC? And then you manually create the remediation CR? In that case, just don't set the owner reference slightly_smiling_face Or if another operator is creating the remediation CR, set the owner ref to something belonging to that other operator.

Yes exactly, I'm using SNR standalone, without NHC like you pointed out.
I have a custom operator that creates the remediation CR and sets the owner ref to its object that's belonging to it.
Eventually this triggers the deletion of the remediation CR while the owner ref object being still active and not deleted or scheduled to be deleted.
The one thing that I can think of is that the remediation CR and the instance of the custom object type I'm using are placed in different namespaces.

I already added the workaround by not adding a owner reference until there is a 'finalizer' set for the remediation CR like you have suggested. This solution worked for me.

slintes:
from https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

> Cross-namespace owner references are disallowed by design. Namespaced dependents can specify cluster-scoped or namespaced owners. A namespaced owner must exist in the same namespace as the dependent. If it does not, the owner reference is treated as absent, and the dependent is subject to deletion once all owners are verified absent.

I'd suggest to create the SNR CR in the same namespace where your custom object is located. SNR is watching all namespaces 🙂 That way you don't need the workaround of setting the owner ref after the finalizer was set.

For NHC this isn't an issue, because NHC CRs are cluster scoped.

Closing this as not a bug, but feel free to ask further questions of course!

/close

epenchev:
> I'd suggest to create the SNR CR in the same namespace where your custom object is located. SNR is watching all namespaces slightly_smiling_face That way you don't need the workaround of setting the owner ref after the finalizer was set.
> 
> For NHC this isn't an issue, because NHC CRs are cluster scoped.
> 
Thanks for the advice, much appreciated.

PR #94, opened 2023-02-21
Support "Processing" condition on SNR

mshitrit:
[ECOPROJECT-1198](https://issues.redhat.com//browse/ECOPROJECT-1198)


controllers/selfnoderemediation_controller.go
```diff
@@ -164,8 +181,51 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	return result, r.updateSnrStatusLastError(snr, err)
 }
 
+func (r *SelfNodeRemediationReconciler) updateSnrProcessingCondition(reason processingChangeReason, snr *v1alpha1.SelfNodeRemediation, isPatchDone bool) error {
```
clobrano, line 184: The signature of this function makes it a bit cryptic when used. I need to look at its definition to know why it's `updateSnrProcessingCondition(..., true)` or `updateSnrProcessingCondition(..., false)`.
It seems it's doing two things: update the processing condition AND (sometimes) patching SNR status, what do you think about having `updateSnrProcessingCondition` and `updateSnrProcessingConditionAndPatch` which calls the first internally?

clobrano: Also, the `DeepCopy` is always done regardless it needs to patch or not.

mshitrit: TBH I don't have a strong preference but I see your  point.
I think that if going in that direction a nicer solution would be just to separate the patch from the status change.
let me know if you are ok with this change.

controllers/selfnoderemediation_controller.go
```diff
@@ -142,7 +152,22 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
...
-		return ctrl.Result{}, nil
+		org := snr.DeepCopy()
+		if err := r.updateSnrProcessingCondition(remediationTerminatedByNHC, snr); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, r.patchSnrStatus(snr, org)
+	}
+
+	//In case snr is being deleted there is no remediation in process
+	if !r.isFencingCompleted(snr) {
```
slintes, line 163: I think this is the wrong place for setting the condition to true. There are some checks following, which will prevent an actual remediation. IMHO the condition should be set to true when at the place where we set the finalizer, and in all other cases to false. That would ensure that NHC tries the next remediation quickly. WDYT?

controllers/selfnoderemediation_controller.go
```diff
@@ -142,7 +152,22 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 
 	if r.isStoppedByNHC(snr) {
 		r.logger.Info("SNR remediation was stopped by Node Healthcheck")
-		return ctrl.Result{}, nil
+		org := snr.DeepCopy()
+		if err := r.updateSnrProcessingCondition(remediationTerminatedByNHC, snr); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, r.patchSnrStatus(snr, org)
```
slintes, line 159: for a later improvement: patchSnrStatus is called more and more often. I'd suggest to automatically patch the status with by deferred call, like NHC does.

slintes (changes requested):

slintes:
just curious if SNR runs into same CI issues as NHC

/test 4.13-openshift-e2e

controllers/selfnoderemediation_controller.go
mshitrit: The way I understand it setting the flag to true is merely an indication that SNR has started remediating, it is by no means an assurance that the remediation will be successful.
It is true that setting the flag on a later stage might save time in NHC escalation use cases, but I think it would not reflect the current situation correctly (hiding information both from the user and from NHC).

I am not sure but  by setting the flag later don't  we run into the opposite risk of NHC assuming SNR is prematurely ?

slintes: as discussed on chat, we might want to revisit this in a follow up

mshitrit: Adding a [link](https://redhat-internal.slack.com/archives/C03M5GKJNBA/p1679306209179739) to the slack chat for future me

PR #95, opened 2023-03-01
Remove upper bound of olm skip range in non release versions (main)

mshitrit:


clobrano:
/lgtm
I wonder why the `opm` error says "more than one bundle with maximum version x.y.z". What are the bundles created?
  
/hold

PR #96, opened 2023-03-05
increase the minimum batch size

mshitrit:
Followup PR for [this](https://github.com/medik8s/self-node-remediation/pull/91#pullrequestreview-1317265132) comment.
increase the minimum number of worker nodes in each batch from 1 to 3 in order to reduce timeout in clusters with low number of worker nodes.

### DIFF
--- a/pkg/reboot/calculator.go
+++ b/pkg/reboot/calculator.go
@@ -98,7 +98,7 @@ func (s *safeTimeCalculator) calcNumOfBatches() int {
 
 	var numberOfBatches int
 	switch {
-	//high number of workers we need max batches (for example 53 nodes will be done in 11 batches -> 1 * 3 + 10 * 5 )
+	//high number of workers: we need max batches (for example 53 nodes will be done in 11 batches -> 1 * 3 + 10 * 5 )
 	case workerNodesCount > maxNumberOfBatches*MinNodesNumberInBatch:
 		numberOfBatches = maxNumberOfBatches
 	//there are few enough nodes to use the min batch (for example 20 nodes will be done in 7 batches -> 1 * 3 +  6 * 3 )
--- a/pkg/reboot/calculator.go
+++ b/pkg/reboot/calculator.go
@@ -19,8 +19,8 @@ import (
 
 const (
 	MaxTimeForNoPeersResponse = 30 * time.Second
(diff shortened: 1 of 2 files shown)


pkg/reboot/calculator.go
```diff
@@ -19,8 +19,8 @@ import (
 
 const (
 	MaxTimeForNoPeersResponse = 30 * time.Second
-	NodesNumberInFirstBatch = 3
-	MaxBatchesAfterFirst    = 10
+	MinNodesNumberInBatch     = 3
+	MaxBatchesAfterFirst      = 10
```
razo7, line 23: `MaxBatchesAfterFirst`? After first batch? I don't understand the variable name. 
Can it just be `MaxBatches`?

pkg/reboot/calculator.go
```diff
@@ -98,16 +98,15 @@ func (s *safeTimeCalculator) calcNumOfBatches() int {
 
 	var numberOfBatches int
 	switch {
-	// 3 workers of first batch + 10% for each batch
-	case workerNodesCount >= MaxBatchesAfterFirst+NodesNumberInFirstBatch:
+	//high number of workers we need max batches (for example 53 nodes will be done in 11 batches -> 1 * 3 + 10 * 5 )
```
razo7, line 101: `high number of workers we need max batches` -> `for high number of workers we need max batches`

pkg/reboot/calculator.go
mshitrit: First batch is always of 3 nodes. this parameter refers to the number of batches that follow the first one.
For example for 53 nodes we'd have 1 batch of 3 nodes and 10 more batches of 5 nodes each, so a total of 53 nodes in 11 batches .

pkg/reboot/calculator.go
clobrano: I'd also rephrase this a little, or even just `high number of workers: we need max batches`

beekhof:
/lgtm

looking forward to the follow up that increases it further :)

clobrano:
/lgtm

/hold

mshitrit:
4.13 passed, overriding flaky tests
/override ci/prow/4.11-openshift-e2e 
/override ci/prow/4.12-openshift-e2e 
/unhold

PR #97, opened 2023-03-11
Skip SNR pod check after node is rebooted

k-keiichi-rh:
Now the remediation is stuck when there is no SNR pod on the unhealthy node after it's rebooted.
Because isNodeRebootCapable() always tries to check if the SNR pod is there on the unhealthy node.
If the SNR pod can't restart, resources won't be deleted after rebooting the unhealthy node. This is the issue this PR fixes.

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -45,9 +45,8 @@ import (
 )
 
 const (
-	SNRFinalizer          = "self-node-remediation.medik8s.io/snr-finalizer"
-	fencingCompletedPhase = "Fencing-Completed"
-	nhcTimeOutAnnotation  = "remediation.medik8s.io/nhc-timed-out"
+	SNRFinalizer         = "self-node-remediation.medik8s.io/snr-finalizer"
+	nhcTimeOutAnnotation = "remediation.medik8s.io/nhc-timed-out"
 )
 
 var (
@@ -74,10 +73,20 @@ var (
 
 type processingChangeReason string
 
-var (
+const (
 	remediationStarted         processingChangeReason = "RemediationStarted"
 	remediationTerminatedByNHC processingChangeReason = "RemediationStoppedByNHC"
 	remediationFinished        processingChangeReason = "RemediationFinished"
+)
+
+type remediationPhase string
+
+const (
+	fencingStartedPhase     remediationPhase = "Fencing-Started"
+	preRebootCompletedPhase remediationPhase = "Pre-Reboot-Completed"
+	rebootCompletedPhase    remediationPhase = "Reboot-Completed"
+	fencingCompletedPhase   remediationPhase = "Fencing-Completed"
+	unknownPhase            remediationPhase = "Unknown"
 )
 
 type UnreconcilableError struct {
@@ -170,7 +179,7 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, r.updateSnrProcessingCondition(remediationTerminatedByNHC, snr)
 	}
 
-	if !r.isFencingCompleted(snr) {
+	if r.getPhase(snr) != fencingCompletedPhase {
 		if err := r.updateSnrProcessingCondition(remediationStarted, snr); err != nil {
 			return ctrl.Result{}, err
 		}
@@ -233,8 +242,17 @@ func (r *SelfNodeRemediationReconciler) patchSnrStatus(ctx context.Context, chan
 	return nil
 }
 
-func (r *SelfNodeRemediationReconciler) isFencingCompleted(snr *v1alpha1.SelfNodeRemediation) bool {
-	return snr.Status.Phase != nil && *snr.Status.Phase == fencingCompletedPhase
+func (r *SelfNodeRemediationReconciler) getPhase(snr *v1alpha1.SelfNodeRemediation) remediationPhase {
+	if snr.Status.Phase == nil {
+		return fencingStartedPhase
+	}
+	phase := remediationPhase(*snr.Status.Phase)
+	switch phase {
+	case preRebootCompletedPhase, rebootCompletedPhase, fencingCompletedPhase:
+		return phase
+	default:
+		return unknownPhase
+	}
 }
 
 func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alpha1.SelfNodeRemediation) (ctrl.Result, error) {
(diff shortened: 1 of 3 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -44,6 +44,7 @@ import (
 const (
 	SNRFinalizer          = "self-node-remediation.medik8s.io/snr-finalizer"
 	fencingCompletedPhase = "Fencing-Completed"
+	rebootCompletedPhase  = "Reboot-Completed"
```
mshitrit, line 47: Nit: I'd consider turning into a new type and replace both `is` methods with a single generic one.
```
type remediationPhase string

var (
	fencingCompletedPhase remediationPhase = "Fencing-Completed"
	rebootCompletedPhase  remediationPhase = "Reboot-Completed"
)

func (r *SelfNodeRemediationReconciler) isAtRemediationPhase(snr *v1alpha1.SelfNodeRemediation, phase remediationPhase) bool {
	return snr.Status.Phase != nil && *snr.Status.Phase == string(phase)
}
```

slintes:
in general the code looks good to me... but I'm wondering if it would be preferable to use conditions instead of phase?
Context: https://github.com/kubernetes/kubernetes/issues/7856

controllers/selfnoderemediation_controller.go
```diff
@@ -161,8 +169,33 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		return err
+	}
+	return nil
+}
+
+func (r *SelfNodeRemediationReconciler) getPhase(snr *v1alpha1.SelfNodeRemediation) remediationPhase {
+	if snr.Status.Phase == nil {
+		return fencingStartedPhase
+	}
+	return remediationPhase(*snr.Status.Phase)
```
mshitrit, line 194: I think we should verify the phase is valid.
At the moment this will work for any string.

controllers/selfnoderemediation_controller.go
```diff
@@ -293,18 +338,7 @@ func (r *SelfNodeRemediationReconciler) remediateWithResourceDeletion(snr *v1alp
...
-		if apiErrors.IsConflict(err) {
-			// conflicts are expected since all self node remediation deamonset pods are competing on the same requests
-			return ctrl.Result{RequeueAfter: 1 * time.Second}, nil
-		}
-		r.logger.Error(err, "failed to mark SNR as fencing completed")
-		return ctrl.Result{}, err
-	}
-
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, r.updateSnrRemediationPhase(fencingCompletedPhase, snr)
```
mshitrit, line 341: At the moment we don't explicitly use rebootCompletedPhase (except from setting it).
IMO the code will be clearer by conditioning the resource deletion part to be at that stage.

mshitrit:
> in general the code looks good to me... but I'm wondering if it would be preferable to use conditions instead of phase? Context: [kubernetes/kubernetes#7856](https://github.com/kubernetes/kubernetes/issues/7856)

No objection on my part to explore conditions.

However in case we are sticking to current format I suggest adjustment as below.
I think it'll make the code clearer, easier to refactor and simplify handling of unknown phases.
```
	phase := r.getPhase(snr)
	switch phase {
	case fencingCompletedPhase:
	//Some Code
	case fencingStartedPhase:
		//Some Code
	case preRebootCompletedPhase:
		// Some Code
	//more cases
	//
	default:
		//Error handling for unknown phase
	} 
```

slintes:
Can you address Michael's comments please and rebase?
Otherwise lgtm

k-keiichi-rh:
Finally, I followed Michael's comments like:
```
        phase := r.getPhase(snr)
        switch phase {
        case fencingStartedPhase:
                result, err = r.prepareReboot(node, snr)
        case preRebootCompletedPhase:
                result, err = r.rebootNode(node, snr)
        case rebootCompletedPhase:
                result, err = r.removeNodeResources(node, snr)
        case fencingCompletedPhase:
                if snr.DeletionTimestamp != nil {
                        result, err = r.recoverNode(node, snr)
                }
        default:
                //this should never happen since we enforce valid values with kubebuilder
                err = errors.New("unknown phase")
                r.logger.Error(err, "Undefined unknown phase", "phase", phase)
        }
```

However I don't rush. So I want to have this patch in the next release. I need your review.

controllers/selfnoderemediation_controller.go.rej
```diff
@@ -0,0 +1,14 @@
+--- controllers/selfnoderemediation_controller.go
```
slintes, line 1: please delete this file :)

controllers/selfnoderemediation_controller.go
```diff
@@ -156,7 +165,7 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, r.patchSnrStatus(snr, org)
 	}
 
-	if !r.isFencingCompleted(snr) {
+	if !r.isAtRemediationPhase(snr, fencingCompletedPhase) {
```
mshitrit, line 168: Nit: I think `isAtRemediationPhase` can be deleted due to latest changes.
You can use instead:
```
if r.getPhase(snr) != fencingCompletedPhase {
```

controllers/suite_test.go
```diff
@@ -77,13 +79,18 @@ const (
...
+	Reader                  client.Reader
+	ShouldSimulateFailure   bool
+	ShouldSimulateVaFailure bool
 }
 
 func (kcw *K8sClientWrapper) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
 	if kcw.ShouldSimulateFailure {
 		return errors.New("simulation of client error")
+	} else if kcw.ShouldSimulateVaFailure {
+		if reflect.TypeOf(list) == reflect.TypeOf(&storagev1.VolumeAttachmentList{}) {
```
slintes, line 91: just curious, is there an advantage of using reflect instead of trying a cast?
`if _, ok := list.(*storagev1.VolumeAttachmentList); ok {...}`

controllers/selfnoderemediation_controller_test.go
```diff
@@ -272,6 +277,20 @@ func verifyVaDeleted(vaName string) {
 	}, 5*time.Second, 250*time.Millisecond).Should(BeTrue())
 }
 
+func verifyVaNotDeleted(vaName string) {
+	vaKey := client.ObjectKey{
+		Namespace: namespace,
+		Name:      vaName,
+	}
+
+	EventuallyWithOffset(1, func() bool {
```
slintes, line 286: I think when verifying that something isn't deleted, using `Consistently` is a better way than `Eventually`. The latter just checks once and succeeds, when the deletion happens with a short delay. The former will fail as expected.

mshitrit:
@k-keiichi-rh looks like there is a bit of sync issue in the Unit tests, I didn't investigate to closely but I'm guessing that using `Eventually` instead of `Expect` in the creation of the SNR and the VA might help.

k-keiichi-rh:
Thank you fir the suggestion. Sorry for my fault. As far as I checked, it will sometime fail to run this test case because exponential backoff is triggered by the intentional failure of calling k8s api to get volumeattachement list and the timeout of verifyVaDeleted() occurs due to the waiting time of reconcile(). So I need to fix this scenario.

slintes:
there was a green run already on the last commit of this PR, and just #106 was merged in the meanwhile

/override ci/prow/4.11-openshift-e2e

slintes:
@k-keiichi-rh thanks for getting this into shape in time for the upcoming release 👍🏼

PR #98, opened 2023-03-20
remove etcd check from cp diagnostics

mshitrit:
Control Plane diagnostics runs when there is no API connectivity and the control plane seems to be partially isolated.
Due to the nature of etcd it's health check is almost guaranteed to fail under those conditions, making it redundant.


clobrano:
/lgtm

/hold

PR #104, opened 2023-04-02
Validate k8s version for out-of-service-taint remediation strategy

mshitrit:
Verifying out of service taint is supported before allowing the user to set it.

### DIFF
--- a/api/v1alpha1/selfnoderemediationtemplate_webhook.go
+++ b/api/v1alpha1/selfnoderemediationtemplate_webhook.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"fmt"
+	"strconv"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var (
+	log = logf.Log.WithName("selfnoderemediationtemplate-resource")
+
+	//out of service taint strategy params
+	isOutOfServiceTaintSupported             bool
+	minK8sVersionSupportingOutOfServiceTaint = 1.26
+)
+
+func (r *SelfNodeRemediationTemplate) SetupWebhookWithManager(mgr ctrl.Manager) error {
+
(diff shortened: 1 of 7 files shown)


api/v1alpha1/selfnoderemediationtemplate_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var (
+	log = logf.Log.WithName("selfnoderemediationtemplate-resource")
+
+	//out of service taint strategy params
+	isOutOfServiceTaintSupported             bool
+	minK8sVersionSupportingOutOfServiceTaint = 1.26
```
slintes, line 38: nit: this should be a const

api/v1alpha1/selfnoderemediationtemplate_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Error(err, "couldn't get retrieve k8s client")
+		return err
+	} else if version, err := cs.Discovery().ServerVersion(); err != nil || version == nil {
+		if version == nil {
+			err = fmt.Errorf("k8s server version is nil")
+		}
+		log.Error(err, "couldn't get retrieve k8s server version")
+		return err
+	} else {
+		k8sVersion, err := strconv.ParseFloat(fmt.Sprintf("%s.%s", version.Major, version.Minor), 8)
```
slintes, line 96: I'm always afraid of rounding errors when using float... I guess even more with that low bitsize? ~~Maybe validate major and minor each? I think `major > 1 || minor >= 26`~~ Even better, there is a `semver` package 🙂

api/v1alpha1/selfnoderemediationtemplate_webhook_test.go
```diff
@@ -0,0 +1,84 @@
...
+					Expect(outOfServiceStrategy.ValidateCreate()).To(Succeed())
+					Expect(snrtValid.ValidateUpdate(outOfServiceStrategy)).To(Succeed())
+				})
+			})
+			When("out of service taint is not supported", func() {
+				BeforeEach(func() {
+					isOutOfServiceTaintSupported = false
+				})
+				It("should be denied", func() {
+					err := outOfServiceStrategy.ValidateCreate()
```
slintes, line 70: nit: could be reduced to `Expect(outOfServiceStrategy.ValidateCreate()).To(MatchError(ContainSubstring("OutOfServiceTaint remediation strategy is not supported at kubernetes version lower than 1.26, please use a different remediation strategy")))`

slintes (changes requested):
As discussed, the k8s version alone does not guarantee that the feature is enabled. However, this still is valueable.
Some comments inline.

slintes (changes requested):

api/v1alpha1/selfnoderemediationtemplate_webhook.go
```diff
@@ -0,0 +1,118 @@
...
+		if version == nil {
+			err = fmt.Errorf("k8s server version is nil")
+		}
+		snrtWebookLog.Error(err, "couldn't get retrieve k8s server version")
+		return err
+	} else {
+		isOutOfServiceTaintSupported = strings.Compare(fmt.Sprintf("%s.%s", version.Major, version.Minor), "1.26") >= 0
+
+		var majorVer, minorVer int64
+		if majorVer, err = strconv.ParseInt(version.Major, 10, 8); err != nil {
```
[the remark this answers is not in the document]
mshitrit: Yep Atoi is nicer, I guess I was caught up on `ParseInt` because I saw it's used in semver :slightly_smiling_face:

slintes (approved):

PR #107, opened 2023-04-04
setting minimum requeue time in case update failed

mshitrit:
As a follow up on [this](https://github.com/medik8s/self-node-remediation/pull/102#discussion_r1156211127) comment, setting minimum requeue time in case update failed instead of keeping the original requeue time.

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -157,8 +157,12 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer func() {
 		if updateErr := r.updateSnrStatus(ctx, snr); updateErr != nil {
 			if apiErrors.IsConflict(updateErr) {
-				if returnErr == nil && !returnResult.Requeue {
-					returnResult = ctrl.Result{RequeueAfter: time.Second}
+				minRequeue := time.Second
+				if returnResult.Requeue && minRequeue > returnResult.RequeueAfter {
+					minRequeue = returnResult.RequeueAfter
+				}
+				if returnErr == nil {
+					returnResult.RequeueAfter = minRequeue
 				}
 			} else {
 				returnErr = utilerrors.NewAggregate([]error{updateErr, returnErr})
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -158,7 +158,8 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		if updateErr := r.updateSnrStatus(ctx, snr); updateErr != nil {
 			if apiErrors.IsConflict(updateErr) {
 				minRequeue := time.Second
-				if returnResult.Requeue && minRequeue > returnResult.RequeueAfter {
+				//if requeue is already set choose the lowest one
+				if returnResult.RequeueAfter > 0 && minRequeue > returnResult.RequeueAfter {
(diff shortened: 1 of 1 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -157,8 +157,12 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer func() {
 		if updateErr := r.updateSnrStatus(ctx, snr); updateErr != nil {
 			if apiErrors.IsConflict(updateErr) {
-				if returnErr == nil && !returnResult.Requeue {
-					returnResult = ctrl.Result{RequeueAfter: time.Second}
+				minRequeue := time.Second
+				if returnResult.Requeue && minRequeue > returnResult.RequeueAfter {
```
clobrano, line 161: I wonder if the condition on `returnResult.Requeue` is necessary. Could it just be

```
if minRequeue > returnResult.RequeueAfter {
    minRequeue = returnResult.RequeueAfter
}
```

?

mshitrit: Default value for `RequeueAfter` is 0 when it isn't set  :man_shrugging:

slintes: I don't understand the logic implemented here tbh.

- The idea was to always requeueu quickly for conflicts, not?
- The `if` statement here does not make sense,  because `Requeue` can be false while `RequeuAfter` is set (will still trigger a requeue)
- The check for a `RequeueAfter` < 1s makes this too verbose for no value IMHO. Even when some of the calculations we do result in < 1s, IMHO 1s is always quick enough. Even more knowing that this is running potentially on many many nodes... let's not stress the API server too much :) 

IMHO this should be enough:
```
if updateErr := r.updateSnrStatus(ctx, snr); updateErr != nil {
    if apiErrors.IsConflict(updateErr) {			
        returnResult.RequeueAfter = 1 * time.Second
    else {
        returnErr = utilerrors.NewAggregate([]error{updateErr, returnErr}) // handles nil errors nicely
    }
}
```

slintes (changes requested):

controllers/selfnoderemediation_controller.go
mshitrit: >The idea was to always requeueu quickly for conflicts, not?

Yes, but in case the requeque of the original result is lower than that of the conflict I'd prefer not to override it.

>The if statement here does not make sense, because Requeue can be false while RequeuAfter is set (will still trigger a requeue)

That's interesting,  initially reading the comment on `RequeueAfter` I've assumed this not the case (i.e Requeue is set to true), after your comment and on a second read I'm not so sure.

I think that in order to stay on the safe side I'll replace `Requeue` with `RequeueAfter > 0` that should assure that the lowest requeue time will be applied.

controllers/selfnoderemediation_controller.go
```diff
@@ -157,8 +157,13 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	defer func() {
 		if updateErr := r.updateSnrStatus(ctx, snr); updateErr != nil {
 			if apiErrors.IsConflict(updateErr) {
-				if returnErr == nil && !returnResult.Requeue {
-					returnResult = ctrl.Result{RequeueAfter: time.Second}
+				minRequeue := time.Second
+				//if requeue is already set choose the lowest one
```
slintes, line 161: I still think this isn't needed, but I won't block on it...

/lgtm

PR #108, opened 2023-04-04
Bug fixes ECOPROJECT-1232  & ECOPROJECT-1233

mshitrit:
This PR
- Adds validation to remediation strategies on SNR (previously only on template) - in order to fix [ECOPROJECT-1232](https://issues.redhat.com//browse/ECOPROJECT-1232)
- Check Major and Minor version (previously Major version was checked instead of minor)  - in order to fix [ECOPROJECT-1233](https://issues.redhat.com//browse/ECOPROJECT-1233)

### DIFF
--- a/api/v1alpha1/selfnoderemediation_webhook.go
+++ b/api/v1alpha1/selfnoderemediation_webhook.go
@@ -28,11 +28,6 @@ var (
 )
 
 func (r *SelfNodeRemediation) SetupWebhookWithManager(mgr ctrl.Manager) error {
-
-	if err := initOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
-		return err
-	}
-
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
 		Complete()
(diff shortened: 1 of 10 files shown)


clobrano:
/lgtm

/hold

api/v1alpha1/selfnoderemediation_webhook.go
```diff
@@ -0,0 +1,62 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+var (
+	webhookRemediationLog = logf.Log.WithName("selfnoderemediation-resource")
+)
+
+func (r *SelfNodeRemediation) SetupWebhookWithManager(mgr ctrl.Manager) error {
+
+	if err := initOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
```
slintes, line 32: This call is duplicate now which at least isn't nice (and just relying here that it's called at the other webhook also isn't great) . Would it make sense to move this to main and pass the flag as arg? Or into its own file here and  just call init from main? 🤔

mshitrit: Indeed, good catch.
I didn't want to create a new file in an already crowded package, and I was worried about cycle dependency setting flags so I ended up putting it in `taints` file.

slintes: >  I ended up putting it in taints file

👍🏼 

/lgtm

PR #115, opened 2023-04-24
Support Succeeded Condition

mshitrit:
In order to fix ticket [ECOPROJECT-1246](https://issues.redhat.com//browse/ECOPROJECT-1246) ,and in context of NHC PR: https://github.com/medik8s/node-healthcheck-operator/pull/208

This PR add the support to Succeeded Condition in SNR

### DIFF
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -23,10 +23,10 @@ import (
 const (
 	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
 	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
-	// ProcessingTypeCondition is the condition type used to signal NHC the remediation status
-	ProcessingTypeCondition = "Processing"
-	// SucceededTypeCondition is the condition type used to signal NHC whether the remediation was successful or not
-	SucceededTypeCondition = "Succeeded"
+	// ProcessingConditionType is the condition type used to signal NHC the remediation status
+	ProcessingConditionType = "Processing"
+	// SucceededConditionType is the condition type used to signal NHC whether the remediation was successful or not
+	SucceededConditionType = "Succeeded"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -23,8 +23,10 @@ import (
 const (
 	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
 	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
-	// SnrConditionProcessing is the condition type used to signal NHC the remediation status
-	SnrConditionProcessing = "Processing"
+	// ProcessingTypeCondition is the condition type used to signal NHC the remediation status
+	ProcessingTypeCondition = "Processing"
+	// SucceededTypeCondition is the condition type used to signal NHC whether the remediation was successful or not
+	SucceededTypeCondition = "Succeeded"
 )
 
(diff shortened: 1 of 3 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -212,27 +212,40 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		succeededConditionStatus = metav1.ConditionFalse
 	default:
-		err := fmt.Errorf("unkown processingChangeReason:%s", reason)
+		err := fmt.Errorf("unkown processingChangeReason:%s", processingTypeReason)
 		r.Log.Error(err, "couldn't update snr processing condition")
 		return err
 	}
 
-	if meta.IsStatusConditionPresentAndEqual(snr.Status.Conditions, v1alpha1.SnrConditionProcessing, conditionStatus) {
+	if meta.IsStatusConditionPresentAndEqual(snr.Status.Conditions, v1alpha1.ProcessingTypeCondition, processingConditionStatus) {
```
slintes, line 233: what about the Succeeded condition?

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -23,8 +23,10 @@ import (
 const (
 	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
 	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
-	// SnrConditionProcessing is the condition type used to signal NHC the remediation status
-	SnrConditionProcessing = "Processing"
+	// ProcessingTypeCondition is the condition type used to signal NHC the remediation status
+	ProcessingTypeCondition = "Processing"
```
slintes, line 27: somehow this reads a bit strange... would either ProcessingConditionType or ConditionTypeProcessing be a better fit?

controllers/selfnoderemediation_controller.go
```diff
@@ -212,27 +212,40 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	return result, r.updateSnrStatusLastError(snr, err)
 }
 
-func (r *SelfNodeRemediationReconciler) updateSnrProcessingCondition(reason processingChangeReason, snr *v1alpha1.SelfNodeRemediation) error {
-	var conditionStatus metav1.ConditionStatus
-	switch reason {
+func (r *SelfNodeRemediationReconciler) updateSnrTypeConditions(processingTypeReason processingChangeReason, snr *v1alpha1.SelfNodeRemediation) error {
```
slintes, line 215: this also reads strange.... maybe just updateConditions()? (IMHO its clear that it is about SNR CRs, and conditions have a type)

slintes (changes requested):
two nits about naming, but also one issue

controllers/selfnoderemediation_controller.go
mshitrit: Maybe I'm missing something but I did think about it. 
Originally the idea was to skip update if Processing condition was not change, logically IIUC there is no such use case where Success condition would change and Processing condition will not so current condition still good enough.

slintes: remediationFinished and remediationTerminatedByNHC hace the same Processing but different Succeeeded status?

mshitrit: They do, but this is not my point.
The point that I'm arguing is that Succedded condition will change only when Processing status also changes.

slintes: >  Succeeded will change only when Processing also changes

when we first have remediationFinished and then remediationTerminatedByNHC or the other way around, Processing will stay at False, but Succeeded changes. Even when we know that this transition is unlikely or even impossible to happen today, that might change. Or we can get new reasons, resulting in other status combinations. IMHO this is is a maintenance burden and/or a potential future bug.

mshitrit: IMO both are fine, on one hand I don't see an issue with the current code but on the other hand the code change is very simple and risk free so I'll be happy to oblige :smile:

slintes (approved):

mshitrit:
Already passed

PR #118, opened 2023-05-04
Add "Community Edition" suffix to community build

mshitrit:


Makefile
```diff
@@ -247,11 +247,16 @@ bundle: manifests operator-sdk kustomize ## Generate bundle manifests and metada
...
-bundle-k8s: bundle ## Generate bundle manifests and metadata customized to k8s community release, then validate generated files.
+.PHONY: bundle-community-k8s
+bundle-community-k8s: bundle-community ## Generate bundle manifests and metadata customized to k8s community release, then validate generated files.
 	# Note that k8s 1.25+ needs PSA label
 	sed -r -i "s|by default\.|by default.\n    Note that prior to installing SNR on a Kubernetes 1.25+ cluster, a user must manually set a [privileged PSA label](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/) on SNR's namespace. It gives SNR's agents permissions to reboot the node (in case it needs to be remediated).|;" ./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
 
+.PHONY: bundle-community
+bundle-community: bundle
+	#Add Community Edition suffix to operator name
+	sed -r -i "s|displayName: Self Node Remediation Operator|displayName: Self Node Remediation Operator - Community Edition|;" ./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
```
clobrano, line 258: nit: If, for any reason, this command is run twice, it will add " - Comminity Edition" twice.
You might avoid it with something like

```
	sed -r -i "s|displayName: Machine Deletion Remediation operator.*|displayName: Machine Deletion Remediation Operator - Community Edition|;" ${CSV}
```

Or just ignore it, since I don't expect this to be called twice by CI :)

clobrano: /lgtm
/hold
for the reason above, but feel free to cancel it

mshitrit: Even if it runs twice it'll not happen (I've checked) but from a different reason (you are correct about how sed command will work).
The reason we'll still have only one `Comminity Edition` is because `bundle` is a prerequisite and will run before and reset the csv.

In any case I agree that the suggested change is better practice :+1:

clobrano: ah, good point :smile: 
/lgtm

clobrano:
actually giving
/lgtm

PR #120, opened 2023-05-09
Update makefile and tools versions

mshitrit:
- adding sort-import
- upgrading tool versions  
Most of the noise is due to import sorting, and automatic manifests update

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -402,3 +402,6 @@ test-imports: sort-imports ## Check for sorted imports
 .PHONY: fix-imports
 fix-imports: sort-imports ## Sort imports
 	$(SORT_IMPORTS) . -w
+
+.PHONY: full-gen
+full-gen:  generate manifests vendor tidy bundle fix-imports bundle-reset ## generates all automatically generated content
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ CONTROLLER_GEN_VERSION = v0.12.0
 # GO_VERSION refers to the version of Golang to be downloaded when running dockerized version
 GO_VERSION = 1.20
 
-# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.23
+# ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary, align with k8s.io/client-go in go.mod
+ENVTEST_K8S_VERSION = 1.25
 
 # versions at https://github.com/operator-framework/operator-sdk/releases
 OPERATOR_SDK_VERSION = v1.28.1
--- a/Makefile
+++ b/Makefile
@@ -2,24 +2,25 @@
 SHELL := /bin/bash
 
 # versions at  https://github.com/kubernetes-sigs/controller-tools/releases
-CONTROLLER_GEN_VERSION = v0.10.0
+CONTROLLER_GEN_VERSION = v0.12.0
 
 # GO_VERSION refers to the version of Golang to be downloaded when running dockerized version
-GO_VERSION = 1.19
+GO_VERSION = 1.20
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
 
 # versions at https://github.com/operator-framework/operator-sdk/releases
-OPERATOR_SDK_VERSION = v1.25.3
+OPERATOR_SDK_VERSION = v1.28.1
 
 # versions at https://github.com/operator-framework/operator-registry/releases
 OPM_VERSION = v1.26.2
 
 # versions at https://github.com/kubernetes-sigs/kustomize/releases
-KUSTOMIZE_VERSION = v4.5.7
+KUSTOMIZE_VERSION = v5.0.2
 
-SORT_IMPORTS_VERSION = v0.1.0
+# versions at https://github.com/slintes/sort-imports/tags
+SORT_IMPORTS_VERSION = v0.2.1
 
 OPERATOR_NAME ?= self-node-remediation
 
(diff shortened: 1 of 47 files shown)


Makefile
```diff
@@ -173,7 +175,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 .PHONY: docker-build
-docker-build: check ## Build docker image with the manager.
+docker-build: test
```
razo7, line 178: Why do you want to remove the `check` target?

Makefile
```diff
@@ -19,6 +19,8 @@ OPM_VERSION = v1.26.2
 # versions at https://github.com/kubernetes-sigs/kustomize/releases
 KUSTOMIZE_VERSION = v4.5.7
 
+SORT_IMPORTS_VERSION = v0.1.0
```
razo7, line 22: There is a newer version. [v0.2.1](https://github.com/slintes/sort-imports/tags).

Makefile
```diff
@@ -2,22 +2,22 @@
...
 # GO_VERSION refers to the version of Golang to be downloaded when running dockerized version
-GO_VERSION = 1.19
+GO_VERSION = 1.20
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
 
 # versions at https://github.com/operator-framework/operator-sdk/releases
-OPERATOR_SDK_VERSION = v1.25.3
+OPERATOR_SDK_VERSION = v1.28.1
```
razo7, line 14: Can we update the scorecard tests images in https://github.com/medik8s/self-node-remediation/tree/main/config/scorecard/patches as part of updating the OPERATOR_SDK_VERSION? See old PR for reference #54

Makefile
mshitrit: Locally it's simpler to build without `test`, but this is a valid point so I've changed it so CI build will still use `check` while adding another option for a build with `test`

Makefile
```diff
@@ -355,8 +359,10 @@ check: protoc ## Dockerized version of make test
 	$(DOCKER_GO) "make test"
 
 .PHONY: container-build
-container-build: ## Build containers
```
razo7, line 356: NIT: It might be nice to keep this target for building the containers locally.

Makefile
```diff
@@ -355,8 +359,10 @@ check: protoc ## Dockerized version of make test
...
-	make docker-build bundle bundle-build
+container-build:docker-build bundle bundle-build ## Build containers
+
+.PHONY: container-build-check
+container-build-check:docker-build-check bundle bundle-build ## Build containers
```
razo7, line 366: Two small NITs:
1. I don't think we need to call for `bundle` twice here. First on this line, and then again in [`bundle-build`](https://github.com/medik8s/self-node-remediation/blob/main/Makefile#L276-L278)
2. A space is missing right after the target name :)

Makefile
razo7: WFM

Makefile
```diff
@@ -368,10 +377,28 @@ vendor: ## Runs go mod vendor
...
+sort-imports: ## Download sort-imports locally if necessary.
+	$(call go-install-tool,$(SORT_IMPORTS),github.com/slintes/sort-imports@$(SORT_IMPORTS_VERSION))
+
+.PHONY: test-imports
+test-imports: sort-imports ## Check for sorted imports
+	$(SORT_IMPORTS) .
+
+.PHONY: fix-imports
+fix-imports: sort-imports ## Sort imports
+	$(SORT_IMPORTS) . -w
```
razo7, line 404: 1. `make fix-imports` doesn't work with `. -w`, since it was changed in v0.2.1- ["requires flags to be before project dir now"](https://github.com/slintes/sort-imports/releases/tag/v0.2.1).
2. v0.2.1 does add few more sorting changes

razo7 (changes requested):
Left some NITs and we should add the _sort-imports_ v0.2.1 changes

Makefile
```diff
@@ -2,22 +2,25 @@
...
 # versions at  https://github.com/kubernetes-sigs/controller-tools/releases
-CONTROLLER_GEN_VERSION = v0.10.0
+CONTROLLER_GEN_VERSION = v0.12.0
 
 # GO_VERSION refers to the version of Golang to be downloaded when running dockerized version
-GO_VERSION = 1.19
+GO_VERSION = 1.20
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
```
razo7, line 11: Bump to 1.25 as we already use K8s v1.25 in [go.mod](https://github.com/medik8s/self-node-remediation/blob/main/go.mod#L18-L21)

Makefile
```diff
@@ -2,22 +2,25 @@
...
 
 # versions at https://github.com/operator-framework/operator-sdk/releases
-OPERATOR_SDK_VERSION = v1.25.3
+OPERATOR_SDK_VERSION = v1.28.1
 
 # versions at https://github.com/operator-framework/operator-registry/releases
 OPM_VERSION = v1.26.2
```
razo7, line 17: Newer OPM version - [v1.26.5](https://github.com/operator-framework/operator-registry/releases/tag/v1.26.5)

Makefile
mshitrit: I did keep it, git view might be a bit confusing since it's showing it was removed (here in red) and later on added (in green)

Makefile
mshitrit: Hmm I don't quite follow.
As far as I see `make fix-imports` is executed ok and does what I expect it to do.
What is the change you suggest ?

mshitrit: My mistake, was running with cached 0.1.0 version, fixed

PR #122, opened 2023-05-21
Add .Phony to Makefile

razo7:
Similar to [NMO's PR](https://github.com/medik8s/node-maintenance-operator/pull/12), and [NHC's PR](https://github.com/medik8s/node-healthcheck-operator/pull/210).

Add [.PHONY](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) to each target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,24 +127,30 @@ all: build
 # More info on the awk command:
 # http://linuxcommand.org/lc3_adv_awk.php
 
+.PHONY: help
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 ##@ Development
 
+.PHONY: manifests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -260,13 +276,12 @@ bundle-community-k8s: bundle-community ## Generate bundle manifests and metadata
...
 .PHONY: bundle-community
-bundle-community: bundle
-	#Add Community Edition suffix to operator name
+bundle-community: bundle ##Add Community Edition suffix to operator name
 	sed -r -i "s|displayName: Self Node Remediation Operator.*|displayName: Self Node Remediation Operator - Community Edition|;" ./bundle/manifests/$(OPERATOR_NAME).clusterserviceversion.yaml
 
 .PHONY: bundle-update
-bundle-update:
-    # update container image in the metadata
+bundle-update: ## Update containerImage, createdAt, skipRange, and icon fields in the bundle's CSV, then validate the bundle directory
```
mshitrit, line 283: I think the comment is wrong, IIUC it only updates  containerImage & createdAt,

mshitrit:
Just one NIT feel free to fix/unhold
/lgtm
/hold

Makefile
razo7: In [lines](https://github.com/medik8s/self-node-remediation/pull/122/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R288-R291) it does update  skipRange, and icon fields.

PR #117, opened 2023-05-01
Automate GitHub Release Draft

razo7:
Similar to the rest of Medik8s operators we add an automation for the creation of Github Release upon Git tag creation.

[ECOPROJECT-1128](https://issues.redhat.com//browse/ECOPROJECT-1128)

1. Add Github Action step for building the draft.
2. Use container-build-community target for CI so _quay.io/medik8s_ will store SNR's community version for K8s.

PR #123, opened 2023-05-24
Use Ubuntu 22 as Hosted Runner in Github CI

razo7:
Follow up PR to #117 which will overcome the problem with **CGO_ENABLED=0** in controller-gen binary, when we update [hosted runner Ubuntu image for github-actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners) - `ubuntu-20.04` -> `ubuntu-22.04`.

### DIFF
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   push_to_registry:
     name: Build and push images to quay.io/medik8s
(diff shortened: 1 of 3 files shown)


slintes:
> Nice! Your idea worked @slintes 👏🏻

perfect, thx for trying it 👍🏼 

/lgtm

PR #125, opened 2023-06-11
Handle plus sign in minor version

mshitrit:
Bug fix for [ECOPROJECT-1325](https://issues.redhat.com//browse/ECOPROJECT-1325)

### DIFF
--- a/main.go
+++ b/main.go
@@ -135,7 +135,6 @@ func initSelfNodeRemediationManager(mgr manager.Manager) {
 	setupLog.Info("Starting as a manager that installs the daemonset")
 
 	if err := utils.InitOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
-		utils.IsOutOfServiceTaintSupported = false
 		setupLog.Error(err, "unable to verify out of service taint support. out of service taint isn't supported")
 	}
 
--- a/main.go
+++ b/main.go
@@ -135,8 +135,8 @@ func initSelfNodeRemediationManager(mgr manager.Manager) {
 	setupLog.Info("Starting as a manager that installs the daemonset")
 
 	if err := utils.InitOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
-		setupLog.Error(err, "unable to verify out of service taint support")
-		os.Exit(1)
+		utils.IsOutOfServiceTaintSupported = false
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -135,8 +135,8 @@ func initSelfNodeRemediationManager(mgr manager.Manager) {
 	setupLog.Info("Starting as a manager that installs the daemonset")
 
 	if err := utils.InitOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
-		setupLog.Error(err, "unable to verify out of service taint support")
-		os.Exit(1)
+		utils.IsOutOfServiceTaintSupported = false
```
slintes, line 138: This is redundant, it's the default value. (and even if it wouldn't be the default value, setting that global var, which isn't nice on its own already, here and not in its own package, wouldn't be nice as well IMHO)

pkg/utils/taints.go
```diff
@@ -51,21 +52,21 @@ func InitOutOfServiceTaintSupportedFlag(config *rest.Config) error {
...
+		utilTaintsLog.Error(err, "couldn't retrieve k8s server version")
 		return err
 	} else {
 		var majorVer, minorVer int
 		if majorVer, err = strconv.Atoi(version.Major); err != nil {
 			utilTaintsLog.Error(err, "couldn't parse k8s major version", "major version", version.Major)
 			return err
 		}
-		if minorVer, err = strconv.Atoi(version.Minor); err != nil {
+		if minorVer, err = strconv.Atoi(strings.ReplaceAll(version.Minor, "+", "")); err != nil {
```
slintes, line 69: what about putting this else block into it's own func with a unit test?

slintes (changes requested):

pkg/utils/taints.go
mshitrit: Good idea.

main.go
mshitrit: True, I think it's a bit more readable this way (i.e an error occurred so we are explicitly disabling this feature) but it's definitely not something I'm feeling strongly about, so I don't mind removing it.

clobrano:
/lgtm 
Giving others a chance to review as wel, feel free to unhold
/hold

slintes (approved):

PR #128, opened 2023-06-21
Removed '+patchMergeKey' and '+patchStrategy' markers.

epenchev:
Removed '+patchMergeKey' and '+patchStrategy' markers.
These are for guiding strategic merge, which is not supported on CRDs.
According to https://github.com/kubernetes-sigs/gateway-api/issues/342 the markers +patchMergeKey' and '+patchStrategy' don't work on CRDs.
I've tested this by running `make generate` and `make manifests` and the resulting yaml CRD didn't change.
The desired effect in config/crd/bases/self-node-remediation.medik8s.io_selfnoderemediations.yaml
```
type: array
                x-kubernetes-list-map-keys:
                - type
                x-kubernetes-list-type: map
```

is from '+listType' and '+listMapKey' as already applied.


slintes:
/approve

@mshitrit fyi

PR #130, opened 2023-06-27
Support custom toleration on ds

mshitrit:
This PR meant to solve ECOPROJECT-1235, it would allow the user to use SNR configuration in order to create customized tolerations on SNR's daemonset.

### DIFF
--- a/api/v1alpha1/selfnoderemediationconfig_webhook.go
+++ b/api/v1alpha1/selfnoderemediationconfig_webhook.go
@@ -161,9 +161,6 @@ func (f *field) validate() error {
 
 func (r *SelfNodeRemediationConfig) validateCustomTolerations() error {
 	customTolerations := r.Spec.CustomDsTolerations
-	if len(customTolerations) == 0 {
-		return nil
-	}
 	for _, toleration := range customTolerations {
 		if err := validateToleration(toleration); err != nil {
 			return err
--- a/api/v1alpha1/selfnoderemediationconfig_webhook.go
+++ b/api/v1alpha1/selfnoderemediationconfig_webhook.go
@@ -172,8 +172,14 @@ func (r *SelfNodeRemediationConfig) validateCustomTolerations() error {
 func validateToleration(toleration v1.Toleration) error {
 	if len(toleration.Operator) > 0 {
 		switch toleration.Operator {
-		case v1.TolerationOpEqual, v1.TolerationOpExists:
-			//Valid nothing to do
+		case v1.TolerationOpEqual:
+		//Valid nothing to do
+		case v1.TolerationOpExists:
+			if len(toleration.Value) != 0 {
+				err := fmt.Errorf("invalid value for tolerarion, value must be empty for Operator value is Exists")
+				selfNodeRemediationConfigLog.Error(err, "invalid value for tolerarion, value must be empty for Operator value is Exists")
+				return err
+			}
 		default:
 			err := fmt.Errorf("invalid operator for tolerarion: %s", toleration.Operator)
 			selfNodeRemediationConfigLog.Error(err, "invalid operator for tolerarion", "valid values", []v1.TolerationOperator{v1.TolerationOpEqual, v1.TolerationOpExists}, "received value", toleration.Operator)
--- a/api/v1alpha1/selfnoderemediationconfig_webhook.go
+++ b/api/v1alpha1/selfnoderemediationconfig_webhook.go
@@ -176,13 +176,13 @@ func validateToleration(toleration v1.Toleration) error {
 		//Valid nothing to do
 		case v1.TolerationOpExists:
 			if len(toleration.Value) != 0 {
-				err := fmt.Errorf("invalid value for tolerarion, value must be empty for Operator value is Exists")
-				selfNodeRemediationConfigLog.Error(err, "invalid value for tolerarion, value must be empty for Operator value is Exists")
+				err := fmt.Errorf("invalid value for toleration, value must be empty for Operator value is Exists")
+				selfNodeRemediationConfigLog.Error(err, "invalid value for toleration, value must be empty for Operator value is Exists")
 				return err
 			}
 		default:
-			err := fmt.Errorf("invalid operator for tolerarion: %s", toleration.Operator)
-			selfNodeRemediationConfigLog.Error(err, "invalid operator for tolerarion", "valid values", []v1.TolerationOperator{v1.TolerationOpEqual, v1.TolerationOpExists}, "received value", toleration.Operator)
+			err := fmt.Errorf("invalid operator for toleration: %s", toleration.Operator)
+			selfNodeRemediationConfigLog.Error(err, "invalid operator for toleration", "valid values", []v1.TolerationOperator{v1.TolerationOpEqual, v1.TolerationOpExists}, "received value", toleration.Operator)
 			return err
 		}
 	}
@@ -192,8 +192,8 @@ func validateToleration(toleration v1.Toleration) error {
 		case v1.TaintEffectNoSchedule, v1.TaintEffectPreferNoSchedule, v1.TaintEffectNoExecute:
 			//Valid nothing to do
 		default:
-			err := fmt.Errorf("invalid taint effect for tolerarion: %s", toleration.Effect)
-			selfNodeRemediationConfigLog.Error(err, "invalid taint effect for tolerarion", "valid values", []v1.TaintEffect{v1.TaintEffectNoSchedule, v1.TaintEffectPreferNoSchedule, v1.TaintEffectNoExecute}, "received value", toleration.Effect)
+			err := fmt.Errorf("invalid taint effect for toleration: %s", toleration.Effect)
+			selfNodeRemediationConfigLog.Error(err, "invalid taint effect for toleration", "valid values", []v1.TaintEffect{v1.TaintEffectNoSchedule, v1.TaintEffectPreferNoSchedule, v1.TaintEffectNoExecute}, "received value", toleration.Effect)
 			return err
 		}
 	}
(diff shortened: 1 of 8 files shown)


controllers/selfnoderemediationconfig_controller.go
```diff
@@ -190,3 +198,54 @@ func (r *SelfNodeRemediationConfigReconciler) syncCerts(cr *selfnoderemediationv
 	}
 	return nil
 }
+
+func (r *SelfNodeRemediationConfigReconciler) updateDsTolerations(objs []*unstructured.Unstructured, tolerations []corev1.Toleration) error {
+	r.Log.Info("Updating DS tolerations")
+	if len(tolerations) == 0 {
```
razo7, line 204: Can tolerations be nil since this field is [`omitempty`](https://github.com/medik8s/self-node-remediation/pull/130/files#diff-792340dfdee839c7939f417270ed1aa40b9f59b0e2d757b5d5417f7a59fbeb9fR118)? If so , then we should check it first.

mshitrit: this check covers nil as well.
i.e:
```
len(nil) is 0
```

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -190,3 +198,54 @@ func (r *SelfNodeRemediationConfigReconciler) syncCerts(cr *selfnoderemediationv
...
+		return nil
+	}
+	for _, toleration := range tolerations {
+		for _, ds := range objs {
+			if err := r.updateTolerationOnDs(ds, toleration); err != nil {
```
slintes, line 209: - why iterating the tolerations, and not adding all at once?
- just to protect our future selfs from surprises in case we add manifests to `/install`: we know that there is only 1 object today, the daemonset. What about ensuring that it is really 1 object only, and process it directly without iteration? And fail if there are more manifests, which means this code needs an update?

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -190,3 +198,54 @@ func (r *SelfNodeRemediationConfigReconciler) syncCerts(cr *selfnoderemediationv
...
+
+	updatedTolerations, ok := tolerations.([]interface{})
+	if !ok {
+		err := fmt.Errorf(fmt.Sprintf("failed to convert tolerations expected of type %T but got %T", []interface{}{}, tolerations))
+		r.Log.Error(err, err.Error())
+		return err
+	}
+	r.Log.Info("updateTolerationOnDs original tolerations", "tolerations", updatedTolerations)
+
+	configTolerations := map[string]interface{}{
```
slintes, line 232: - `configTolerations` isn't a very descriptive name, is it? 🤔 
- this conversion can be done easier: `newToleration, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&toleration)`

slintes (changes requested):

controllers/selfnoderemediationconfig_controller.go
mshitrit: >why iterating the tolerations, and not adding all at once?

IIUC each toleration needs to be converted from `v1.Toleration` to `map[string]interface{}`

>just to protect our future selfs from surprises in case we add manifests to /install: we know that there is only 1 object today, the daemonset. What about ensuring that it is really 1 object only, and process it directly without iteration? And fail if there are more manifests, which means this code needs an update?

Good idea :+1:

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -134,6 +136,12 @@ func (r *SelfNodeRemediationConfigReconciler) syncConfigDaemonSet(ctx context.Co
 		logger.Error(err, "Fail to render config daemon manifests")
 		return err
 	}
+
+	if err := r.updateDsTolerations(objs, snrConfig.Spec.CustomDsTolerations); err != nil {
+		logger.Error(err, "Fail update daemonset tolerations")
+		return err
```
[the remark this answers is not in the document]
mshitrit: I agree, webhook is probably the best solution here

controllers/selfnoderemediationconfig_controller.go
slintes: > is there a simple way to convert the whole slice ?

probably not, but you can have that loop where you do that conversion IMHO

mshitrit: If it's not something you feel strongly about I think I prefer to keep as is.
Since we need to iterate over the tolerations in any case I do prefer the pattern of having one method that handles the whole slice which calls a method that handle a single element (toleration).
IMO it's improves readability and is more modular.

controllers/selfnoderemediationconfig_controller.go
mshitrit: Not sure about `can't` but it definitely `shouldn't` be empty

controllers/selfnoderemediationconfig_controller.go
slintes: It doesn't make a difference on performance in this case, but what kind of pattern is it where you do more work than needed inside the loop? (processing the whole DS object vs just converting the tolerations to unstructured) 🤔 

> IMO it's improves readability

let's not start arguing about this 😎 

>  and is more modular.

if you want to have this more modular, put the conversion of the toleration slice into its own function?

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -95,15 +96,19 @@ var _ webhook.Validator = &SelfNodeRemediationConfig{}
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *SelfNodeRemediationConfig) ValidateCreate() error {
 	selfNodeRemediationConfigLog.Info("validate create", "name", r.Name)
-
-	return r.validateTimes()
+	if err := r.validateTimes(); err != nil {
```
slintes, line 99: Better run all validations ad aggregate errors, so users know about all problems in one run.
See e.g. https://github.com/medik8s/node-healthcheck-operator/blob/main/api/v1alpha1/nodehealthcheck_webhook.go#L116

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -149,3 +154,41 @@ func (f *field) validate() error {
 
 	return nil
 }
+
+func (r *SelfNodeRemediationConfig) validateCustomTolerations() error {
+	customTolerations := r.Spec.CustomDsTolerations
+	if len(customTolerations) == 0 {
```
slintes, line 160: redundant, the range assignment in the loop can deal with nil and empty slices

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -149,3 +154,41 @@ func (f *field) validate() error {
...
+		}
+	}
+	return nil
+}
+
+func validateToleration(toleration v1.Toleration) error {
+	if len(toleration.Operator) > 0 {
+		switch toleration.Operator {
+		case v1.TolerationOpEqual, v1.TolerationOpExists:
+			//Valid nothing to do
```
slintes, line 175: please add: when operator is Exists, value must be empty

api/v1alpha1/selfnoderemediationconfig_webhook.go
```diff
@@ -149,3 +154,41 @@ func (f *field) validate() error {
...
+		default:
+			err := fmt.Errorf("invalid operator for tolerarion: %s", toleration.Operator)
```
slintes, line 177: typo: tolerarion (not here only)

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -134,6 +155,7 @@ func testValidCR(validationType string) {
 	snrc.Spec.PeerRequestTimeout = &metav1.Duration{Duration: 30 * time.Second}
 	snrc.Spec.ApiCheckInterval = &metav1.Duration{Duration: 10*time.Second + 500*time.Millisecond}
 	snrc.Spec.PeerUpdateInterval = &metav1.Duration{Duration: 10 * time.Second}
+	snrc.Spec.CustomDsTolerations = []v1.Toleration{{Key: "validValue", Effect: v1.TaintEffectNoExecute}, {}, {Operator: v1.TolerationOpEqual, TolerationSeconds: pointer.Int64(-5)}, {Value: "SomeValidValue"}}
```
slintes, line 158: this test should check if  the error contains all problems

slintes (changes requested):

controllers/selfnoderemediationconfig_controller.go
mshitrit: > but what kind of pattern is it where you do more work than needed inside the loop? 

Fair point

>if you want to have this more modular, put the conversion of the toleration slice into its own function?

Sure, that WFM :+1:

api/v1alpha1/selfnoderemediationconfig_webhook.go
mshitrit: cool, I didn't know range is valid over nil :+1:

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
mshitrit: Hmm maybe I'm missing something IIUC these are valid values, the test check that there are no errors

slintes: oh sorry, jumping too much around in the code. But maybe add it to testMultipleInvalidFields() above?

slintes:
/lgtm

left one comment about testing multiple validation errors, will leave it to you if you want to add that, feel free to unhold :) 
/hold

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
mshitrit: Sure, sorry I've initial missed your latest respons -, for some reason this thread didn't automatically update in main conversation tab (I saw it once I switched to files tab).

Thanks for mentioning it again with the lgtm comment :+1:

slintes (approved):

PR #131, opened 2023-07-02
use common repo functionality in SNR (Node Role labels)

mshitrit:


mshitrit:
Hmm IIUC the problem here is that on the one hand common (which is consumed by SNR in this PR) requires `k8s.io/apimachinery v0.26.3` instead of `k8s.io/apimachinery v0.25.0` which was previously used in SNR, and on the other hand SNR consumes `github.com/openshift/machine-api-operator` which in turn does not support the newer version of `k8s.io/apimachinery`.
AFAIS there is no newer version of `github.com/openshift/machine-api-operator` which supports `k8s.io/apimachinery v0.26.3`.

So IIUC in theory the following options are:

- raise an issue and wait for newer version of `github.com/openshift/machine-api-operator` 
- downgrade k8s.io/apimachinery  in common
- close this PR (i.e SNR will not consume common)
- remove usage of `github.com/openshift/machine-api-operator` from SNR

razo7:
/lgtm
Giving others a chance to review as wel, feel free to unhold
/hold

pkg/render/render.go
```diff
@@ -24,7 +22,6 @@ type Data struct {
 }
 
 type Config struct {
-	*mcfgv1.ControllerConfigSpec
```
clobrano, line 27: I don't get why this was removed

slintes (changes requested):
see comment inside

main.go
```diff
@@ -68,7 +66,6 @@ func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
 	utilruntime.Must(selfnoderemediationv1alpha1.AddToScheme(scheme))
-	utilruntime.Must(machinev1beta1.AddToScheme(scheme))
```
slintes, line 71: and here the comment 😉 
I'm pretty sure we need this, I expect SNR to fail without it when we get a CR created by MHC, where we need to get the Machine

pkg/render/render.go
mshitrit: That's a miss on my part, has to do with some experiments I've tried and forgot to revert.
good catch.

main.go
mshitrit: Good point,
Since we intend to have MHC functionality moved to NHC I wonder if we can also leap into the conclusion that that eventually SNR will work only  with NHC ? :thinking: 
If so we can put this PR "on ice" until then.

Or maybe there is a better option that I'm missing , WDYT ?

pkg/render/render.go
mshitrit: However looking at the change I'm not sure it's wrong in that context.
It eliminates the dependency in `machine-config-operator` and it looks like this config isn't used.

main.go
slintes: > until then

"then" won't be in the near future... why not just keeping this line here? 🤷🏼‍♂️

mshitrit: IIUC the meaning of keeping this line is keeping the dependency in `machine-api-operator` which brings us back to the [initial issue](https://github.com/medik8s/self-node-remediation/pull/131#issuecomment-1616603651).

slintes: yes, the machine API moved from `github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1` to `github.com/openshift/api/machine/v1beta1`, so we need to add it to our scheme from there. It's the same like in `selfnoderemediation_controller.go.`

slintes (approved):

PR #133, opened 2023-07-11
Match labels

mshitrit:
In order to solve [ECOPROJECT-1307](https://issues.redhat.com//browse/ECOPROJECT-1307) using and matching unique labels in DS


install/self-node-remediation-deamonset.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app: self-node-remediation-agent
```
slintes, line 11: I'd suggest to use recommended label names, e.g.:

- `app.kubernetes.io/name: self-node-remediation`
- `app.kubernetes.io/component: agent`

See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Maybe out of scope for this PR, but it might make sense to update all labels, see https://github.com/medik8s/node-healthcheck-operator/pull/146 as example.

config/manager/manager.yaml
```diff
@@ -11,16 +11,17 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: controller-manager
+    app.kubernetes.io/component: controller-manager
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app.kubernetes.io/component: controller-manager
```
slintes, line 18: heads up: updates to the label selector of running deployments are forbidden, a new deployment is needed in that case. Unfortunately OLM doesn't deal with this, in the past at least. When updating the operator, it tries the forbidden label selector update and fails, instead of replacing the deployment.
So either please test if this is still the case, or rename the deployment.

mshitrit: Thanks, I think I'll stick only to the DS labels in this PR

razo7: Similar issue as for FAR PR https://github.com/medik8s/fence-agents-remediation/pull/46

slintes: I'd suggest to test an update also for just the ds change, I wouldn't be that much surprised if we run into the same issue with our own code similar to OLM

razo7:
IIUC the suggested changes, then we should update the label in few more places:
SNR:
- https://github.com/medik8s/self-node-remediation/blob/main/pkg/utils/pods.go#L17
- https://github.com/medik8s/self-node-remediation/blob/main/controllers/selfnoderemediation_controller_test.go#L546
- https://github.com/medik8s/self-node-remediation/blob/main/controllers/selfnoderemediation_controller_test.go#L570
NHC
-  https://github.com/medik8s/node-healthcheck-operator/blob/main/.github/workflows/pre-submit.yaml#L173-L175

config/manager/manager.yaml
mshitrit: Indeed, same issue occured.
tested again successfully after [this](https://github.com/medik8s/self-node-remediation/pull/133/commits/be01bcb8b67a5aa2bcdad86eda39343f898b7702) commit (removing old ds)

slintes (commented):
Wouldn't it make sense to add this to the place where we already deal with the Daemonset? 🤔 

https://github.com/medik8s/self-node-remediation/blob/main/controllers/selfnoderemediationconfig_controller.go#L102

pkg/updatehelper/updatehelper.go
```diff
@@ -0,0 +1,56 @@
...
+		return errors.Wrap(err, "unable to get the deployment namespace")
+	}
+	ds := &appsv1.DaemonSet{}
+	key := types.NamespacedName{
+		Namespace: ns,
+		Name:      dsName,
+	}
+	if err := ui.client.Get(ctx, key, ds); err == nil {
+		if err = ui.client.Delete(ctx, ds); err != nil {
+			ui.log.Error(err, "snr update failed could not delete old damenoset")
```
clobrano, line 43: typo: damenoset

clobrano (changes requested):

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -243,3 +255,32 @@ func (r *SelfNodeRemediationConfigReconciler) convertTolerationsToUnstructed(tol
...
+
+func (r *SelfNodeRemediationConfigReconciler) removeOldDsOnUpdateOperator(ctx context.Context) error {
+	ds := &v1.DaemonSet{}
+	key := types.NamespacedName{
+		Namespace: r.Namespace,
+		Name:      dsName,
+	}
+
+	if err := r.Client.Get(ctx, key, ds); err == nil {
+		if lastChangedActualVal, _ := ds.Annotations[lastChangedAnnotationKey]; lastChangedActualVal == lastChangeAnnotationVal {
```
razo7, line 267: NIT: `lastChangedActualVal` and `lastChangeAnnotationVal` are a little bit similar. Can we use `Expected` rather than `Actual`?

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -38,6 +40,12 @@ import (
 	"github.com/medik8s/self-node-remediation/pkg/render"
 )
 
+const (
+	dsName                   = "self-node-remediation-ds"
+	lastChangedAnnotationKey = "dsLastChangeVersion"
+	lastChangeAnnotationVal  = "rhwa-23.3"
```
razo7, line 46: typo: `lastChange` -> `lastChanged`

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -103,6 +111,10 @@ func (r *SelfNodeRemediationConfigReconciler) syncConfigDaemonSet(ctx context.Co
 	logger := r.Log.WithName("syncConfigDaemonset")
 	logger.Info("Start to sync config daemonset")
 
+	if err := r.removeOldDsOnUpdateOperator(ctx); err != nil {
```
slintes, line 114: What about moving this to just before the `r.syncK8sResource` call where we loop the resources we are going to sync? There we have the new DaemonSet available, and can compare its annotation value with the deployed version, and delete the deployed ds when values don't match. Then you don't need the ds name var (just take it from the new ds), and also don't need the hardcoded value var to compare with.

controllers/selfnoderemediationconfig_controller.go
mshitrit: I'm not sure `Expected` fits this use case, since I don't expect old value to equal it, however I do agree that using `Actual` usually makes sense when coupled with `Expected` .
How about replacing `lastChangedActualVal` with `lastChangedFoundVal` ?

slintes (changes requested):

controllers/selfnoderemediationconfig_controller.go
mshitrit: why do you consider it a typo ?
Can't you say something like `what was the last change ? ` ?

controllers/selfnoderemediationconfig_controller.go
razo7: SGTM

controllers/selfnoderemediationconfig_controller.go
razo7: Yes, it was just that you have used `lastChanged` in few other places, so I though you prefer `lastChanged` instead of `lastChange` 🤷🏻‍♂️

slintes (commented):
I would reorder things in the `removeOldDsOnUpdateOperator()` a bit (error handling first, then the happy path), but I won't block on this. Btw, why `Operator` in the method name?

Can we have a unit test for this?

mshitrit:
> I would reorder things in the `removeOldDsOnUpdateOperator()` a bit (error handling first, then the happy path), 

Yep, I agree it would make more sense 

>Btw, why `Operator` in the method name?

IIUC this use case is relevant for when the operator is updated, just "update" seemed very generic to me and I wasn't sure the context will be clear enough
 
> Can we have a unit test for this?

Indeed, good point.

slintes:
> IIUC this use case is relevant for when the operator is updated, just "update" seemed very generic to me and I wasn't sure the context will be clear enough
>

Ah ok, makes sense. Maybe `removeOldDsOnOperatorUpdate()` is a bit less confusing? 🙂

slintes (approved):
nice test 👍🏼

PR #135, opened 2023-07-12
Explicitly set MinVersion of TLS

clobrano:
Currently, the default MinVersion value for TLS configuration is used,
which is TLS1.0 and considered insecure.

Explicitly set the MinVersion to a secure version of TLS.

closes: https://issues.redhat.com/browse/ECOPROJECT-1419

Signed-off-by: Carlo Lobrano <c.lobrano@gmail.com>

### DIFF
--- a/pkg/controlplane/manager.go
+++ b/pkg/controlplane/manager.go
@@ -16,6 +16,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/medik8s/self-node-remediation/pkg/certificates"
 	"github.com/medik8s/self-node-remediation/pkg/peers"
 	"github.com/medik8s/self-node-remediation/pkg/utils"
 )
@@ -150,7 +151,10 @@ func (manager *Manager) isEndpointAccessible() bool {
 func (manager *Manager) isKubeletServiceRunning() bool {
 	url := fmt.Sprintf("https://%s:%s/pods", manager.nodeName, kubeletPort)
 	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+			MinVersion:         certificates.TLSMinVersion,
+		},
 	}
 	httpClient := &http.Client{Transport: tr}
 
--- a/pkg/certificates/credentials.go
+++ b/pkg/certificates/credentials.go
@@ -8,6 +8,8 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
+const TLSMinVersion = tls.VersionTLS13
+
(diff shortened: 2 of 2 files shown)


clobrano:
Testing with my cluster I am seeing some problems
/hold

clobrano:
but not with `ci/prow/4.13-openshift-e2e` :thinking:

clobrano:
test passed also in my local cluster /unhold

mshitrit:
I might be misunderstanding but I think it might be relevant here (see below - credentials.go L18-22) as well
```
return credentials.NewTLS(&tls.Config{
		Certificates: []tls.Certificate{*keyPair},
		ClientAuth:   tls.RequireAndVerifyClientCert,
		ClientCAs:    pool,
	}), nil
```

Also since it's being used in couple of places,  maybe keep a global var with the version  ?

clobrano:
> I might be misunderstanding but I think it might be relevant here (see below - credentials.go L18-22) as well
> 
> ```
> return credentials.NewTLS(&tls.Config{
> 		Certificates: []tls.Certificate{*keyPair},
> 		ClientAuth:   tls.RequireAndVerifyClientCert,
> 		ClientCAs:    pool,
> 	}), nil
> ```
> 
> Also since it's being used in couple of places, maybe keep a global const with the version ?

Interesting, these two haven't been found by the scanner

clobrano:
> Interesting, these two haven't been found by the scanner

nevermind, at least one of them have been found by the scanner, my bad

pkg/controlplane/manager.go
```diff
@@ -150,7 +150,10 @@ func (manager *Manager) isEndpointAccessible() bool {
 func (manager *Manager) isKubeletServiceRunning() bool {
 	url := fmt.Sprintf("https://%s:%s/pods", manager.nodeName, kubeletPort)
 	tr := &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+			MinVersion:         tls.VersionTLS13,
```
mshitrit, line 155: can we use the const here as well ?

clobrano: I wonder if `credentials` is the right package where to store this. Maybe a new file under `utils`?

mshitrit: IMO `credentials` is a decent fit since this field has to do with `credentials`.
`utils` is more general so usually I'd try to avoid it when a better fit is possible.
Not a deal breaker as far as I'm concerned, so  if you strongly feel different that's fine too.

clobrano: I agree with `credentials`

clobrano:
I wonder if there is something that is making this test fail. I reproduce the failure also in my cluster (and the same test on the same cluster passes without the change of this PR), but I was waiting for FF to have a deeper analysis

PR #136, opened 2023-07-16
[WIP] use sysrq-trigger for software reboot

mshitrit:


pkg/reboot/rebooter.go
```diff
@@ -82,7 +82,7 @@ func (r *watchdogRebooter) Reboot() error {
 func (r *watchdogRebooter) softwareReboot() error {
 	r.log.Info("about to try software reboot")
 	// hostPID: true and privileged:true required to run this
-	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot", "--force", "--force")
+	rebootCmd := exec.Command("bash", "-c", "echo c > /proc/sysrq-trigger")
```
k-keiichi-rh, line 85: The 'c' command triggers a crashdump. If kdump is not enabled in the os, it will be stuck.
So I think the 'b' is fine to do emergency restart.

mshitrit: I'm far from being a linux expert but I found the following (see below) [here](https://www.kernel.org/doc/html/v4.18/admin-guide/sysrq.html)
```
b | Will immediately reboot the system without syncing or unmounting your disks.
c | Will perform a system crash by a NULL pointer dereference. A crashdump will be taken if configured.
```

IIUC it's mandatory to configure the crashdump.

Also in that context I'm not sure what is a difference between a `reboot` and a `system crash` (i.e whether the successful execution of one of them is more reliable )

k-keiichi-rh: I updated my understanding to the latest RHEL9 code(5.14.0-284.23.1.el9_2 used by OCP4.13.6).

Please let me share the ways of software reboot(reboot, softdog reset, panic() in kernel and so on):
We have the following ways to software reboot. It's listed in order of reliability from top to bottom:
1. softdog reset
2. echo b > /proc/sysrq-trigger
3. echo c > /proc/sysrq-trigger & enabling kdump with reboot action after collecting a dump

Compared with 1 and 2, both of them will reboot with the same way(the best effort to reboot in kernel space), but the "2" needs to write 'c' to /proc/sysrq-trigger in user space. So the "1" is more reliable than the "2"

Compared with 2 and 3, the "3" will reboot with "systemctl reboot -f" after collecting a memory dump. This way is a safer way to reboot than the reboot way of "2" and "3" because it performs a clean reboot like hw resetting and so on.  However the reboot way of "2" and "3" is  more reliable. it focusses on rebooting, not saving server.

3. echo c > /proc/sysrq-trigger & "kernel.panic = <secs before rebooting>" in sysctl.conf & disabling kdump

k-keiichi-rh: We have another option:

5. echo c > /proc/sysrq-trigger & "kernel.panic = <secs before rebooting>" in sysctl.conf & disabling kdump

Without enabling kdump, we have to wait forever after "echo c > /proc/sysrq-trigger".
However the /proc/sys/kernel/panic will help us to change this situation. If the timeout value is set to the /proc/sys/kernel/panic, the node will be rebooted with the same reboot way of "1" and "2" after waiting for the timeout seconds.
We can reboot the node with the panic messages although the additional configuration to sysctl.conf is required.

To summarize:
```
1. softdog reset
2. echo b > /proc/sysrq-trigger
3. systemctl reboot -f -f(current)
4. echo c > /proc/sysrq-trigger & enabling kdump("systemctl reboot -f" after collecting a dump)
5. echo c > /proc/sysrq-trigger & "kernel.panic = <secs before rebooting>" in sysctl.conf & disabling kdump
```
IMO, we can replace the current method("3") with the "2".

And the "4" or "5" may be a good option if users can choose.
However, I think HW watchdog for baremetal envs and softdog for general envs(virtual envs) will cover the most of use cases. I am not sure if collecting debug information is required when a failover to software reboot occurs.

Reliability : we can reboot 
1 > 2 > 5 > 3 > 4
Traceability : debugging and troubleshooting
4 > 5 >3 > 2 > 1
Safety : protecting servers
4 > 3 > 5, 2, 1
Generality : less configuration and less maintenance
2, 3 > 5 > 1 > 4

mshitrit: Thanks @k-keiichi-rh ,

your research and effort is much appreciated ! :1st_place_medal: 
Based on that I agree with your recommendation, and I've made the changes.

razo7:
/lgtm
Holding so @k-keiichi-rh can have more time for another review (very interesting research and findings) 
/hold

pkg/reboot/rebooter.go
```diff
@@ -82,7 +82,7 @@ func (r *watchdogRebooter) Reboot() error {
 func (r *watchdogRebooter) softwareReboot() error {
 	r.log.Info("about to try software reboot")
 	// hostPID: true and privileged:true required to run this
-	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot", "--force", "--force")
+	rebootCmd := exec.Command("bash", "-b", "echo c > /proc/sysrq-trigger")
```
beekhof, line 85: do we need the full path to bash?

beekhof: And maybe it needs to be done inside an `nsenter` context?

 ```suggestion
	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "bash", "-b", "echo c > /proc/sysrq-trigger")
```

File: pkg/reboot/rebooter.go (hunk 1 above)
Comment (slintes) [reply pkg/reboot/rebooter.go]: Also, would it make sense to use more commands before the reboot?

From https://www.kernel.org/doc/html/v4.18/admin-guide/sysrq.html#okay-so-what-can-i-use-them-for:

>  I generally sync(s), umount(u), then reboot(b) when my system locks. It’s saved me many a fsck

Or from https://www.thegeekstuff.com/2008/12/safe-reboot-of-linux-using-magic-sysrq-key/:

> To perform a safe reboot of a Linux computer which hangs up, do the following. This will avoid the fsck during the next re-booting. i.e Press Alt+SysRq+letter highlighted below.

> ~unRaw (take control of keyboard back from X11,~
tErminate (send SIGTERM to all processes, allowing them to terminate gracefully),
kIll (send SIGILL to all processes, forcing them to terminate immediately),
Sync (flush data to disk),
Unmount (remount all filesystems read-only),
reBoot.

Comment (k-keiichi-rh) [reply pkg/reboot/rebooter.go]: > I understand this means "b" is the best option for us, not "c" like in the updated code? 🤔

Yes, I think "b" is the best option in this case.

> why does this work at all on OCP? I understand you can configure what sysrq is allowed to do in /proc/sys/kernel/sysrq, and on my current cluster it has value 16, which doesn't include reboots according to https://www.kernel.org/doc/html/v4.18/admin-guide/sysrq.html#how-do-i-enable-the-magic-sysrq-key. 

Because this checking(/proc/sys/kernel/sysrq) you mentioned is bypassed when using /proc/sysrq-trigger.
It's for using the magic sysrq key in our keyboard. And there is no way  for /proc/sysrq-trigger to restrict triggering some actions like "b" and "c".

Please refer to https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/sysrq.rst for more information.
This is the doc for the latest kernel, but I confirmed we can say same thing for the RHEL8 and RHEL9 kernels.

**Comment by slintes** [comment]
/lgtm cancel

there are comments which need an answer at least

File: pkg/reboot/rebooter.go (hunk 1 above)
Comment (k-keiichi-rh) [reply pkg/reboot/rebooter.go]: > Also, would it make sense to use more commands before the reboot?

If I understand correctly, I think the 1st priority thing for SNR is to reset the failed node to schedule workloads on the failed node to another node.
So we don't need more commands like sync(s), umount(u) for safe reboot via /proc/sysrq-trigger. 
These safe commands don't help us to improve reliability for rebooting the node.

And either the softdog reset or the HW watchdog reset won't do anything to save the server.
It think it's difficult to imagine that soft reboot as a failover action needs the safe actions.

File: pkg/reboot/rebooter.go (hunk 2 above)
Comment (mshitrit) [reply pkg/reboot/rebooter.go:85]: >And maybe it needs to be done inside an nsenter context?

Not an expert on `nsenter` , but from the [little reading](https://man7.org/linux/man-pages/man1/nsenter.1.html) I've done my understanding is that  the  `mnt` [option](https://man7.org/linux/man-pages/man7/mount_namespaces.7.html)  would create an isolated namespace for the command to run in.

Since in any case we plan to reboot the whole machine is there a value in that separation which  I'm missing ? (my initial intuition is that it's just an unneeded overhead)

File: pkg/reboot/rebooter.go (hunk 1 above)
Comment (k-keiichi-rh) [reply pkg/reboot/rebooter.go]: We may have a reason to use the more commands in terms of the compatibility with "reboot -f -f" in the current implementation.

Comment (slintes) [reply pkg/reboot/rebooter.go]: > Because this checking(/proc/sys/kernel/sysrq) you mentioned is bypassed when using /proc/sysrq-trigger.

Ok 👍🏼 

> These safe reboot commands don't help us to improve reliability for rebooting the node.

Makes sense.

Thanks for clarifying!

File: pkg/reboot/rebooter.go (hunk 2 above)
Comment (slintes) [reply pkg/reboot/rebooter.go:85]: > my understanding is that the mnt [option](https://man7.org/linux/man-pages/man7/mount_namespaces.7.html) would create an isolated namespace for the command to run in.

no, you don't create that namespace, you *enter* it. With `-m/proc/1/ns/mnt` you enter the "mount" namespace of the process with PID `1`. That's the host' s very first process. This gives you access to binaries and files which are installed on the host but not in the container you are running in, like `/bin/systemctl` in the old version, or `/bin/bash` and `/proc/sysrq-trigger` in your new version. That's what I understand at least. So while bash *might*  be installed in the container as well, I'm not sure if/how `/proc/sysrq-trigger` works in containers...

File: pkg/reboot/rebooter.go [hunk 3]
```diff
@@ -82,7 +82,7 @@ func (r *watchdogRebooter) Reboot() error {
 func (r *watchdogRebooter) softwareReboot() error {
 	r.log.Info("about to try software reboot")
 	// hostPID: true and privileged:true required to run this
-	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot", "--force", "--force")
+	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "bash", "-b", "echo c > /proc/sysrq-trigger")
```
slintes, line 85: I think it should be `"/bin/bash", "-c", "echo b..."`

PR #138, opened 2023-08-03
Stop remediation when node not found

mshitrit:
[ECOPROJECT-1492](https://issues.redhat.com//browse/ECOPROJECT-1492)

Incorrect Status Conditions in case non existed node in Remediation CR .
When SNR is created for a non existing node Processing condition is True and Reason is RemediationStarted. 
Expected values are processing = false, succeeded = false with reason NodeNotFound.

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -76,7 +76,7 @@ type processingChangeReason string
 
 const (
 	remediationStarted              processingChangeReason = "RemediationStarted"
-	remediationTerminatedByNHC      processingChangeReason = "RemediationStoppedByNHC"
+	remediationTimeoutByNHC         processingChangeReason = "RemediationTimeoutByNHC"
 	remediationFinishedSuccessfully processingChangeReason = "RemediationFinishedSuccessfully"
 	remediationFinishedNodeNotFound processingChangeReason = "RemediationFinishedNodeNotFound"
 )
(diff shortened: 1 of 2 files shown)


clobrano:
/lgtm

/hold

controllers/selfnoderemediation_controller.go
```diff
@@ -75,9 +75,10 @@ var (
 type processingChangeReason string
 
 const (
-	remediationStarted         processingChangeReason = "RemediationStarted"
-	remediationTerminatedByNHC processingChangeReason = "RemediationStoppedByNHC"
-	remediationFinished        processingChangeReason = "RemediationFinished"
+	remediationStarted              processingChangeReason = "RemediationStarted"
+	remediationTerminatedByNHC      processingChangeReason = "RemediationStoppedByNHC"
```
razo7, line 79: NIT: I would follow the pattern of the const name match its value by updating the const name to be `remediationStoppedByNHC `

razo7:
NIT: Whenever we use `verifyTypeConditions` (in the UT file) the expected reason is explicitly written. Why not use the const from `selfnoderemediation_controller.go` -> Use `RemediationStarted` const instead of `"RemediationStarted"`

controllers/selfnoderemediation_controller.go
slintes: and NHC is signalling a timeout actually, not stopping or terminating anything 😉

slintes:
> NIT: Whenever we use `verifyTypeConditions` (in the UT file) the expected reason is explicitly written. Why not use the const from `selfnoderemediation_controller.go` -> Use `RemediationStarted` const instead of `"RemediationStarted"`

when you reuse the same const, you don't detect typos in const values

controllers/selfnoderemediation_controller.go
razo7: Too many verbs for one action :D

PR #139, opened 2023-09-03
configure host port

mshitrit:
[ECOPROJECT-1581](https://issues.redhat.com//browse/ECOPROJECT-1581)
Currently the hostPort: 30001 is fixed by the operator and is no possible to change, making it impossible for pods to deploy if that port is already in use.
This PR makes the port configurable.

### DIFF
--- a/api/v1alpha1/selfnoderemediationconfig_types.go
+++ b/api/v1alpha1/selfnoderemediationconfig_types.go
@@ -114,6 +114,12 @@ type SelfNodeRemediationConfigSpec struct {
 	// It will be ignored when empty (which is the default).
 	EndpointHealthCheckUrl string `json:"endpointHealthCheckUrl,omitempty"`
 
+	// HostPort is used for internal communication between SNR agents.
+	// +optional
+	// +kubebuilder:default:=30001
+	// +kubebuilder:validation:Minimum=1
+	HostPort int `json:"hostPort,omitempty"`
+
 	// CustomDsTolerations allows to add custom tolerations snr agents that are running on the ds in order to support remediation for different types of nodes.
 	CustomDsTolerations []v1.Toleration `json:"customDsTolerations,omitempty"`
 }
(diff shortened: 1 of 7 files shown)


api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -114,6 +114,12 @@ type SelfNodeRemediationConfigSpec struct {
 	// It will be ignored when empty (which is the default).
 	EndpointHealthCheckUrl string `json:"endpointHealthCheckUrl,omitempty"`
 
+	// HostPort is used for internal communication between SNR agents.
+	// +optional
+	// +kubebuilder:default:=30001
+	// +kubebuilder:validation:Minimum=1
```
clobrano, line 120: Should we also enforce only 1 port value here?

mshitrit: I don't follow :sweat_smile: 
ATM 30001 is hard coded, the suggest change make it configurable while leaving 30001 as the default choice.
What do you suggest ?

clobrano: The configuration should also allow no more than 1 value? 
I saw in the code you also check that the list contains only 1 value (in a test), so maybe we can also set `maximum=1`?

mshitrit: I think I understand now, IIUC there is a bit of a mix-up.
It is true that container may have several ports, but it's not relevant to this PR because we create this container according to the `ds` and it will always have one port.
This validation is about the configuration value  of that port (i.e HostPort) and not the number of ports  - so in a nutshell it states that the configured port must have a positive number.

clobrano: Oh, I misunderstood then, thanks for the clarification!

clobrano:
/lgtm
giving a chance to get more feedback before merging,feel free to un-hold
/hold

PR #140, opened 2023-09-06
config value of safe time to reboot was calculated wrongly

mshitrit:
[ECOPROJECT-1630](https://issues.redhat.com//browse/ECOPROJECT-1630)
safeTimeToAssumeNodeRebootedSeconds doesn't impact actual time of node returning to health.

### DIFF
--- a/main.go
+++ b/main.go
@@ -250,9 +250,9 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 	apiServerTimeout := getDurEnvVarOrDie("API_SERVER_TIMEOUT")       //timeout for each api-connectivity check
 	peerDialTimeout := getDurEnvVarOrDie("PEER_DIAL_TIMEOUT")         //timeout for establishing connection to peer
 	peerRequestTimeout := getDurEnvVarOrDie("PEER_REQUEST_TIMEOUT")   //timeout for each peer request
-	timeToAssumeNodeRebooted := getDurEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
+	timeToAssumeNodeRebootedInSeconds := getIntEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -250,9 +250,9 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 	apiServerTimeout := getDurEnvVarOrDie("API_SERVER_TIMEOUT")       //timeout for each api-connectivity check
 	peerDialTimeout := getDurEnvVarOrDie("PEER_DIAL_TIMEOUT")         //timeout for establishing connection to peer
 	peerRequestTimeout := getDurEnvVarOrDie("PEER_REQUEST_TIMEOUT")   //timeout for each peer request
-	timeToAssumeNodeRebooted := getDurEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
+	timeToAssumeNodeRebootedInSeconds := getIntEnvVarOrDie("TIME_TO_ASSUME_NODE_REBOOTED")
```
clobrano, line 253: I don't understand why using `getIntEnvVarOrDie` here in place of `getDurEnvVarOrDie`.
It seems to me that the only difference is that later on the variable is multiplied by `time.Seconds`, but that should be possible with the output of `getDurEnvVarOrDie` as well, I think.

mshitrit: You are correct in the sense that both will work.
However the bug originated from the fact that we've treated an int (which represented the number of seconds) as Duration.
So instead of 180 (which represent number of seconds) we'd have 180 Nano Seconds which is what we get when changed into duration.
So in a nutshell I'm saying that even though both would work, in this case it would be more logically correct to treat this number as an int and not duration.

clobrano: I understand now, and agree. Thanks for the explanation!

clobrano:
/lgtm 
Giving others a chance to review as well, feel free to un-hold
/hold

PR #141, opened 2023-09-13
Handle large cluster load

mshitrit:
[ECOPROJECT-1261](https://issues.redhat.com//browse/ECOPROJECT-1261)
At the moment, upon a remediation detected by SNR all DaemSets (i.e all SNR agent - each on every node) will perform APIServer calls as part of SNR remediation process.
In large clusters that can cause an overload on the APIServer.

We need to consider a solution to reduce the amount of APIServer call in order to minimize that risk.

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -130,6 +130,7 @@ type SelfNodeRemediationReconciler struct {
 	//40s of grace period for the node to reappear before it deletes the pods.
 	//see here: https://github.com/kubernetes/kubernetes/blob/7a0638da76cb9843def65708b661d2c6aa58ed5a/pkg/controller/podgc/gc_controller.go#L43-L47
 	RestoreNodeAfter time.Duration
+	reboot.SafeTimeCalculator
 }
 
 // SetupWithManager sets up the controller with the Manager.
@@ -153,6 +154,14 @@ func (r *SelfNodeRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error
 
 func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (returnResult ctrl.Result, returnErr error) {
 	r.logger = r.Log.WithValues("selfnoderemediation", req.NamespacedName)
+
+	if r.IsAgent() {
+		if req.Name != r.MyNodeName {
+			r.logger.Info("agent pod skipping remediation because node belongs to a different agent")
+			return ctrl.Result{}, nil
+		}
+		r.logger.Info("agent pod starting remediation on owned node")
+	}
 
 	snr := &v1alpha1.SelfNodeRemediation{}
 	if err := r.Get(ctx, req.NamespacedName, snr); err != nil {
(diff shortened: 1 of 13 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -154,6 +155,14 @@ func (r *SelfNodeRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error
 func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (returnResult ctrl.Result, returnErr error) {
 	r.logger = r.Log.WithValues("selfnoderemediation", req.NamespacedName)
 
+	if r.IsAgent() {
+		if req.Name != r.MyNodeName {
+			r.logger.Info("agent pod skipping remediation because node belongs to a different agent")
```
razo7, line 160: It would be nice to print the CR/node name (`req.Name`)

main.go
```diff
@@ -181,6 +188,29 @@ func initSelfNodeRemediationManager(mgr manager.Manager) {
 		setupLog.Error(err, "failed to add template creator to the manager")
 		os.Exit(1)
 	}
+
+	myNodeName := os.Getenv(nodeNameEnvVar)
+	if myNodeName == "" {
+		setupLog.Error(errors.New("failed to get own node name"), "node name was empty",
```
clobrano, line 194: Is it ok to continue if it cannot find the node's name?
IIUC, for Agents it quite necessary

clobrano: nevermind, just noticed that we're initializing the manager here :)

pkg/reboot/calculator.go
```diff
@@ -78,6 +105,17 @@ func (s *safeTimeCalculator) calcMinTimeAssumeRebooted() {
 	minTime += 15 * time.Second
 	s.log.Info("calculated minTimeToAssumeNodeRebooted is:", "minTimeToAssumeNodeRebooted", minTime)
 	s.minTimeToAssumeNodeRebooted = minTime
+
+	if s.timeToAssumeNodeRebooted < minTime {
+		err := fmt.Errorf("snr agent can't start requested value for timeToAssumeNodeRebooted is too low")
```
clobrano, line 110: NIT: it took me a couple of seconds to understand the error message. Could you consider to use something to separate the problem from it's cause?
Something like:

"snr agent can't start: requested value for timeToAssumeNodeRebooted is too low"

main.go
mshitrit: I agree, but I think the location is wrong - this is where the manager is initilized. 
The fix should be where the agent is initialized: [here](https://github.com/medik8s/self-node-remediation/pull/141/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L205-R239)

clobrano (approved):
/lgtm, leaving a chance to get more feedbacks /hold

clobrano:
/lgtm 
/hold

(inline commands do not work sorry)

mshitrit:
test 4.14-openshift-e2e

mshitrit:
/override 4.12-openshift-e2e
(already passed re-run by mistake)

mshitrit:
/override ci/prow/4.12-openshift-e2e
(already passed re-run by mistake)

PR #143, opened 2023-10-13
Provide TLS Certificate to kubelet service status request

clobrano:
This change enhances the security of the kubelet service request in the
control plane manager by providing a TLS certificate when making the
request.

- Provide a CertStorageReader to the Manager
- Re-use PrepareCredentials function from certificates package to create
  the proper certificates
- Change TLSConfig providing Certificates and setting InsecureSkipVerify
  to false

see https://issues.redhat.com/browse/ECOPROJECT-1421


clobrano:
wrong source branch

PR #144, opened 2023-10-13
[WIP] Provide TLS Certificate to kubelet service status request

clobrano:
This change enhances the security of the kubelet service request in the
control plane manager by providing a TLS certificate when making the
request.

- Provide a CertStorageReader to the Manager
- Re-use PrepareCredentials function from certificates package to create
  the proper certificates
- Change TLSConfig providing Certificates and setting InsecureSkipVerify
  to false

see https://issues.redhat.com/browse/ECOPROJECT-1421

### DIFF
--- a/main.go
+++ b/main.go
@@ -311,7 +311,7 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 		MaxTimeForNoPeersResponse: reboot.MaxTimeForNoPeersResponse,
 	}
 
-	controlPlaneManager := controlplane.NewManager(myNodeName, mgr.GetClient())
+	controlPlaneManager := controlplane.NewManager(myNodeName, mgr.GetClient(), certReader)
 
 	if err = mgr.Add(controlPlaneManager); err != nil {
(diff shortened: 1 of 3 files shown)


mshitrit:
/lgtm
/hold
giving chance for others to review

slintes:
I'n not confident that this works tbh. The certificate you use here is self signed, and it's intended to be used by SNR's peer health check, so for connections between SNR pods only. It's not a cert created by the API server. So I'm not sure if kubelet will actually accept the cert. 

Since there isn't a test case which uses the modified code, test results are useless.
Did you verify that this is works in a manual test?

/hold

slintes:
I also expect `InsecureSkipVerify: false,` to fail, because we don't have a cluster CA cert to verify kubelet's server cert

clobrano:
> Since there isn't a test case which uses the modified code, test results are useless.

I didn't consider this, I expected e2e test to cover this case. I'll figure out a new test then

clobrano:
This PR needs rework, since the solution proposed does not work

clobrano:
Closing the PR while rethinking the solution

PR #154, opened 2023-10-25
print versions

mshitrit:
Adding info about go version (and additional info) in logs.
[ECOPROJECT-1743](https://issues.redhat.com//browse/ECOPROJECT-1743)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream8 AS builder
-RUN yum install golang -y && yum clean all
+RUN yum install git golang -y && yum clean all
 
 # Ensure correct Go version
 ENV GO_VERSION=1.20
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -13,23 +13,23 @@ WORKDIR /workspace
...
 # Copy the go source
+COPY vendor/ vendor/
+COPY version/ version/
 COPY main.go main.go
+COPY hack/ hack/
 COPY api/ api/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
 COPY install/ install/
 # Build
```
razo7, line 26: IIRC you are missing `git` which should be installed in line 3 and copied right before building the binary.
```
# for getting version info
COPY .git/ .git/
```

razo7 (changes requested):
I think `git` is missing for getting the git commit and the operator's version.

In addition, the[ `build` target](https://github.com/medik8s/self-node-remediation/blob/main/Makefile#L181-L183) could be updated according to this PR. But I haven't seen it being used in code, so this target might be redundant 🤷🏻‍♂️

PR #155, opened 2023-10-26
Do not create PeerHealth new client without credentials

clobrano:
Currently if peerhealth.NewClient is called with no
credentials.TransportCredentials it uses a DialOption which disables
transport security for this ClientConn.

Considering also the in the current code base it is not possible to
reach this point without valid credentials, this commit removes the
usage of such DialOption and returns with error instead for security
reasons.


razo7:
/lgtm
/hold
giving a chance for others to review

PR #158, opened 2023-10-29
remove VA deletion

mshitrit:
[ECOPROJECT-1748](https://issues.redhat.com//browse/ECOPROJECT-1748)
VA should not be deleted by SNR. kube-control-manager (KCM) detects the pod deletion and automatically detaches the VAs which are not connected to any pod, forcefully deleting the VA has the risk of interfering with the process of detaching the VAs.

### DIFF
--- a/controllers/tests/controller/selfnoderemediation_controller_test.go
+++ b/controllers/tests/controller/selfnoderemediation_controller_test.go
@@ -60,12 +60,11 @@ var _ = Describe("SNR Controller", func() {
 
 	AfterEach(func() {
 		k8sClient.ShouldSimulateFailure = false
-		k8sClient.ShouldSimulateVaFailure = false
+		k8sClient.ShouldSimulatePodDeleteFailure = false
 		isAdditionalSetupNeeded = false
 		deleteRemediations()
 		deleteSelfNodeRemediationPod()
 		deleteVolumeAttachment(vaName, false)
-		verifyVaDeleted(vaName)
 		//clear node's state, this is important to remove taints, label etc.
 		Expect(k8sClient.Update(context.Background(), getNode(shared.UnhealthyNodeName)))
 		Expect(k8sClient.Update(context.Background(), getNode(shared.PeerNodeName)))
@@ -184,8 +183,6 @@ var _ = Describe("SNR Controller", func() {
 
 				verifySelfNodeRemediationPodDoesntExist()
 
-				verifyVaDeleted(vaName)
-
 				verifyEvent("Normal", "DeleteResources", "Remediation process - finished deleting unhealthy node resources")
 
 				verifyFinalizerExists()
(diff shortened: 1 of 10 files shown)


razo7:
Would it make more sense to move `deleteResources` function into the common repo, then [FAR](https://github.com/medik8s/fence-agents-remediation/tree/main) and SNR can use it as well for [resource deletion](https://github.com/medik8s/fence-agents-remediation/blob/main/pkg/utils/resources.go)?

mshitrit:
> Would it make more sense to move `deleteResources` function into the common repo, then [FAR](https://github.com/medik8s/fence-agents-remediation/tree/main) and SNR can use it as well for [resource deletion](https://github.com/medik8s/fence-agents-remediation/blob/main/pkg/utils/resources.go)?

I have no strong opinion either way.
On the one hand it's always a good thing avoiding code duplication but on the other hand I am not sure the overhead will be worth the value (specifically because it's a pretty short method).

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -391,20 +388,6 @@ func deleteVolumeAttachment(vaName string, verifyExist bool) {
 	ExpectWithOffset(1, err).To(Succeed())
 }
 
-func verifyVaDeleted(vaName string) {
```
razo7, line 394: Is there a reason not to verify VA deletion?

mshitrit: VA are not deleted anymore, it's now up to KCM to handle the VAs

razo7: Also in unit testing? It makes sense not to test that in e2e which resembles reality but in unit-test, there is no KCM, right?

mshitrit: > in unit-test, there is no KCM, right?

That is correct, but because of the fact that there is no KCM the VA will never be deleted (SNR does not delete the VA anymore, and KCM will not delete them), so it make no sense to verify they are deleted - this sort of verification will always fail.

razo7: Oh, now I get it.

razo7:
> > Would it make more sense to move `deleteResources` function into the common repo, then [FAR](https://github.com/medik8s/fence-agents-remediation/tree/main?rgh-link-date=2023-10-29T16%3A46%3A24Z) and SNR can use it as well for [resource deletion](https://github.com/medik8s/fence-agents-remediation/blob/main/pkg/utils/resources.go?rgh-link-date=2023-10-29T16%3A46%3A24Z)?
> 
> I have no strong opinion either way. On the one hand it's always a good thing avoiding code duplication but on the other hand I am not sure the overhead will be worth the value (specifically because it's a pretty short method).

Can you update commons and use `DeletePods` from https://github.com/medik8s/common/releases/tag/v1.9.0? Other than that looks good for me.

clobrano:
/lgtm

Just leaving the time to close the other threads
/hold

PR #160, opened 2023-11-07
Automatic remediation strategy

mshitrit:
[ECOPROJECT-1752](https://issues.redhat.com//browse/ECOPROJECT-1752)
At the moment the default remediation Strategy is `ResourceDeletion`, for newer clusters (ocp 4.14+) it's recommended to use the newer implementation of the Out of Service taint which is used in the `OutOfService` strategy.
In order to achieve this we plan to create a new default strategy `Automatic` which will decide the remediation strategy SNR will use at runtime depending on the cluster.
This change also meant for a better integration of other future strategies

### DIFF
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -24,7 +24,7 @@ const (
 	AutomaticRemediationStrategy         = RemediationStrategyType("Automatic")
 	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
 	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
-	DefaultRemediationStrategy           = ResourceDeletionRemediationStrategy
+	DefaultRemediationStrategy           = AutomaticRemediationStrategy
 	// ProcessingConditionType is the condition type used to signal NHC the remediation status
 	ProcessingConditionType = "Processing"
 	// SucceededConditionType is the condition type used to signal NHC whether the remediation was successful or not
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -21,6 +21,7 @@ import (
 )
 
 const (
+	AutomaticRemediationStrategy         = RemediationStrategyType("Automatic")
 	ResourceDeletionRemediationStrategy  = RemediationStrategyType("ResourceDeletion")
 	OutOfServiceTaintRemediationStrategy = RemediationStrategyType("OutOfServiceTaint")
 	// ProcessingConditionType is the condition type used to signal NHC the remediation status
@@ -41,8 +42,8 @@ type SelfNodeRemediationSpec struct {
 	//The first will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
 	//The latter will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
 	//that enables automatic deletion of pv-attached pods on failed nodes, "OutOfServiceTaint" is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
-	// +kubebuilder:default:="ResourceDeletion"
-	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
+	// +kubebuilder:default:="Automatic"
+	// +kubebuilder:validation:Enum=Automatic;ResourceDeletion;OutOfServiceTaint
 	RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
 }
 
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -42,7 +42,7 @@ type SelfNodeRemediationSpec struct {
 	//Currently, it could be either "Automatic", "OutOfServiceTaint" or "ResourceDeletion".
 	//ResourceDeletion will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
 	//OutOfServiceTaint will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
-	//that enables automatic deletion of pv-attached pods on failed nodes, "OutOfServiceTaint" is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
+	//that enables automatic deletion of pv-attached pods on failed nodes, "out-of-service" taint is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
 	// Automatic will choose the most appropriate strategy during runtime.
 	// +kubebuilder:default:="Automatic"
 	// +kubebuilder:validation:Enum=Automatic;ResourceDeletion;OutOfServiceTaint
--- a/api/v1alpha1/selfnoderemediation_types.go
+++ b/api/v1alpha1/selfnoderemediation_types.go
@@ -38,10 +38,11 @@ type RemediationStrategyType string
 // SelfNodeRemediationSpec defines the desired state of SelfNodeRemediation
 type SelfNodeRemediationSpec struct {
 	//RemediationStrategy is the remediation method for unhealthy nodes.
-	//Currently, it could be either "ResourceDeletion" or "OutOfServiceTaint".
-	//The first will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
-	//The latter will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
+	//Currently, it could be either "Automatic", "OutOfServiceTaint" or "ResourceDeletion".
+	//ResourceDeletion will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
+	//OutOfServiceTaint will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -276,6 +276,10 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
 		wasWatchdogInitiated = true
 	}
 
+	if err := utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
```
slintes, line 279: this is done for manager and agent now, maybe move into main() instead?

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -41,8 +42,8 @@ type SelfNodeRemediationSpec struct {
 	//The first will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
 	//The latter will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
 	//that enables automatic deletion of pv-attached pods on failed nodes, "OutOfServiceTaint" is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
-	// +kubebuilder:default:="ResourceDeletion"
-	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
+	// +kubebuilder:default:="Automatic"
```
slintes, line 45: comment above needs an update :)

slintes (changes requested):
Pretty much straight forward :)
Two nits inside, otherwise lgtm

slintes:
lgtm, but needs rebase

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -38,12 +39,13 @@ type RemediationStrategyType string
...
-	//Currently, it could be either "ResourceDeletion" or "OutOfServiceTaint".
-	//The first will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
-	//The latter will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
+	//Currently, it could be either "Automatic", "OutOfServiceTaint" or "ResourceDeletion".
+	//ResourceDeletion will iterate over all pods and VolumeAttachment related to the unhealthy node and delete them.
+	//OutOfServiceTaint will add the out-of-service taint which is a new well-known taint "node.kubernetes.io/out-of-service"
 	//that enables automatic deletion of pv-attached pods on failed nodes, "OutOfServiceTaint" is only supported on clusters with k8s version 1.26+ or OCP/OKD version 4.13+.
-	// +kubebuilder:default:="ResourceDeletion"
-	// +kubebuilder:validation:Enum=ResourceDeletion;OutOfServiceTaint
+	// Automatic will choose the most appropriate strategy during runtime.
```
razo7, line 46: This is kind of vague. Is that on purpose or could we just write which remediation will be used most of the time?

``` suggestion
Automatic will choose the most appropriate strategy during runtime. By default, OutOfServiceTaint will be selected If the running Kubernetes can support the new taint, otherwise, ResourceDeletion will be selected.
```

controllers/selfnoderemediation_controller.go
```diff
@@ -907,3 +912,18 @@ func (r *SelfNodeRemediationReconciler) isResourceDeletionExpired(snr *v1alpha1.
...
+	if strategy != v1alpha1.AutomaticRemediationStrategy {
+		return strategy
+	}
+
+	if utils.IsOutOfServiceTaintGA {
+		r.logger.Info("Automatically selected OutOfServiceTaint Remediation strategy")
+		return v1alpha1.OutOfServiceTaintRemediationStrategy
+	}
+
+	r.logger.Info("Automatically selected ResourceDeletion Remediation strategy")
```
razo7, line 926: Wouldn't be helpful to state that the selected strategy was selected because the other one is not supported?

main.go
```diff
@@ -150,10 +154,6 @@ func initSelfNodeRemediationManager(mgr manager.Manager, enableHTTP2 bool) {
 
 	configureWebhookServer(mgr, enableHTTP2)
 
-	if err := utils.InitOutOfServiceTaintSupportedFlag(mgr.GetConfig()); err != nil {
-		setupLog.Error(err, "unable to verify out of service taint support. out of service taint isn't supported")
```
razo7, line 154: NIT: Sometimes we refer to the taint as `out of service` with spaces and sometimes without spaces `OutOfService` in comments and prints. I don't have a strong preference for these two, but I see that `out-of-service` is used in the [official Kubernetes documentation](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-out-of-service). So if we want to change it or to align ourself, thus I would suggest using `out-of-service`.
We can probably do this change under [common](https://github.com/medik8s/common/tree/main/pkg) and then SNR and FAR can use it after the [FAR out-of-service PR](https://github.com/medik8s/fence-agents-remediation/pull/92) is merged.

I also think that this kind of PR will be meaningful for FAR eventually.

pkg/utils/taints.go
```diff
@@ -79,7 +84,9 @@ func setOutOfTaintSupportedFlag(version *version.Info) error {
 		return err
 	}
 
-	IsOutOfServiceTaintSupported = majorVer > minK8sMajorVersionSupportingOutOfServiceTaint || (majorVer == minK8sMajorVersionSupportingOutOfServiceTaint && minorVer >= minK8sMinorVersionSupportingOutOfServiceTaint)
+	IsOutOfServiceTaintSupported = majorVer > minK8sMajorVersionOutOfServiceTaint || (majorVer == minK8sMajorVersionOutOfServiceTaint && minorVer >= minK8sMinorVersionSupportingOutOfServiceTaint)
 	utilTaintsLog.Info("out of service taint strategy", "isSupported", IsOutOfServiceTaintSupported, "k8sMajorVersion", majorVer, "k8sMinorVersion", minorVer)
+	IsOutOfServiceTaintGA = majorVer > minK8sMajorVersionOutOfServiceTaint || (majorVer == minK8sMajorVersionOutOfServiceTaint && minorVer >= minK8sMinorVersionGAOutOfServiceTaint)
+	utilTaintsLog.Info("out of service taint strategy", "isGA", IsOutOfServiceTaintGA, "k8sMajorVersion", majorVer, "k8sMinorVersion", minorVer)
```
razo7, line 90: IIUC for k8s version 1.28+ SNR would have two prints, one for that out-of-service taint being supported and one for being GA. Can we maybe only print that it is GA in the case of k8s version 1.28+ and when the k8s version is  either 1.27/1.26, then we would only print that out-of-service taint is supported?

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -158,10 +158,13 @@ var _ = Describe("SNR Controller", func() {
 			isAdditionalSetupNeeded = true
 		})
 
-		Context("ResourceDeletion strategy", func() {
+		Context("Automatic strategy - ResourceDeletion selected", func() {
 
 			BeforeEach(func() {
-				remediationStrategy = v1alpha1.ResourceDeletionRemediationStrategy
+				remediationStrategy = v1alpha1.AutomaticRemediationStrategy
+				prevVal := utils.IsOutOfServiceTaintGA
```
razo7, line 165: NIT: Why do we take the `IsOutOfServiceTaintGA` value instead of the `IsOutOfServiceTaintSupported` value? I believe it won't change the functionality of the unit tests, but it seems more accurate to test the supported version rather than the GA version

main.go
mshitrit: Not sure if I'd go as far as using common for text in log messages, but I agree that sticking to official docs with 'out-of-taint' is a good idea.
BTW this comment was misplaced - it's on removed line, but still relevant elsewhere 🙂

pkg/utils/taints.go
mshitrit: Hmm, I think I'm leaning into keeping as is, it seems like a bit of overcomplicating it without a real benefit.
Keep in mind that this log is only printed only once when the operator starts.

controllers/tests/controller/selfnoderemediation_controller_test.go
mshitrit: This change is to allow the test to test AutomaticRemediationStrategy, that strategy is only affected by IsOutOfServiceTaintGA value.

controllers/selfnoderemediation_controller.go
mshitrit: Basically the same rational as in the prev [comment](https://github.com/medik8s/self-node-remediation/pull/160#discussion_r1391104944)

pkg/utils/taints.go
razo7: SGTM

main.go
razo7: >BTW this comment was misplaced - it's on removed line, but still relevant elsewhere 

Correct, and there few more places that we print `outofservice` if you want to fix:

- [https://github.com/medik8s/self-node-remediation/pull/160/files#diff-dd6f1328b9432817228e47edffdc3fe1e4570bfa2e39099223c05131e89e9068R871 and ](https://github.com/medik8s/self-node-remediation/pull/160/files#diff-dd6f1328b9432817228e47edffdc3fe1e4570bfa2e39099223c05131e89e9068R870-R871)
- https://github.com/medik8s/self-node-remediation/pull/160/files#diff-dd6f1328b9432817228e47edffdc3fe1e4570bfa2e39099223c05131e89e9068R848-R849

razo7 (approved):
/lgtm
left small a NIT for replacing to the `out-of-service` in more places. Not so important, thus feel free to ignore. I hold it just in case you want to fix it now.
/hold

main.go
mshitrit: Thanks, good catch ! - fixed

PR #163, opened 2023-11-23
adding node selector to snr daemonset

mshitrit:
[ECOPROJECT-1747](https://issues.redhat.com//browse/ECOPROJECT-1747)
Adding a node selector which will allow preventing SNR daemonset to run on a specific node.
This functionality is required for use cases where we would like to disable SNR (for example Node maintenance).


clobrano:
/lgtm 
/hold

PR #164, opened 2023-11-28
SNR wrongly identify the remediation created by MHC (instead of NHC)

mshitrit:
[ECOPROJECT-1772](https://issues.redhat.com//browse/ECOPROJECT-1772)
SNR Misinterprets CR Association when template configures in openshift-machine-api ns, leading to Incorrect peer check in case of API issues.

When a SNR template is configured in the openshift-machine-api for use with NHC, an issue arises. SNR incorrectly assumes that CRs  are generated by MHC and wrongly identifies the CR name as the Machine name. 
However, the CR name corresponds to the Node name.This misinterpretation becomes problematic during peer checks,in case of API issues. 
SNR searches for CRs using the incorrect Machine name instead of the Node name. 

Consequently, it consistently returns a "you're healthy" status, as it fails to locate the expected CR associated with the Node name.

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -87,7 +87,7 @@ var (
 	}
 
 	lastSeenSnrNamespace  string
-	wasLastSeenSnrMachine bool
+	isSnrMatchMachineName bool
 )
 
 type processingChangeReason string
@@ -124,12 +124,12 @@ func (r *SelfNodeRemediationReconciler) GetLastSeenSnrNamespace() string {
 	return lastSeenSnrNamespace
 }
 
-// WasLastSeenSnrMachine returns a boolean indicating if the last reconcile SNR
+// IsSnrMatchMachineName returns a boolean indicating if the last reconcile SNR
 // was pointing an unhealthy machine or a node
-func (r *SelfNodeRemediationReconciler) WasLastSeenSnrMachine() bool {
+func (r *SelfNodeRemediationReconciler) IsSnrMatchMachineName() bool {
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
-	return wasLastSeenSnrMachine
+	return isSnrMatchMachineName
 }
 
 // SelfNodeRemediationReconciler reconciles a SelfNodeRemediation object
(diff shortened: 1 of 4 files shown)


controllers/selfnoderemediation_controller.go
```diff
@@ -644,13 +644,14 @@ func (r *SelfNodeRemediationReconciler) getNodeFromSnr(snr *v1alpha1.SelfNodeRem
...
-		if ownerRef.Kind == "Machine" {
-			r.mutex.Lock()
-			wasLastSeenSnrMachine = true
-			r.mutex.Unlock()
-			return r.getNodeFromMachine(ownerRef, snr.Namespace)
+	if !r.isOwnedByNHC(snr) {
+		for _, ownerRef := range snr.OwnerReferences {
+			if ownerRef.Kind == "Machine" {
+				r.mutex.Lock()
+				wasLastSeenSnrMachine = true
```
slintes, line 651: - comment above needs an update, that NHC potentially also set machine owner
- the name of this var (and its getter function) can be improved IMHO, reflecting that it's about the name of the CR, which matches either node or machine name. I know, wasn't touched by this PR, but it would make the code better readable and understandable IMHO
- or, even better, get rid of this var at all, it's an ugly workaround IMHO. We can do a proper search for a matching CR in ` peerhealth/server.go`  using node and machine owners. Then we don't need this helper var and the last namespace var. That would also cover the case that someone mixes SNRs for nodes and machines (which wouldn't work today). WDYT?

slintes (changes requested):

controllers/selfnoderemediation_controller.go
mshitrit: I agree with 1 & 2.
Regarding  3 , I think it may be a bit tricky.

In order to do a proper search we need SNR to exist.
The way the code is written now, we first check for Machine/Node health and only if healthy we verify SNR doesn't exist before returning `Healthy` response. 
I assume this is done in order to address a use-case where SNR doesn't exist but the node/machine is still unhealthy :thinking:

slintes: > In order to do a proper search we need SNR to exist.

Why? 🤔  

> we first check for Machine/Node health

I don't see that? 👀 

The check is basically: 

old:
- if machine:
  - get node, on failure: _API error_ (🤔)
  - get machine name, if not found: _unhealthy_ (🤔)
- get CR by either node or machine name
- if not exists: _healthy_
- any other error: _API error_
- if exists: _unhealthy_

suggested:
- always get machine name from node as above, but do not fail if machine annotation not found (*)
- list CRs
- if error: _API error_
- search CR in list:
``` 
if (CR has Machine owner && CR has not NHC owner)
  if CR name matches machine name: _unhealthy_
else
  if CR name matches node name: _unhealthy_
```
- no match (or empty list): _healthy_

Does that make sense?

(*) we can save this step, if we add the machine name to the PeerHealth API alongside the node name, and let the client side find its machine on pod start... 🤔 That would solve the trouble of correctly handling issues when trying to get the machine name...

mshitrit: >we can save this step, if we add the machine name to the PeerHealth API alongside the node name

Sounds like a good idea.
I've looked into it a bit though.
it seems that ApiConnectivityCheckConfig (which would have the new MachineName later injected to the grpc)  is initialized before the manager is initialized.
IIUC that means that we will need another Runnable component in order to use the client to fetch the node (in order to  get the machine name from it).

Since in any case this is not initially part of the original PR I think I prefer to open a ticket and address this later on.

slintes: I don't think it needs a runnable, we just need get the own node once and check for the machine label. When you use a Reader instead of a full Client, it works before the manager starts. 

But ok if you want to address it later.

slintes (changes requested):
one nit with imports, otherwise lgtm

controllers/selfnoderemediation_controller.go
mshitrit: Thanks for the Reader tip ! I wasn't aware of that :smile: 
I've open another ticket for the refactoring: https://issues.redhat.com/browse/ECOPROJECT-1784

PR #166, opened 2023-11-29
Add e2e tests for OutOfServiceTaint remediation

k-keiichi-rh:
This PR is adding a new e2e tests for OutOfServiceTaint remediation strategy based on https://github.com/kubernetes/enhancements/pull/1116.
And https://github.com/openshift/release/pull/47048 depends on this pr.

### DIFF
--- a/e2e/self_node_remediation_test.go
+++ b/e2e/self_node_remediation_test.go
@@ -123,6 +123,25 @@ var _ = Describe("Self Node Remediation E2E", func() {
 					})
 				})
 
+				Context("OutOfService Remediation Strategy", func() {
+					var oldPodCreationTime time.Time
+
+					BeforeEach(func() {
+						remediationStrategy = v1alpha1.OutOfServiceTaintRemediationStrategy
+						oldPodCreationTime = findSnrPod(node).CreationTimestamp.Time
+					})
+
+					It("should delete pods", func() {
+						checkPodRecreated(node, oldPodCreationTime)
+						//Simulate NHC trying to delete SNR
+						deleteAndWait(snr)
+						snr = nil
+
+						checkNoExecuteTaintRemoved(node)
+						checkOutOfServiceTaintRemoved(node)
+					})
+				})
+
 			})
 		})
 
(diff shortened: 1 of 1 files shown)


e2e/self_node_remediation_test.go
```diff
@@ -342,6 +361,25 @@ var _ = Describe("Self Node Remediation E2E", func() {
 					})
 				})
 
+				Context("OutOfService Remediation Strategy", func() {
+					var oldPodCreationTime time.Time
+
+					BeforeEach(func() {
+						remediationStrategy = v1alpha1.OutOfServiceTaintRemediationStrategy
+						oldPodCreationTime = findSnrPod(node).CreationTimestamp.Time
```
mshitrit, line 369: **node** is undefined in this context, note that this Context is nested under `Describe("Control Plane Remediation", ...` 
Where node is defined in the context of  `Describe("Workers Remediation", func() { ... `

k-keiichi-rh:
@mshitrit Thank you for running the e2e tests in OCP4.12.
As you mentioned before, we can skip the e2e test with adding an env param like [this](https://github.com/openshift/release/blob/35bfa5d0e99d3ef7911913a48ebb71a9ef9c79fd/ci-operator/config/medik8s/self-node-remediation/medik8s-self-node-remediation-main__4.12.yaml#L65) in the  OCP versions that this feature isn't supported on.

Now we have another way to do same thing using the validation way(IsOutOfServiceTaintSupported) you implemented.

I think the latter way is preferred. What do you think about it?

mshitrit:
I think I'm leaning into env variable, but I don't have a strong preference so feel free to choose whichever implementation you prefer. 

This is what I had in mind:
- in the test code env variable give more flexibility (e.g we can also turn off this test for a specific supported version if we decide so)
- implementation seems simpler
- Once CI version that doesn't support the taint role out there is only an 'if' block in the code to remove instead of the version mechanism.

e2e/self_node_remediation_test.go
```diff
@@ -123,6 +124,25 @@ var _ = Describe("Self Node Remediation E2E", func() {
 					})
 				})
 
+				Context("OutOfService Remediation Strategy", func() {
```
mshitrit, line 127: I see that you've added the same test both in context of Worker remediation and of Control Plane remediation.
- IMO it's enough to leave it just in the context of worker remediation
- I see that for this use case (worker remediation) we don't check the env variable in order to skip the test when needed, is that intended ?

k-keiichi-rh: Thank you for the comment.
> IMO it's enough to leave it just in the context of worker remediation
I agree with you.
> I see that for this use case (worker remediation) we don't check the env variable in order to skip the test when needed, is that intended ?
That's not intended. We need to check both of the use cases. I will fix it.

mshitrit:
Hi @k-keiichi-rh , as expected 4.12 e2e fails because ci doesn't add the `SKIP_OOST_REMEDIATION_VERIFICATION`  variable.
please create a PR to [openshift/release](https://github.com/openshift/release) and update the variable [here](https://github.com/openshift/release/blob/master/ci-operator/config/medik8s/self-node-remediation/medik8s-self-node-remediation-main__4.12.yaml#L66) , once tests are clear, I think we could merge.

slintes:
https://github.com/openshift/release/pull/47048

PR #167, opened 2023-11-29
add 'replaces' to csv

mshitrit:
[ECOPROJECT-1767](https://issues.redhat.com//browse/ECOPROJECT-1767)

We need to update our CSVs with the following required metadata:

1. "replace" field:
while we use "skipRange" for defining from which versions we can update to the version of the CSV, we also need to add the "replace" field. It should point to the last released version, and ensures that the old version isn't deleted from the index, so customers can continue to use their tested and approved version.

https://docs.engineering.redhat.com/display/CFC/Best_Practices#Best_Practices-SkipRange

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,35 +291,8 @@ bundle-community-k8s: bundle-community ## Generate bundle manifests and metadata
 bundle-community: bundle ##Add Community Edition suffix to operator name
 	sed -r -i "s|displayName: Self Node Remediation Operator.*|displayName: Self Node Remediation Operator - Community Edition|;" ${BUNDLE_CSV}
 
-
-# Create the DEC_VERSION variable
-DEC_VERSION := 0.0.0
-
-# Check if PREV_VERSION is defined
-ifdef PREV_VERSION
-    # If PREV_VERSION is defined, use its value for DEC_VERSION
-    DEC_VERSION := $(PREV_VERSION)
-else
-	# Split the VERSION into major, minor, and patch parts
-	MAJOR := $(word 1, $(subst ., ,$(VERSION)))
-	MINOR := $(word 2, $(subst ., ,$(VERSION)))
-	PATCH := $(word 3, $(subst ., ,$(VERSION)))
-
-	# Determine which part to decrease
-	ifeq ($(PATCH),0)
-		# If PATCH is 0, decrease MINOR
-		MINOR := $(shell echo $$(($(MINOR) - 1)))
-	else
-		# If PATCH is greater than 0, decrease PATCH
-	   PATCH := $(shell echo $$(($(PATCH) - 1)))
-	endif
-
-	# Recreate the DEC_VERSION variable
-	DEC_VERSION := $(MAJOR).$(MINOR).$(PATCH)
-endif
-
 .PHONY: bundle-update
-bundle-update: ## Update containerImage, createdAt, skipRange, and icon fields in the bundle's CSV, then validate the bundle directory
+bundle-update: verify-previous-version ## Update containerImage, createdAt, skipRange, and icon fields in the bundle's CSV, then validate the bundle directory
 	# update container image in the metadata
 	sed -r -i "s|containerImage: .*|containerImage: ${IMG}|;" ${BUNDLE_CSV}
 	# set creation date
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -291,6 +291,18 @@ bundle-community-k8s: bundle-community ## Generate bundle manifests and metadata
 bundle-community: bundle ##Add Community Edition suffix to operator name
 	sed -r -i "s|displayName: Self Node Remediation Operator.*|displayName: Self Node Remediation Operator - Community Edition|;" ${BUNDLE_CSV}
 
+# Split the VERSION into major, minor, and patch parts
+MAJOR := $(word 1, $(subst ., ,$(VERSION)))
+MINOR := $(word 2, $(subst ., ,$(VERSION)))
+PATCH := $(word 3, $(subst ., ,$(VERSION)))
+
+# Subtract 1 from the MINOR part
```
slintes, line 299: I guess it depends:
For minor releases (patch version is 0) this is correct
For patch releases (patch version is >0) we need to decrease the patch version

bundle/manifests/self-node-remediation.clusterserviceversion.yaml
```diff
@@ -439,6 +439,7 @@ spec:
   provider:
     name: Medik8s
     url: https://www.medik8s.io/
+  replaces: self-node-remediation.v0.0.0
```
slintes, line 442: I'm wondering what happens when the referred version doesn't exist, this might cause trouble in CI? 🤔  We will see...

slintes (changes requested):

bundle/manifests/self-node-remediation.clusterserviceversion.yaml
mshitrit: looks ok I guess :man_shrugging:

Makefile
slintes: I think it's even more complicated, and automating it would require much more work: finding the actual previous release.
Reason: when releasing e.g. 0.6.0, the replace field needs to point to the last release. There is a good chance that the last one is 0.5.0 indeed, but it can also be any other patch version 0.5.x 🤷🏼‍♂️

mshitrit: >but it can also be any other patch version 0.5.x 🤷🏼‍♂️

indeed this is a good point.

How about passing PREV_VERSION as you suggested, but in case it's not passed default to current behavior ?

slintes: IMHO that would be an invite to forget setting it when needed...? 🤔

mshitrit: I admit that this thought crossed my mind as well, I think it's a trade off between comfort and security. 
IMO we already have strong gateways around security so I think it's ok to lean the other way here. 
However it's not something I feel strongly about so if you prefer it the other way I'm fine with changing it.

slintes: I'm not sure about it.
@clobrano @razo7 WDYT?

razo7: I am ok with the new automation for the old version. But it is a step/issue that we should remember.

clobrano: Can we rely on git for this? The repo already knows what's the previous tag without the need to infer it from the current `VERSION`

clobrano: I was a bit optimistic. Git doesn't really know what is the previous version, just the _last_ one, so some use cases will still remain uncovered.

My question is, if `PREV_VERSION` is not provided by us explicitly and we don't have an automatic way to get it always right, how can we instruct the CI what version to use?

mshitrit: In case it's not provided we deduct it from the current version (i.e either 0.X-1.0 or 0.X.Y-1), but since this cover only common case and not every case we can also provide PREV_VERSION explicitly.

clobrano: To be honest, I'd rather keep it mandatory. We still need to think if the replaced version will fall into the cases covered by the automation, so at this point just write it in the CLI.

Makefile
```diff
@@ -292,17 +292,26 @@ bundle-community: bundle ##Add Community Edition suffix to operator name
 	sed -r -i "s|displayName: Self Node Remediation Operator.*|displayName: Self Node Remediation Operator - Community Edition|;" ${BUNDLE_CSV}
 
 .PHONY: bundle-update
-bundle-update: ## Update containerImage, createdAt, skipRange, and icon fields in the bundle's CSV, then validate the bundle directory
+bundle-update: verify-previous-version ## Update containerImage, createdAt, skipRange, and icon fields in the bundle's CSV, then validate the bundle directory
```
clobrano, line 295: you might want to update the comment with the new `replaces` field, or just change it to a more generic "update CSV's fields and validate the bundle directory"

clobrano (changes requested):

PR #169, opened 2023-12-12
rapidast scan execution as kuttl testsuite

frajamomo:


frajamomo:
Deferred until final version of the equivalent code is reviewed and integrated for fence-agents-remediation operator.

PR #172, opened 2023-12-25
[WIP] Add machine name to grpc 

mshitrit:
ECOPROJECT-1784
As a Followup for [this](https://github.com/medik8s/self-node-remediation/pull/164#discussion_r1415029123) PR Add machine name to grpc request in order to remove workaround of using a global variable that checks whether snr is created for a machine.


clobrano:
/lgtm 
Givng others a chance to review as well, feel free to unhold
/hold

slintes (changes requested):

pkg/peerhealth/server.go
```diff
@@ -140,39 +138,12 @@ func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthR
 	}
 
 	if isMachine {
-		return toResponse(s.isHealthyMachine(ctx, nodeName, namespace))
+		return toResponse(s.isHealthyBySnr(ctx, request.MachineName, namespace))
```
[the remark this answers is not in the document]
mshitrit: I'll use an example to make sure I follow.
Let's say that I'm checking node named `node1` with the relevant machine named `machine1`.
- when querying all the namespaces I find 2 snrs (to keep it simple let's assume that they are on the same ns).
- First one named `Foo` with owner NHC (let's say that it was wrongly created by the user )
- Second one named `machine1` with kind `Machine` (without NHC owner), which was created by MHC.

IIUC following the suggest logic in that particular use case `isHealthyBySnr` would return `Healthy` response which seems wrong :thinking: 

Wouldn't it be safer to check each found SNR and returns a `Healthy` response only if all of them are ?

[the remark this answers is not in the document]
slintes: > IIUC following the suggest logic in that particular use case isHealthyBySnr would return Healthy

No, that's not what I wanted to suggest.
To clarify, with "check their owners: if one them is of kind NHC" I meant check owners of **_ALL CRs_**, and if one of **_THE OWNERS_** of a CR  is..." 

> check each found SNR and returns a Healthy response only if all of them are

sure 🙂

pkg/peerhealth/server.go
```diff
@@ -121,56 +119,24 @@ func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthR
...
-	namespacedMachine, exists := ann[machineAnnotation]
-
-	if !exists {
-		s.log.Info("node doesn't have machine annotation")
-		return selfNodeRemediationApis.Unhealthy //todo is this the correct response?
+	//fetch all snrs from all ns
+	snrs := &v1alpha1.SelfNodeRemediationList{}
+	if err := s.snr.List(ctx, snrs); err != nil {
+		s.log.Error(err, "failed to fetch snrs")
+		return nil, err
```
slintes, line 126: we need to return `selfNodeRemediationApis.ApiError` here (see current code in `isHealthyBySnr`)

pkg/peerhealth/server.go
```diff
@@ -121,56 +119,24 @@ func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthR
...
 	}
-	_, machineName, err := cache.SplitMetaNamespaceKey(namespacedMachine)
 
-	if err != nil {
-		s.log.Error(err, "failed to parse machine annotation on the node")
-		return selfNodeRemediationApis.Unhealthy //todo is this the correct response?
+	//return healthy only if all of snrs are considered healthy for that node
+	for _, snr := range snrs.Items {
+		if controllers.IsOwnedByNHC(&snr) {
```
slintes, line 131: this rings my alarm about using pointers in loops...!

pkg/peerhealth/server.go
slintes: and we need use a new context with `apiServerTimeout`

pkg/peerhealth/server.go
```diff
@@ -121,56 +119,24 @@ func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthR
...
+			if healthCode := s.isHealthyBySnr(ctx, request.NodeName, snr.Namespace); healthCode != selfNodeRemediationApis.Healthy {
```
slintes, line 132: I don't think we need `isHealhyBySnr` anymore:
- it gets the SNR, which isn't needed, we already have it here. We just need to compare the SNR name here with either nodeName or machineName. If it matches, return `Unhealthy`
- the `ApiError` case should be handled above, see previous comment
- the `Healthy` case is correctly handled below in case we didn't find a matching SNR

slintes (changes requested):

pkg/peerhealth/server.go
mshitrit: I think you are referring to the common error when we use it in a closure.
something like:
```
for _, snr := range snrs.Items {
   func() {
      if controllers.IsOwnedByNHC(&snr) {
      ...
      }()
      ...
 }
```		

In this case we are in the clear.

pkg/peerhealth/server.go
mshitrit: Thanks, great feedback :+1:

slintes (approved):

PR #176, opened 2024-01-17
fix flaky UT

mshitrit:
Fixing flaky UT, and removing VA logic since SNR does not delete VA's anymore

### DIFF
--- a/controllers/tests/controller/selfnoderemediation_controller_test.go
+++ b/controllers/tests/controller/selfnoderemediation_controller_test.go
@@ -10,7 +10,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	v1 "k8s.io/api/core/v1"
-	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -35,7 +34,6 @@ const (
 var _ = Describe("SNR Controller", func() {
 	var snr *v1alpha1.SelfNodeRemediation
 	var remediationStrategy v1alpha1.RemediationStrategyType
-	var vaName = "some-va"
 	var nodeRebootCapable = "true"
 	var isAdditionalSetupNeeded = false
 
@@ -49,9 +47,6 @@ var _ = Describe("SNR Controller", func() {
 
 	JustBeforeEach(func() {
 		if isAdditionalSetupNeeded {
-			createVolumeAttachment(vaName)
-			verifyVaNotDeleted(vaName)
-
 			createSelfNodeRemediationPod()
 			verifySelfNodeRemediationPodExist()
 		}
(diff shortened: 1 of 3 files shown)


pkg/peerhealth/client_server_test.go
```diff
@@ -111,11 +111,12 @@ var _ = Describe("Checking health using grpc client and server", func() {
...
-				NodeName: nodeName,
-			})
-			Expect(err).ToNot(HaveOccurred())
-			Expect(api.HealthCheckResponseCode(resp.Status)).To(Equal(api.Unhealthy))
+			Eventually(func() bool {
+				resp, err := phClient.IsHealthy(ctx, &HealthRequest{
+					NodeName: nodeName,
+				})
+				return err == nil && api.HealthCheckResponseCode(resp.Status) == api.Unhealthy
+			}, time.Millisecond*250, time.Second*5).Should(BeTrue())
```
razo7, line 119: IIUC then you messed up with the order of parameters here. 
The first parameter would be timeout, and the second would be a polling interval https://pkg.go.dev/github.com/onsi/gomega#Eventually.

Using these variables it seems like `Eventually` ran one and wouldn't be retriggered, since `timeout <  polling interval` ( e.g., 250ms < 5s) . But if these variables were good for fixing the flakiness and it is indeed run once (maybe add some debugging prints inside to verify my assumption), then you don't need this new _Eventually_ and you can revert to what we have before.

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -493,39 +462,10 @@ func verifyTypeConditions(nodeName string, expectedProcessingConditionStatus, ex
...
-	ConsistentlyWithOffset(1, func() bool {
-		va := &storagev1.VolumeAttachment{}
-		err := k8sClient.Get(context.Background(), vaKey, va)
-		return apierrors.IsNotFound(err)
-
-	}, 5*time.Second, 250*time.Millisecond).Should(BeFalse())
-}
-
 func verifyLastErrorKeepsApiError() {
 	By("Verify that LastError in SNR status has been kept kube-api error for VA")
```
razo7, line 466: Mentioning VA could be removed :)

razo7:
Can you separate fixing flaky UT, and removing VA logic into two commits? It would be easier for future us to know if flaky UT is related to VA logic. Do you happen to understand why this unit test is flaky?

pkg/peerhealth/client_server_test.go
mshitrit: Thanks, good catch

mshitrit:
>Can you separate fixing flaky UT, and removing VA logic into two commits?

Sure done.

>Do you happen to understand why this unit test is flaky?

Not sure if you refer to a specific failure regarding the VA or in general.
In any case in general most of the flakiness is due to sync issues (i.e mostly fetching a resource before it has reached the desired state).
Regarding the VAs it seems like a waste of time dealing with syncing with them when they are not needed in any case anymore.

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -933,13 +866,15 @@ func clearEvents() {
...
-	ExpectWithOffset(1, isEventMatch).To(BeTrue())
+	EventuallyWithOffset(1, func() bool {
+		return isEventOccurred(eventType, reason, message)
+	}, 5*time.Second, 250*time.Millisecond).Should(BeTrue())
 }
 
 func verifyNoEvent(eventType, reason, message string) {
-	isEventMatch := isEventOccurred(eventType, reason, message)
-	ExpectWithOffset(1, isEventMatch).To(BeFalse())
+	EventuallyWithOffset(1, func() bool {
```
slintes, line 875: this should be Consistently I guess?

mshitrit: Nice catch, thanks.

mshitrit: Here: https://github.com/medik8s/self-node-remediation/pull/178

PR #182, opened 2024-01-30
Remove combined NHC / SNR index image 

mshitrit:
Remove combined NHC / SNR index image because it couples NHC and SNR go versions


clobrano:
/lgtm
Giving a chance to get more reviews, feel free to unhold
/hold

PR #184, opened 2024-01-30
Updating "Automatic" log message

mshitrit:
[ECOPROJECT-1862](https://issues.redhat.com//browse/ECOPROJECT-1862) 
- Considering a use case where the finalizer is placed by the agent and than we loose connectivity after reboot, missing this log message.
- Also fixing a flaky UT in config

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -913,10 +913,7 @@ func (r *SelfNodeRemediationReconciler) getRuntimeStrategy(snr *v1alpha1.SelfNod
 		remediationStrategy = v1alpha1.OutOfServiceTaintRemediationStrategy
 	}
 
-	//used as an indication not to spam the log
-	if isFinalizerAlreadyAdded := controllerutil.ContainsFinalizer(snr, SNRFinalizer); !isFinalizerAlreadyAdded {
-		r.logger.Info(fmt.Sprintf("Automatically selected %s Remediation strategy", remediationStrategy))
-	}
+	r.logger.Info(fmt.Sprintf("Remediating with %s Remediation strategy (auto-selected)", remediationStrategy))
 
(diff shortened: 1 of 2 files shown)


slintes (approved):
Idea: we could put the runtime strategy into an annotation, so it's visible on the CR, and we don't need to spam the logs with it in each reconcile? 

Can also be done in a follow up...

/hold

slintes:
> > Idea: we could put the runtime strategy into an annotation, so it's visible on the CR, and we don't need to spam the logs with it in each reconcile?
> 
> Maybe, TBH I am not even so sure that we should even print this log to begin with. I think it comes down to how visible we want to be with the inner working of "Automatic".
> 
> I'm not sure what's the proper balance from a user perspective between: `Automatic: I choose that because I trust you to do it best without involving me in the details` to `Automatic: here is how it works we selected X because Y`

Maybe it's fair to discuss how much info to reveal to the user. But we also need it ourself for debugging, not?

> 
> I'm also a bit worried from followup bugs (when we chance the Automatic selection/ or when run on different ocp version) such as the runtime annotation used to be X and now it's Y

Actually I see this as an pro argument for the annotation. In case SNR or the cluster updates during remediation, it would be bad to switch from one strategy to another one, not? The cleanup would not match anymore. With the annotation we can ensure that we use the same strategy for the whole remediation process.

mshitrit:
>Maybe it's fair to discuss how much info to reveal to the user. But we also need it ourself for debugging, not?

Sure SGTM.
In any case I consider logs as something less visible than annotations/events and targeted for more technical users.

> it would be bad to switch from one strategy to another one, not? 

I'm not sure.
In case the remediation still exist in this use case there is a very good chance that we weren't able to fix it for awhile, might not be such a bad idea to try with a different strategy - it might even be the purpose of a change in the Automatic strategy.

PR #186, opened 2024-02-01
[WIP] increase SafeTimeToAssumeNodeRebootedSeconds 

mshitrit:
increase SafeTimeToAssumeNodeRebootedSeconds as in some clusters it's value is lower than minimum calculated value


mshitrit:
Suggested a better approach to solve this:  https://issues.redhat.com/browse/ECOPROJECT-1875

PR #190, opened 2024-02-18
Events refactoring

mshitrit:
[ECOPROJECT-1836](https://issues.redhat.com//browse/ECOPROJECT-1836)
- not omitting remediation started event for not found nodes
- using node not found event from common
- adding indication of whether the message was omitted by the agent or manager

### DIFF
--- a/controllers/tests/controller/selfnoderemediation_controller_test.go
+++ b/controllers/tests/controller/selfnoderemediation_controller_test.go
@@ -167,7 +167,7 @@ var _ = Describe("SNR Controller", func() {
 			It("Remediation flow", func() {
 				node := verifyNodeIsUnschedulable()
 
-				verifyEvent("Normal", "RemediationStarted", "Remediation started")
+				verifyEvent("Normal", "RemediationStarted", "Remediation started by SNR manager")
 
 				verifyEvent("Normal", "MarkUnschedulable", "Remediation process - unhealthy node marked as unschedulable")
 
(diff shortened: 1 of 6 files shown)


controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -167,7 +167,7 @@ var _ = Describe("SNR Controller", func() {
 			It("Remediation flow", func() {
 				node := verifyNodeIsUnschedulable()
 
-				verifyEvent("Normal", "RemediationStarted", "Remediation started")
+				verifyEvent("Normal", "RemediationStarted", "Remediation started by SNR manager")
```
razo7, line 170: How about using _EventTypeNormal_ and _EventTypeNormal_  from https://github.com/kubernetes/api/blob/v0.29.2/core/v1/types.go#L6354-L6359 instead of typing the string? Here and in other occurrences in this file.

razo7: Instead of using the string _RemediationStarted_ we can also use the [common remediation event reasons](https://github.com/medik8s/common/blob/main/pkg/events/events.go#L15-L18).

mshitrit: I don't think it's very significant in test code.
One can even argue against it in test code, because it would be safer to validate vs a "decoupled " hardcoded value (personally not an issue as well IMO).

razo7: What about the event type?

mshitrit: TBH my original response for the event type question  :sweat_smile: (I only saw the followup after I posted)
I was just luck that my answer fitted nicely the second question

razo7:
/lgtm
Unless you would like to make a minor change by using the eventType from Corev1

PR #191, opened 2024-02-20
Automate RH Community bundle creation

clobrano:
- Install envsubst locally if necessary
- Add target to build RH community bundle with OCP Annotations
- Automate RH community bundle release with Github Action

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ KUSTOMIZE_VERSION = v5.2.1
 # versions at https://github.com/slintes/sort-imports/tags
 SORT_IMPORTS_VERSION = v0.2.1
 
+# version at https://github.com/a8m/envsubst/releases
+ENVSUBST_VERSION = v1.4.2
+
 OPERATOR_NAME ?= self-node-remediation
 
 # VERSION defines the project version for the bundle.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -246,6 +252,12 @@ ifeq (,$(wildcard $(CONTROLLER_GEN)))
 	}
 endif
 
+ENVSUBST_BIN_FOLDER = $(shell pwd)/bin/envsubst
+ENVSUBST = $(ENVSUBST_BIN_FOLDER)/$(ENVSUBST_VERSION)/envsubst
+.PHONY: envsubst
+envsubst: ## Download envsubst locally if necessary.
+	$(call go-install-tool,$(ENVSUBST),github.com/a8m/envsubst/cmd/envsubst@${ENVSUBST_VERSION}) ;\
```
slintes, line 259: Do we need this just for OCP_VERSION?
Why not using the actual value directly in `bundle-community-rh` 🤔

clobrano: It is not specifically for OCP_VERSION.
The CI was complaining that `envsubst` does not exist in the container and the resulting `bundle` directory was almost empty.

I think* this came up now because this new CI _removes_ the bundle first and then recreates it (instead than just update what's already committed) so we notice that some files cannot be recreated in CI.

* I didn't actually verify this, let me try first if I'm right

clobrano: A bit more context. When running this CI, and specifically the `bundle` recipe, locally (so with [act](https://github.com/nektos/act)) I got this error

```
| /bin/bash: line 1: envsubst: command not found
```

and the corresponding step fails, so the bundle lacks of many files (it basically only have csv and annotations.yaml).
This worflow makes it very noticeable because it first delete the whole bundle folder, so it easy to spot the missing files, which is a step that the current workfiles we have don't do, and I supposed that those files simply stays the same.

However, I can't see the same error in any of the jobs already run in the past :shrug:

slintes: > It is not for OCP_VERSION.

Oh, correct, sorry

> locally (so with act) I got this error

Fair enough.
Also, I somehow oversaw that envsubst was used before already, for setting ${IMG} in the env of a container in the CSV, sorry again 🙂 

/lgtm

clobrano: No worries :)

PR #192, opened 2024-02-20
Remove catalog build and push to registry from Makefile

clobrano:


clobrano:
Change plans. I'll update the target to create a file-based catalog instead

PR #201, opened 2024-05-20
update base image to ubi9

mshitrit:
#### Why we need this PR
Upgrading ubi8 to ubi9

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM quay.io/centos/centos:stream8 AS builder
+FROM quay.io/centos/centos:stream9 AS builder
 RUN yum install git golang -y && yum clean all
 
 WORKDIR /workspace
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Build the manager binary
 FROM quay.io/centos/centos:stream9 AS builder
-RUN yum install git golang jq -y && yum clean all
+RUN yum install git jq -y && yum clean all
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ COPY install/ install/
 # Build
 RUN ./hack/build.sh
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM registry.access.redhat.com/ubi9/ubi:latest
 
 WORKDIR /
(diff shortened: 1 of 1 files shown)


razo7:
seems like a chicken and egg problem, but don't forget to update CI afterward for using the UBI9 base image (in https://github.com/openshift/release/blob/master/ci-operator/config/medik8s/self-node-remediation/medik8s-self-node-remediation-main__4.12.yaml#L5 and in similar places)

Dockerfile
```diff
@@ -30,7 +30,7 @@ COPY install/ install/
 # Build
 RUN ./hack/build.sh
 
-FROM registry.access.redhat.com/ubi8/ubi:latest
+FROM registry.access.redhat.com/ubi9/ubi:latest
```
slintes, line 33: not an issue as long as we build a statically linked binary AFAIK, but I think it's better to update the builder image to centos:stream9 as well

slintes:
This is using the wrong Go version 😞 
From the image job:
```
[...]
2024-05-22T08:16:41.825834595Z   Installing       : golang-1.22.2-1.el9.x86_64                         106/107 
[...]
2024-05-22T08:16:43.170425063Z [1/2] STEP 7/18: ENV GOTOOLCHAIN=auto
2024-05-22T08:16:43.215108442Z [1/2] STEP 8/18: RUN export GO_VERSION=$(grep -E "go [[:digit:]]\.[[:digit:]][[:digit:]]" go.mod | awk '{print $2}') &&     go get go@${GO_VERSION} &&     go version
2024-05-22T08:16:44.051438329Z go: upgraded go 1.21 => 1.21.10
2024-05-22T08:16:44.051438329Z go: added toolchain go1.22.2
2024-05-22T08:16:44.063705553Z go version go1.22.2 (Red Hat 1.22.2-1.el9) linux/amd64
```
We've seen this before, I'd suggest to install Go from `go.dev` like we do in NHC already: https://github.com/medik8s/node-healthcheck-operator/blob/main/Dockerfile#L12-L26

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build the manager binary
-FROM quay.io/centos/centos:stream8 AS builder
-RUN yum install git golang -y && yum clean all
+FROM quay.io/centos/centos:stream9 AS builder
+RUN yum install git golang jq -y && yum clean all
```
slintes, line 3: you can remove go here :)

slintes (changes requested):

PR #204, opened 2024-05-29
Fix flaky UT

mshitrit:
#### Why we need this PR
improve pipeline stability for shorter dev cycles

#### Changes made
There was a race condition in the test where in some case the daemonset was created twice, it was fixed to make sure it will only happen once.

### DIFF
--- a/controllers/tests/config/selfnoderemediationconfig_controller_test.go
+++ b/controllers/tests/config/selfnoderemediationconfig_controller_test.go
@@ -11,6 +11,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -24,7 +25,12 @@ var _ = Describe("SNR Config Test", func() {
 	var config *selfnoderemediationv1alpha1.SelfNodeRemediationConfig
 	var ds *appsv1.DaemonSet
 	dummySelfNodeRemediationImage := "self-node-remediation-image"
-
+	var createDsBeforeConfig bool
+	key := types.NamespacedName{
+		Namespace: shared.Namespace,
+		Name:      dsName,
+	}
+	var isSkipCleanup bool
 	BeforeEach(func() {
 		ds = &appsv1.DaemonSet{}
 		_ = os.Setenv("SELF_NODE_REMEDIATION_IMAGE", dummySelfNodeRemediationImage)
@@ -36,21 +42,49 @@ var _ = Describe("SNR Config Test", func() {
 		config.Name = selfnoderemediationv1alpha1.ConfigCRName
 		config.Namespace = shared.Namespace
 		config.Spec.HostPort = 30111
+		createDsBeforeConfig = false
 
 	})
 
-	Context("DS installation", func() {
-		key := types.NamespacedName{
-			Namespace: shared.Namespace,
-			Name:      dsName,
+	AfterEach(func() {
+		if isSkipCleanup {
+			isSkipCleanup = false
+			return
 		}
 
+		tmpConfig := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
+		tmpDs := &appsv1.DaemonSet{}
+		//verify ds and config exist
+		Eventually(func(g Gomega) bool {
+
+			g.Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(config), tmpConfig)).To(Succeed())
+			g.Expect(k8sClient.Get(context.Background(), key, tmpDs)).To(Succeed())
+			return true
+		}, 10*time.Second, 250*time.Millisecond).Should(BeTrue())
+
+		Expect(k8sClient.Delete(context.TODO(), tmpConfig)).To(Succeed())
+		Expect(k8sClient.Delete(context.TODO(), tmpDs)).To(Succeed())
+
+		//verify delete is done
+		Eventually(func(g Gomega) bool {
+			err := k8sClient.Get(context.Background(), key, &appsv1.DaemonSet{})
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			err = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(config), &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{})
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			return true
+		}, 10*time.Second, 250*time.Millisecond).Should(BeTrue())
+	})
+
+	Context("DS installation", func() {
+
 		JustBeforeEach(func() {
+			if createDsBeforeConfig {
+				Eventually(func() error {
+					return k8sClient.Create(context.Background(), ds)
+				}, 2*time.Second, 250*time.Millisecond).Should(Succeed())
+			}
+			//Creates config which in turn will create the DS
 			Expect(k8sClient.Create(context.Background(), config)).To(Succeed())
-			DeferCleanup(func() {
-				Expect(k8sClient.Delete(context.Background(), config)).To(Succeed())
-				k8sClient.Delete(context.Background(), ds)
-			})
 
 		})
 
@@ -70,10 +104,13 @@ var _ = Describe("SNR Config Test", func() {
 		})
 
 		It("Cert Secret should be created", func() {
-			Eventually(func() error {
+			Eventually(func(g Gomega) bool {
 				_, _, _, err := certReader.GetCerts()
-				return err
-			}, 15*time.Second, 250*time.Millisecond).ShouldNot(HaveOccurred())
+				g.Expect(err).ShouldNot(HaveOccurred())
+				g.Expect(k8sClient.Get(context.Background(), key, ds)).To(Succeed())
+				return true
+			}, 15*time.Second, 250*time.Millisecond).Should(BeTrue())
+
 		})
 
 		It("Daemonset should be created", func() {
@@ -169,16 +206,15 @@ var _ = Describe("SNR Config Test", func() {
 		Context("DS Recreation on Operator Update", func() {
 			var timeToWaitForDsUpdate = 6 * time.Second
 			var oldDsVersion, currentDsVersion = "0", "1"
-
+			BeforeEach(func() {
+				createDsBeforeConfig = true
+			})
 			JustBeforeEach(func() {
-				Eventually(func() error {
-					return k8sClient.Create(context.Background(), ds)
-				}, 2*time.Second, 250*time.Millisecond).Should(Succeed())
-
-				Eventually(func() error {
-					return k8sClient.Get(context.Background(), key, ds)
-				}, 2*time.Second, 250*time.Millisecond).Should(BeNil())
-
+				//Wait for Ds to be created
+				Eventually(func(g Gomega) bool {
+					g.Expect(k8sClient.Get(context.Background(), key, &appsv1.DaemonSet{})).To(Succeed())
+					return true
+				}, 10*time.Second, 250*time.Millisecond).Should(BeTrue())
 			})
 			When("ds version has not changed", func() {
 				BeforeEach(func() {
@@ -189,7 +225,6 @@ var _ = Describe("SNR Config Test", func() {
 					time.Sleep(timeToWaitForDsUpdate)
 					Expect(k8sClient.Get(context.Background(), key, ds)).To(BeNil())
 					Expect(ds.Annotations["original-ds"]).To(Equal("true"))
-
 				})
 			})
 
@@ -219,6 +254,9 @@ var _ = Describe("SNR Config Test", func() {
 		config.APIVersion = "self-node-remediation.medik8s.io/v1alpha1"
 		config.Name = "config-defaults"
 		config.Namespace = shared.Namespace
+		BeforeEach(func() {
+			isSkipCleanup = true
+		})
 
 		It("Config CR should be created with default values", func() {
 			Expect(k8sClient).To(Not(BeNil()))
--- a/controllers/tests/config/selfnoderemediationconfig_controller_test.go
+++ b/controllers/tests/config/selfnoderemediationconfig_controller_test.go
@@ -246,22 +246,21 @@ var _ = Describe("SNR Config Test", func() {
 	})
 
 	Context("SNRC defaults", func() {
-		config := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
-		config.Kind = "SelfNodeRemediationConfig"
-		config.APIVersion = "self-node-remediation.medik8s.io/v1alpha1"
-		config.Name = "config-defaults"
-		config.Namespace = shared.Namespace
+		defaultConfig := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
 		BeforeEach(func() {
-			//Config and DS aren't created so no need to clean them
-			isSkipCleanup = true
+			defaultConfig.Kind = "SelfNodeRemediationConfig"
+			defaultConfig.APIVersion = "self-node-remediation.medik8s.io/v1alpha1"
+			//Setting the same name and namespace so it'll be picked up by the AfterEach cleanup
+			defaultConfig.Name = selfnoderemediationv1alpha1.ConfigCRName
+			defaultConfig.Namespace = shared.Namespace
 		})
 
 		It("Config CR should be created with default values", func() {
 			Expect(k8sClient).To(Not(BeNil()))
-			Expect(k8sClient.Create(context.Background(), config)).To(Succeed())
+			Expect(k8sClient.Create(context.Background(), defaultConfig)).To(Succeed())
 
 			createdConfig := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
-			configKey := client.ObjectKeyFromObject(config)
+			configKey := client.ObjectKeyFromObject(defaultConfig)
 
 			Eventually(func() error {
 				return k8sClient.Get(context.Background(), configKey, createdConfig)
--- a/controllers/tests/config/selfnoderemediationconfig_controller_test.go
+++ b/controllers/tests/config/selfnoderemediationconfig_controller_test.go
@@ -202,12 +202,6 @@ var _ = Describe("SNR Config Test", func() {
 					return k8sClient.Get(context.Background(), dsKey, ds)
 				}, 2*time.Second, 250*time.Millisecond).Should(BeNil())
 			}
-			JustBeforeEach(func() {
-				//Wait for Ds to be created
-				Eventually(func(g Gomega) {
-					g.Expect(k8sClient.Get(context.Background(), dsKey, &appsv1.DaemonSet{})).To(Succeed())
-				}, 10*time.Second, 250*time.Millisecond).Should(Succeed())
-			})
 			When("ds version has not changed", func() {
 				BeforeEach(func() {
 					ds = generateDs(dsName, shared.Namespace, currentDsVersion)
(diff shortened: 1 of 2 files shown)


clobrano:
Giving a chance to get more reviews
/hold

slintes (changes requested):
A better description with the actual error would have been nice.
However, I found 2 failed tests in CI history, so I assume this is about the failed creation of the ds in `JustBeforeEach` of the `DS Recreation on Operator Update` tests, because the ds exists?

Tbh, your "fix" looks way to complicated to me:
- `createDsBeforeConfig` should not be needed. It should be sufficient to define order of code execution by level of nesting of `BeforeEach` and `JustBeforeEach`
- I don't see the value of `isSkipCleanup`. And directly related, what's the value of `AfterEach` checking the existence of the resources, before deleting them?
- Did you try to just add the `verify delete is done` part to the old `DeferCleanup()`? That should ensure that no ds exists when the operator update tests run...

mshitrit:
> * `createDsBeforeConfig` should not be needed. It should be sufficient to define order of code execution by level of nesting of `BeforeEach` and `JustBeforeEach`

That's a nice idea, but the inner `JustBeforeEach`  that belongs to the particular test in which I want to create the DS before the config runs after the outer `JustBeforeEach` which is shared with multiple tests so I'm not sure how to do it.


> * I don't see the value of `isSkipCleanup`. 

In the cleanup process we verify that the delete was successful, we can't do it if no resources were created.


>And directly related, what's the value of `AfterEach` checking the existence of the resources, before deleting them?

At that stage I know the resources  were set up to be created but since the creation is async we need to verify they were created, otherwise we run into the risk of resources being created after the delete was executed. 


> * Did you try to just add the `verify delete is done` part to the old `DeferCleanup()`? That should ensure that no ds exists when the operator update tests run...

It's possible to run the AfterEach code in DeferCleanup, in this case I preferred using an "AfterEach" block because I think it has slightly better readability,  DeferCleanup creates another nesting and adds a bit of complication to the BeforeEach so I prefer to use it with a relative short cleanup code. 
It's definitely  something I consider very minor so DeferCleanup is fine by me as well.

slintes:
>> createDsBeforeConfig should not be needed. It should be sufficient to define order of code execution by level of nesting of BeforeEach and JustBeforeEach

> That's a nice idea, but the inner JustBeforeEach that belongs to the particular test in which I want to create the DS before the config runs after the outer JustBeforeEach which is shared with multiple tests so I'm not sure how to do it.

ooohhhh.... that's a very fair point, I remembered it the wrong around. However, that made me wonder how the test ever succeeds! And revisiting the code, the problem isn't that we run into a race condition which fails the test sometimes, the problem is that it's coincidence that the test succeeds most times! Let me explain:

- the outer JustBeforeEach creates the config
- reconciling the config triggers ds creation
- this competes with the ds creation of the inner JustBeforeEach

This can easily be demonstrated: just add a one second sleep at the end of the outer JustBeforeEach, that gives the config reconcile time to always create the ds before the inner JustBeforeEach creates the ds. The test always fails now.

Solution: make the code which is in the inner JustBeforeEach a function, and call it in the inner BeforeEach of both tests. That way the "inner" ds is always created before the one created by the "outer" config reconcile :)

> In the cleanup process we verify that the delete was successful, we can't do it if no resources were created.

I don't see a need for this. Cleanup should ensure that resources don't exist after cleanup, no matter if they exist before. So you can just ignore DoesNotExist errors (don't remember the exact name). Let's keep it simple :)

> At that stage I know the resources were set up to be created but since the creation is async we need to verify they were created

fair point, such verification should happen in test setup or execution though, not cleanup IMHO.

> It's possible to run the AfterEach code in DeferCleanup, in this case I preferred using an "AfterEach" block because I think it has slightly better readability

I don't have a strong preference here as well. For more complex cleanup AfterEach() is probably easier to read, for just deleting a resource it's nice to have the cleanup code right after the creation code IMHO. However, my point here was that you added something valuable. And that could have been added to the existing DeferCleanUp(), in contrast to the bigger refactoring :)

mshitrit:
> the problem isn't that we run into a race condition which fails the test sometimes, the problem is that it's coincidence that the test succeeds most times!

Not splitting hair about the terminology, since I assume we mean the same thing here.
We can probably agree that there is an issue with timing.

> just add a one second sleep

I'm not a fan of adding a fixed sleep to solve timing issues (not to say I never do it, but I try to avoid doing so).
I think it's harder to maintain and very dependent on system performance (which may vary) and it'll also cause a delay in all the rest of the tests.

>Solution: make the code which is in the inner JustBeforeEach a function, and call it in the inner BeforeEach of both tests. That way the "inner" ds is always created before the one created by the "outer" config reconcile :)

I'm having a hard time to follow, IIUC the inner JustBeforeEach [code](https://github.com/medik8s/self-node-remediation/pull/204/files#diff-91365162a1d1121a55c65ac4980e04d17aa1dead56aaa10a688b2032de96a831R216-R219) Just verifies the ds is created, so it can't run before (i.e as part of a BeforeEach block) the outer JustBeforeEach [code](https://github.com/medik8s/self-node-remediation/pull/204/files#diff-91365162a1d1121a55c65ac4980e04d17aa1dead56aaa10a688b2032de96a831R81-R87) creates the config/ds.
Maybe you can give an example with a code snippet ?

>I don't see a need for this. Cleanup should ensure that resources don't exist after cleanup, no matter if they exist before. So you can just ignore DoesNotExist errors (don't remember the exact name). Let's keep it simple :)

I think there is value for checking because we verify that the test is going as expected which means we'll catch errors earlier and improve maintainability, I do agree though that there is a trade-off  with readability.

Another reason is: that in case we are agnostic in the cleanup phase to whether resources were created or not, we risk the resources being created after we've already assumed we've deleted them in the cleanup.  

>And that could have been added to the existing DeferCleanUp()

The exiting DeferCleanUp() was located in a more nested JustBeforeEach block so it excluded some tests which are now part of the cleanup.

slintes:
> > the problem isn't that we run into a race condition which fails the test sometimes, the problem is that it's coincidence that the test succeeds most times!
> 
> Not splitting hair about the terminology, since I assume we mean the same thing here. We can probably agree that there is an issue with timing.

sure, my point just is that the problem is more that it succeeds most times, not that it fails sometimes. The failure exposes a serious issue in the test setup
 
> 
> > just add a one second sleep
> 
> I'm not a fan of adding a fixed sleep to solve timing issues (not to say I never do it, but I try to avoid doing so). I think it's harder to maintain and very dependent on system performance (which may vary) and it'll also cause a delay in all the rest of the tests.

as written, it can **demonstrate** the actual issue. I didn't want to propose adding a timeout permanently (it makes the test always fail). The proposal was to add the sleep on your machine so you can reproduce the issue always, and not just sometimes 

> 
> > Solution: make the code which is in the inner JustBeforeEach a function, and call it in the inner BeforeEach of both tests. That way the "inner" ds is always created before the one created by the "outer" config reconcile :)
> 
> I'm having a hard time to follow, IIUC the inner JustBeforeEach [code](https://github.com/medik8s/self-node-remediation/pull/204/files#diff-91365162a1d1121a55c65ac4980e04d17aa1dead56aaa10a688b2032de96a831R216-R219) Just verifies the ds is created, so it can't run before (i.e as part of a BeforeEach block) the outer JustBeforeEach [code](https://github.com/medik8s/self-node-remediation/pull/204/files#diff-91365162a1d1121a55c65ac4980e04d17aa1dead56aaa10a688b2032de96a831R81-R87) creates the config/ds. Maybe you can give an example with a code snippet ?

I was explaining the issue we have in the current test code, not in the PR code. When you look at the current code, my explanation of the issue should make sense, does it?
 
> > I don't see a need for this. Cleanup should ensure that resources don't exist after cleanup, no matter if they exist before. So you can just ignore DoesNotExist errors (don't remember the exact name). Let's keep it simple :)
> 
> I think there is value for checking because we verify that the test is going as expected which means we'll catch errors earlier and improve maintainability, I do agree though that there is a trade-off with readability.

sure, but verifying "that the test is going as expected" is very clearly a task for test setup (`(Just)BeforeEach()`) or execution (`It()`), not cleanup (`(Just)AfterEach()`)!

> Another reason is: that in case we are agnostic in the cleanup phase to whether resources were created or not, we risk the resources being created after we've already assumed we've deleted them in the cleanup.

On one hand: as stated above, the test setup or test execution should verify that resources we expect to be created are created already.
On the other hand you have a point indeed in this specific case: if we test the config only and don't care about ds, we might miss cleanup of the ds in case the test finishes faster than the ds is created. However, IMHO a `Consistently` with 1s for the ds deletion would be nicer than that boolean flag + `Eventually` for the getting the ds. Then we don't need to care about setting that flag for each test or not.

> > And that could have been added to the existing DeferCleanUp()
> 
> The exiting DeferCleanUp() was located in a more nested JustBeforeEach block so it excluded some tests which are now part of the cleanup.

fair point

btw, looking at the later tests, there are more things to improve:
- "SNRC defaults": move config setup and creation into a BeforeEach()
- "Wrong self node remediation config CR": can be removed, we don't allow multiple configs anymore

mshitrit:
>as written, it can demonstrate the actual issue. I didn't want to propose adding a timeout permanently 

you are correct of course, I apologize for misunderstanding :+1:  


>I was explaining the issue we have in the current test code, not in the PR code. When you look at the current code, my explanation of the issue should make sense, does it?

I think I get what you mean, IIUC basically the BeforeEach of every test will trigger it's own resource creation (please correct me if I got it wrong again). 
The downside of this approach IMO is that we don't utilize the a shared block (BeforeEach/JustBeforeEach) for several tests, so each Test needs to have a BeforeEach block of it's own, I think the more common pattern is to create the resource in a Joined JustBeforeEach block and only in case it's necessary use individual BeforeEach block to modify the resources prior to creation.

>IMHO a Consistently with 1s for the ds deletion would be nicer than that boolean flag + Eventually for the getting the ds. Then we don't need to care about setting that flag for each test or not.

I think the downside of this approach that it would create false positives test results.
For example let's say that we delete and use a Consistently block to verify that the ds is gone: in case the timing was off (and the delete executed before the ds was created) the test will fail.
With the current approach it'll not happen and the ds will be successfully deleted.

>btw, looking at the later tests, there are more things to improve:

In general  agree it can be improved, ATM I prefer not to have the discussion about the best way to it in this PR.

slintes:
Maybe code can better explain what I mean 🤷🏼‍♂️ 
https://github.com/medik8s/self-node-remediation/pull/208

mshitrit:
>Maybe code can better explain what I mean 🤷🏼‍♂️
#208

Thanks for sharing that code, I understand better what you mean now, it's definitely an interesting approach I haven't considered.

I think though that there are couple of things to consider with this approach:
- At the first block (in case we didn't get the DS in the test) it is still needs to be cleaned since not getting it in the test does not mean it wasn't scheduled to be created, by skipping the clean in case it isn't found we risk the ds being created after the cleanup phase. 
- Using this approach we will always wait for the maximum time (2 seconds in this case) in order to verify the ds deleted (vs the other approach were we just wait until the resource is actually deleted)
- Because we always wait for the maximum time it would be an issue choosing a higher "safer" value (for example 5 or 10 seconds) for the verification. That means we need to settle for a lower wait time that might make the test less stable on resource heavy environment, comparing to the other approach where using a higher "maximum wait time" isn't an issue because it'll be applied only when necessary.
- I think it's more intuitive to understand the approach of  a) I've delete a resource b) I wait to make sure it's deleted Vs a) Delete the resource b) make sure it doesn't exist c) keep doing that to make sure it's actually deleted.  I do agree though  that this is my subjective point of view and others my have  a different opinion which is no less valid.

slintes:
> At the first block (in case we didn't get the DS in the test) it is still needs to be cleaned

That's what it's doing, not?

> we will always wait for the maximum time

Yes, but with the advantage of not having to care about cleanup by setting bools in the test code

> make the test less stable on resource heavy environment

This is just a unit test with the fake envtest api server. We just need to be sure that the config was reconciled and it created the ds. I don't expect that we need a higher duration than 2 seconds. I would even go with 1 second only.

mshitrit:
>That's what it's doing, not?

Not the way I understand the code.
In case the DS wasn't set in the test (i.e ds.GetName() == "") and it isn't found (for example because it wasn't created yet) we will immediately exit the Consistently block on the `return` statement without waiting for the 2 seconds period.

> I don't expect that we need a higher duration than 2 seconds

I wish I was that optimistic as well, but from what I've seen so far (there where several PRs in the past where I had to increase wait times)  there are situations where our unit tests needs more time.
I've found it to be particular true for `Test container build` phase which seems more flaky.


>Yes, but with the advantage of not having to care about cleanup by setting bools in the test code

Slowing all the tests down in order to avoid a variable isn't a good tradeoff IMO.

slintes:
> > That's what it's doing, not?
> 
> Not the way I understand the code. In case the DS wasn't set in the test (i.e ds.GetName() == "") and it isn't found (for example because it wasn't created yet) we will immediately exit the Consistently block on the `return` statement without waiting for the 2 seconds period.

No, the return is the same as leaving the function at the end, it does not change the behaviour of Consistently. The function will be called again after 500ms until the 2s is reached. I think it's not possible to leave Consitently early in any way, that's the whole point of Consistently in contrast to Eventually

mshitrit:
> No, the return is the same as leaving the function at the end, it does not change the behaviour of Consistently

Cool, didn't know that, thanks for the explanation :+1:

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -39,18 +44,40 @@ var _ = Describe("SNR Config Test", func() {
...
 		}
 
+		tmpConfig := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
+		tmpDs := &appsv1.DaemonSet{}
+		//verify ds and config exist
+		Eventually(func(g Gomega) bool {
+
+			g.Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(config), tmpConfig)).To(Succeed())
+			g.Expect(k8sClient.Get(context.Background(), dsKey, tmpDs)).To(Succeed())
+			return true
```
slintes, line 60: you never return false, so the bool is not needed. You can Just check for Succeed() below

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -39,18 +44,40 @@ var _ = Describe("SNR Config Test", func() {
...
+		Expect(k8sClient.Delete(context.TODO(), tmpConfig)).To(Succeed())
+		Expect(k8sClient.Delete(context.TODO(), tmpDs)).To(Succeed())
+
+		//verify delete is done
+		Eventually(func(g Gomega) bool {
+			err := k8sClient.Get(context.Background(), dsKey, &appsv1.DaemonSet{})
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			err = k8sClient.Get(context.Background(), client.ObjectKeyFromObject(config), &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{})
+			g.Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			return true
```
slintes, line 72: same here about redundant bool return value

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -70,15 +97,18 @@ var _ = Describe("SNR Config Test", func() {
...
 
 		It("Cert Secret should be created", func() {
-			Eventually(func() error {
+			Eventually(func(g Gomega) bool {
 				_, _, _, err := certReader.GetCerts()
-				return err
-			}, 15*time.Second, 250*time.Millisecond).ShouldNot(HaveOccurred())
+				g.Expect(err).ShouldNot(HaveOccurred())
+				g.Expect(k8sClient.Get(context.Background(), dsKey, ds)).To(Succeed())
+				return true
```
slintes, line 104: same here about redundant bool return value

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -169,39 +199,49 @@ var _ = Describe("SNR Config Test", func() {
...
 					return k8sClient.Create(context.Background(), ds)
 				}, 2*time.Second, 250*time.Millisecond).Should(Succeed())
 
 				Eventually(func() error {
-					return k8sClient.Get(context.Background(), key, ds)
+					return k8sClient.Get(context.Background(), dsKey, ds)
 				}, 2*time.Second, 250*time.Millisecond).Should(BeNil())
-
+			}
+			JustBeforeEach(func() {
```
slintes, line 211: this block is redundant, this is done in createInitialDs() already above

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -219,6 +259,10 @@ var _ = Describe("SNR Config Test", func() {
 		config.APIVersion = "self-node-remediation.medik8s.io/v1alpha1"
 		config.Name = "config-defaults"
 		config.Namespace = shared.Namespace
+		BeforeEach(func() {
+			//Config and DS aren't created so no need to clean them
+			isSkipCleanup = true
```
slintes, line 264: a) the lines above should be in this BeforeEach() as well
b) this config is not cleaned up?

slintes (changes requested):

controllers/tests/config/selfnoderemediationconfig_controller_test.go
mshitrit: I've looked into that.
>a) the lines above should be in this BeforeEach() as well

The reason the lines (specificlly `config := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}`)  were not in the BeforeEach block, is because the `config` in this context was shadowing the the general config var.
However this config was inteded to be a different instance (that is also the reason it wasn't cleaned, clean would have failed because it had a different name)

I've included it in the cleanup by aligning the name to that of the main config, I've also changed the var name to reduce the confusion.

slintes: I won't block on this, but IMHO instead of shadowing or using a new config, it would be even cleaner and less confusing to move the non default config values of the outer config into BeforeEach() of the "DS installation" context. Then we don't need a BeforeEach at this place at all. WDYT?

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -39,18 +44,43 @@ var _ = Describe("SNR Config Test", func() {
...
 		}
 
+		tmpConfig := &selfnoderemediationv1alpha1.SelfNodeRemediationConfig{}
+		tmpDs := &appsv1.DaemonSet{}
+		//verify config exist
+		Eventually(func(g Gomega) {
+			g.Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(config), tmpConfig)).To(Succeed())
+		}, 10*time.Second, 250*time.Millisecond).Should(Succeed())
+
+		//ds may take quite a while to create (for example with a 5 sec timeout this test usually fails locally)
```
slintes, line 60: This doesn't sound right, it should not need so long... maybe this is hiding another real issue in test setup or even production code 🤔 Will have a look...

controllers/tests/config/selfnoderemediationconfig_controller_test.go
mshitrit: Sure

controllers/tests/config/selfnoderemediationconfig_controller_test.go
slintes: I run the test successfully over 10 times now, with 1 sec only here, and above... 

It fails on your laptop? Any chance your laptop is very busy with other stuff? Otherwise needing more than 5 seconds here doesn't make sense IMHO. However, since this doesn't slow down tests if they run fast, lgtm

/lgtm

mshitrit: >It fails on your laptop?

Yep

>Any chance your laptop is very busy with other stuff?

I've retested and verified. CPU and Memory seems fine (I haven't check other criteria) 

In any case since we assume it's a resource associated issue I think I also prefer the "better safe than sorry" approach, especially considering previous experience with resource load in the CI environment.

PR #206, opened 2024-06-04
e2e test improvement

clobrano:
#### Why we need this PR

This PR introduces some required and nice-to-have improvements to E2E tests

#### Changes made

- Replace deprecated ginko progress with increased verbosity
- Improve logging
- Set labels for unrestricted PSA on namespace

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,7 +379,7 @@ protoc-gen-go-grpc: ## Download protoc-gen-go-grpc locally if necessary.
 e2e-test:
 	# KUBECONFIG must be set to the cluster, and PP needs to be deployed already
     # count arg makes the test ignoring cached test results
-	go test ./e2e -ginkgo.v -ginkgo.progress -test.v -timeout 60m -count=1 ${TEST_OPS}
+	go test ./e2e -ginkgo.vv -test.v -timeout 60m -count=1 ${TEST_OPS}
 
 .PHONY: operator-sdk
 OPERATOR_SDK_BIN_FOLDER = ./bin/operator-sdk
--- a/Makefile
+++ b/Makefile
@@ -203,7 +203,14 @@ bundle-cleanup: operator-sdk ## Remove bundle installed via bundle-run
 
 .PHONY: create-ns
 create-ns: ## Create namespace
-	$(KUBECTL) get ns $(OPERATOR_NAMESPACE) 2>&1> /dev/null || $(KUBECTL) create ns $(OPERATOR_NAMESPACE)
+	$(KUBECTL) get ns $(OPERATOR_NAMESPACE) 2>&1> /dev/null || \
+		($(KUBECTL) create ns $(OPERATOR_NAMESPACE) && $(MAKE) set-labels-to-namespace)
+
+.PHONY: set-labels-to-namespace
+set-labels-to-namespace: ## Set labels on NS as workaround for OLM pod not running with restricted PSA
+	oc label --overwrite ns $(OPERATOR_NAMESPACE) security.openshift.io/scc.podSecurityLabelSync=false
+	oc label --overwrite ns $(OPERATOR_NAMESPACE) pod-security.kubernetes.io/enforce=privileged
+
 
 ##@ Build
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -373,7 +380,7 @@ protoc-gen-go-grpc: ## Download protoc-gen-go-grpc locally if necessary.
 e2e-test:
 	# KUBECONFIG must be set to the cluster, and PP needs to be deployed already
     # count arg makes the test ignoring cached test results
-	go test ./e2e -ginkgo.v -ginkgo.progress -test.v -timeout 60m -count=1 ${TEST_OPS}
+	go test ./e2e -ginkgo.v -ginkgo.show-node-events -test.v -timeout 60m -count=1 ${TEST_OPS}
```
razo7, line 383: This is nice, it would be interesting to see the change of verbosity on failure runs!
https://onsi.github.io/ginkgo/#controlling-verbosity

> Very-verbose mode contains additional information over verbose mode. In particular, -vv timelines indicate when individual nodes start and end and also include the full failure descriptions for every failure encountered by the spec. Verbose mode does not include the node start/end events (though this can be turned on with --show-node-events) and does not include detailed failure information for anything other than the first (primary) failure. (Additional/subseuqent failures typically occur in clean-up nodes and are not as relevant as the primary failure that occurs in a subject or setup node).

It would have been nice to leverage that in unit testing as well at `test` target but it [doesn't use ginkgo](https://github.com/medik8s/self-node-remediation/pull/206/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R187)

Makefile
```diff
@@ -197,7 +197,14 @@ bundle-cleanup: operator-sdk ## Remove bundle installed via bundle-run
 
 .PHONY: create-ns
 create-ns: ## Create namespace
-	$(KUBECTL) get ns $(BUNDLE_RUN_NAMESPACE) 2>&1> /dev/null || $(KUBECTL) create ns $(BUNDLE_RUN_NAMESPACE)
+	$(KUBECTL) get ns $(BUNDLE_RUN_NAMESPACE) 2>&1> /dev/null || \
```
razo7, line 200: I have modified the var name [_BUNDLE_RUN_NAMESPACE_ to _OPERATOR_NAMESPACE_](https://github.com/medik8s/self-node-remediation/commit/4689318f2218803aad755f98bedf89d3707d9107), and it is used in e2e test (similarly to other Medik8s operators)

razo7: This why there is a conflict at the moment :)

clobrano: Yep, this was an actual WIP :D

Makefile
clobrano: > target but it doesn't use ginkgo

I don't think it's a problem, `go test` supports passing argument to ginkgo via `-gingko.<flag>` syntax

e2e/self_node_remediation_test.go
```diff
@@ -84,7 +84,11 @@ var _ = Describe("Self Node Remediation E2E", func() {
 			pod := findSnrPod(healthyNode)
 			logs, err := utils.GetLogs(k8sClientSet, pod)
 			Expect(err).ToNot(HaveOccurred())
-			logger.Info("logs of healthy self-node-remediation pod", "logs", logs)
+			logger.Info("BEGIN logs of healthy self-node-remediation pod", "pod", pod.GetName())
```
slintes, line 87: looks like the complete AfterEach block can be extracred into a function, it's used twice

clobrano: oh, thank you for notice that!
Yes, it is used completely equal twice _now_, because it should have been  
"logs of healthy self-node-remediation pod" and "logs of **un**healthy self-node-remediation pod" :sweat_smile: 
However, I can get around that and make a function

Makefile
clobrano: I'll try with `-vv` and see if it is enough alone, or it still needs `--show-node-events`

clobrano: `-vv` is equal to `-v` + `--show-node-events` + `detailed failure information`
This means that is quite valuable in e2e tests when run in CI.
I'm not completely sure it's as good as in unittests. We (can) also run them locally and in that context I lean over a short report, since I can always re-run with more specific option defining `TEST_OPTS` env variable

e2e/self_node_remediation_test.go
clobrano: nevermind it has always been "healthy" :see_no_evil:

Makefile
razo7: SGTM

slintes (approved):

clobrano:
rebased on main, let's see if the e2e tests benefit

PR #207, opened 2024-06-06
[WIP] Make template spec and strategy mandatory 

mshitrit:
#### Why we need this PR
ATM SNR template can be created without spec, even though that would make it completely unusable.

Note: Even though this might be considered an API change I don't think it's an issue because any pre-existing templates without spec or strategy are already invalid. 

#### Changes made
setting the template's spec and strategy as mandatory fields

ECOPROJECT-1999


clobrano:
do we also need to update the [sample](https://github.com/medik8s/self-node-remediation/blob/c4a785c94d521b4a5520dfe0c6e76f87e999e65f/config/samples/self-node-remediation_v1alpha1_selfnoderemediationtemplate.yaml#L1) ?

mshitrit:
> do we also need to update the [sample](https://github.com/medik8s/self-node-remediation/blob/c4a785c94d521b4a5520dfe0c6e76f87e999e65f/config/samples/self-node-remediation_v1alpha1_selfnoderemediationtemplate.yaml#L1) ?


Good question, but I think that since Automatic is still defined as default we are in the clear.

clobrano:
IIRC, the sample is used as template by UI (need to check though)

mshitrit:
> IIRC, the sample is used as template by UI (need to check though)
You are correct.

What I meant in my previous comment is that even though the template's strategy is empty when it's created in the UI it'll be populated with the default Automatic strategy

clobrano (approved):

api/v1alpha1/selfnoderemediation_types.go
```diff
@@ -46,7 +46,8 @@ type SelfNodeRemediationSpec struct {
 	// Automatic will choose the most appropriate strategy during runtime.
 	// +kubebuilder:default:="Automatic"
 	// +kubebuilder:validation:Enum=Automatic;ResourceDeletion;OutOfServiceTaint
-	RemediationStrategy RemediationStrategyType `json:"remediationStrategy,omitempty"`
+	// +required
```
slintes, line 49: why is this needed, we have the default value? :thinking:

slintes (changes requested):

api/v1alpha1/selfnoderemediation_types.go
mshitrit: I think there is value in making this field explicitly mandatory.

Even though as you've suggested it technically will not  be empty because of the default value, making it mandatory clears our intention as what we expect of this field.
This may also become relevant in case in the future we decide we want to remove the default value.

Do you see a specific downside for this that I'm missing  ?

slintes: The obvious downside is: this is different than it's done at all other places with default values 🤷🏼‍♂️ 

The less obvious but significant downside is (I wasn't sure about it, but I just tried it in test code at least): the validation for mandatory fields runs before setting default values. So you can't create SNRs anymore without setting the strategy, which makes the default value useless. Try it by removing the strategy in modified `client_server_test.go`.

slintes (approved):

clobrano:
Please update the PR title as well

PR #208, opened 2024-06-07
Fix flaky unit test

slintes:
#### Why we need this PR
The "DS Recreation on Operator Update"  unit tests are flaky, they fail to create the ds sometimes because it already exists. Reason that the outer `JustBeforeEach` runs first, which creates the SNR config, which async triggers creation of a ds. That conflicts with the creation of ds in the inner `JustBeforeEach`.

#### Changes made
- Move creation of the initial ds from `JustBeforeEach` to `BeforeEach`, so it runs before the outer `JustBeforeEach`.
- Improve cleanup by ensuring the ds is always deleted


slintes:
seems we go on with #204 🤷🏼‍♂️ 

/close

PR #209, opened 2024-06-09
handle configuration been deleted

mshitrit:
#### Why we need this PR
SNR should have only one default configuration, since the configuration affects the SNR agents which are running on each node (and every node has one agent).
Deleting this configuration will prevent the operator working properly, since preventing deletion of the configuration is problematic (for example it'll prevents OLM cleanup) we're making sure that SNR is properly disabled when the configuration is deleted.

#### Changes made
- Issue a webhook warning when the configuration is deleted
- Delay any outstanding remediation while there is no configuration

[ECOPROJECT-1996](https://issues.redhat.com//browse/ECOPROJECT-1996)


slintes:
/hold

see my comments on the issue

controllers/selfnoderemediation_controller.go
```diff
@@ -244,6 +251,27 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		return true, err
+	} else {
+		snrConfig := &v1alpha1.SelfNodeRemediationConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      v1alpha1.ConfigCRName,
+				Namespace: ns,
+			},
+		}
+		err := r.Get(ctx, client.ObjectKeyFromObject(snrConfig), snrConfig)
+		if (err != nil && apiErrors.IsNotFound(err)) || snrConfig.DeletionTimestamp != nil {
```
slintes, line 266: nit: the nil check isn't needed on the left side. Not nit: but on the right side, because the config can be nil. So it should be `	if apiErrors.IsNotFound(err) || err == nil && snr.GetDeletionTimestamp() != nil {`

controllers/selfnoderemediation_controller.go
```diff
@@ -168,6 +168,13 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		r.logger.Info("agent pod starting remediation on owned node")
 	}
 
+	if isConfigurationExist, err := r.isConfigurationExist(ctx); err != nil {
+		return ctrl.Result{}, err
+	} else if !isConfigurationExist {
+		r.logger.Info("snr is disabled because configuration does not exist, waiting for configuration creation ")
+		return ctrl.Result{RequeueAfter: time.Second * 30}, nil
```
slintes, line 175: two suggestions:
- set a status condition on the CR for better visibility of SNR being disabled
- and then in the config reconciler, when the config CR is created, update status on all SNR CRs again, which will automatically trigger reconcile. So no need for a fixed requeue here.
WDYT?

slintes (changes requested):

controllers/selfnoderemediation_controller.go
mshitrit: I like it, it's a nice solution.

controllers/selfnoderemediation_controller.go
```diff
@@ -185,6 +189,19 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}()
 
+	if isConfigurationExist, err := r.isConfigurationExist(ctx); err != nil {
+		return ctrl.Result{}, err
+	} else if !isConfigurationExist {
+		r.logger.Info("snr is disabled because configuration does not exist, waiting for configuration creation ")
+		meta.SetStatusCondition(&snr.Status.Conditions, metav1.Condition{
+			Type:   string(v1alpha1.DisabledConditionType),
+			Status: metav1.ConditionTrue,
+			Reason: string(snrDisabledNoConfig),
```
slintes, line 199: it's best practice to add a human readable message as well

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -343,3 +404,33 @@ func generateDs(name, namespace, version string) *appsv1.DaemonSet {
...
+	snrTmp := &selfnoderemediationv1alpha1.SelfNodeRemediation{}
+	Eventually(func(g Gomega) {
+		g.Expect(k8sClient.Get(context.Background(), client.ObjectKeyFromObject(snr), snrTmp)).To(Succeed())
+		snrTmp.Status.Conditions = []metav1.Condition{metav1.Condition{
+			Type:               "Disabled",
+			Status:             metav1.ConditionTrue,
+			Reason:             "SNRDisabledConfigurationNotFound",
+			LastTransitionTime: metav1.Time{Time: time.Now()},
+		}}
+		g.Expect(k8sClient.Status().Update(context.Background(), snrTmp)).To(Succeed())
```
slintes, line 434: Maybe we should verify that the update was persisted? Otherwise the test might return success before the CR even got the condition?

slintes (commented):
Something is wrong with this PR. It contains commits and modified code of changes which are in main already... :/

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -119,6 +119,34 @@ var _ = Describe("SelfNodeRemediationConfig Validation", func() {
 
 	})
 
+	Describe("deleting SelfNodeRemediationConfig CR", func() {
+		var conf *SelfNodeRemediationConfig
+		BeforeEach(func() {
+			conf = createDefaultSelfNodeRemediationConfigCR()
```
clobrano, line 125: It seems a bit misleading. The function is called `createDefaultSelfNodeRemediationConfigCR` but it doesn't create the "default" SNRC (i.e. the name is "test").

It might be better to use the default name in the function, and then change it to something else only when necessary (e.g. in "SelfNodeRemediationConfig CR is non default").

clobrano (changes requested):
Sorry, I still need to check the part where SNR is disabled, but I left some comments

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
mshitrit: > but it doesn't create the "default" SNRC

Good point !
I've removed the "default" from the method name, I think I prefer this over the alternative of changing the name.
Let me know if that's ok.

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -119,6 +119,34 @@ var _ = Describe("SelfNodeRemediationConfig Validation", func() {
...
+		})
+
+		When("SelfNodeRemediationConfig CR is default", func() {
+			BeforeEach(func() {
+				conf.Name = "self-node-remediation-config"
+				//Mock deployment namespace
+				_ = os.Setenv("DEPLOYMENT_NAMESPACE", conf.Namespace)
```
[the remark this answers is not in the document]
mshitrit: I'm not sure it's necessary in this case.
The reason is that the value set is the "expected" value, it just so happens that ATM it is required only in this test.
I assume that in the future in case we'll add more tests that requires this variable we will set it at higher level which is common to all the tests instead of only here.

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
```diff
@@ -119,6 +119,34 @@ var _ = Describe("SelfNodeRemediationConfig Validation", func() {
...
+			})
+			It("should have warning", func() {
+				war, err := conf.ValidateDelete()
+				Expect(err).To(Succeed())
+				Expect(war[0]).To(ContainSubstring("The default configuration is deleted, Self Node Remediation is now disabled"))
+			})
+		})
+
+		When("SelfNodeRemediationConfig CR is non default", func() {
```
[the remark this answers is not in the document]
mshitrit: In theory you are correct, however the create part belongs to different functionality so IMO no harm in testing that part as well in the delete module (it might become more relevant in case create functionality is modified).

Nothing I feel strongly about, so if you feel the confusion it creates override the value let me know and I'll remove the test.

controllers/selfnoderemediation_controller.go
```diff
@@ -244,7 +261,28 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 	return result, r.updateSnrStatusLastError(snr, err)
 }
 
-func (r *SelfNodeRemediationReconciler) updateConditions(processingTypeReason processingChangeReason, snr *v1alpha1.SelfNodeRemediation) error {
+func (r *SelfNodeRemediationReconciler) isConfigurationExist(ctx context.Context) (bool, error) {
+	if ns, err := utils.GetDeploymentNamespace(); err != nil {
+		r.logger.Error(err, "Failed getting snr namespace")
+		return true, err
```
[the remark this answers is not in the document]
mshitrit: Good comment, I think you are completely right

mshitrit:
>Something is wrong with this PR. It contains commits and modified code of changes which are in main already... :/

I might have messed it up when handling the merge conflicts, hopefully fine now.

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
clobrano: I think I'd prefer something like `CreateTestSelfNodeRemediationConfigCR` then, or making the name a input, for clarity

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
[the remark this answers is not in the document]
clobrano: I wonder if ignoring the deletion of non-default SNRC _only_ is a feature.
We are not intentionally allowing that, because we don't allow creating it in the first place.
It is kind of a side effect of the implementation we chose, while the feature we want is preventing the creation of the 2nd and the deletion of the last one, but this would require to _count_ the configurations.

controllers/selfnoderemediation_controller.go
```diff
@@ -185,6 +189,20 @@ func (r *SelfNodeRemediationReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 	}()
 
+	if isConfigurationExist, err := r.isConfigurationExist(ctx); err != nil {
+		return ctrl.Result{}, err
+	} else if !isConfigurationExist {
+		r.logger.Info("snr is disabled because configuration does not exist, waiting for configuration creation ")
```
clobrano, line 195: nit: Uppercase SNR (same below)

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -295,3 +299,24 @@ func (r *SelfNodeRemediationConfigReconciler) removeOldDsOnOperatorUpdate(ctx co
...
+func (r *SelfNodeRemediationConfigReconciler) setCRsEnabledStatus(ctx context.Context) error {
+	crs := selfnoderemediationv1alpha1.SelfNodeRemediationList{}
+	if err := r.List(ctx, &crs); err != nil {
+		r.Log.Error(err, "Failed to list self node remediation CRs")
+		return err
+	}
+
+	for _, cr := range crs.Items {
+		//is disabled
+		if meta.IsStatusConditionTrue(cr.Status.Conditions, string(selfnoderemediationv1alpha1.DisabledConditionType)) {
```
clobrano, line 312: This might be a bit misleading. The comment says "enable", but we remove the condition.
We can actually set it to false instead, as "advertisement" for the new condition (If users start seeing the condition set to False, they'll know it _can_ be True)

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -310,6 +311,40 @@ var _ = Describe("SNR Config Test", func() {
 		})
 	})
 
+	Context("Config is created for when SNR CR alreay exists", func() {
```
clobrano, line 314: typo s/alreay/already

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -310,6 +311,40 @@ var _ = Describe("SNR Config Test", func() {
...
+			DeferCleanup(func() {
+				Expect(k8sClient.Delete(context.Background(), snr)).To(Succeed())
+				Eventually(func(g Gomega) bool {
+					err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(snr), &selfnoderemediationv1alpha1.SelfNodeRemediation{})
+					return apierrors.IsNotFound(err)
+				}, 10*time.Second, 250*time.Millisecond).Should(BeTrue())
+			})
+			//simulating manager
+			setSNRStatusDisabled(snr)
+			//Verify status is presisted before continuing
```
clobrano, line 335: This seems a typo: "presisted"

clobrano (changes requested):

controllers/selfnoderemediationconfig_controller.go
mshitrit: Feels to me like something of a personal preference. I agree that for some conditions it's clearer to have either a false or a true value.

Specifically I think that for this one though I think having no condition is better than disabled false, trying to explain my rational:
-  the general  assumption is that the operator is working so I  having a Disabled False condition to describe this basic assumption seems a bit confusing
- I think that in case we want to have the Disabled False condition we should probably set it from the get go otherwise we can create a confusion about the difference between  having  a Disabled False condition (after a deleted config is created) and not setting this condition (when operator starts).

Having said all that it's not something I feel too strongly about :sweat_smile: , so let me know if you still prefer it changed.

P.S in case we decide to use this condition all the time I think that Enabled is probably better than Disabled

clobrano: My point on having it always set (enabled or disabled) was mostly as advertisement, but otherwise I'm OK with removing it.
I was only a bit confused by the term "set enable status", when what the function does is removing the _disabled_ condition.

mshitrit: NP, I'll improve the comment syntax

api/v1alpha1/selfnoderemediationconfig_webhook_test.go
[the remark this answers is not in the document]
mshitrit: NP, test removed.

controllers/selfnoderemediation_controller.go
slintes: > I don't think it could be nil since it's already initialized (it could be empty though).

fair point, I thought it can set to nil by the Get function, but that's not possible

controllers/selfnoderemediation_controller.go
```diff
@@ -221,6 +225,20 @@ func (r *SelfNodeRemediationReconciler) ReconcileManager(ctx context.Context, re
...
 
+	if isConfigurationExist, err := r.isConfigurationExist(ctx); err != nil {
+		return ctrl.Result{}, err
+	} else if !isConfigurationExist {
+		r.logger.Info("SNR is disabled because configuration does not exist, waiting for configuration creation ")
+		meta.SetStatusCondition(&snr.Status.Conditions, metav1.Condition{
+			Type:    string(v1alpha1.DisabledConditionType),
+			Status:  metav1.ConditionTrue,
+			Reason:  string(snrDisabledNoConfig),
+			Message: "snr is disabled because configuration does not exist",
```
slintes, line 236: nit: this is a user facing message, IMHO uppercase SNR or even full written SelfNodeRemediation would be better here

controllers/selfnoderemediation_controller.go
```diff
@@ -86,13 +86,17 @@ var (
...
 
-type processingChangeReason string
+type conditionReason string
 
 const (
-	remediationStarted              processingChangeReason = "RemediationStarted"
-	remediationTimeoutByNHC         processingChangeReason = "RemediationTimeoutByNHC"
-	remediationFinishedSuccessfully processingChangeReason = "RemediationFinishedSuccessfully"
-	remediationSkippedNodeNotFound  processingChangeReason = "RemediationSkippedNodeNotFound"
+	//Reasons related to ProcessingConditionType
```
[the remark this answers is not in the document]
slintes: I hit the wrong button, review still in progress...

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -281,6 +282,40 @@ var _ = Describe("SNR Config Test", func() {
 		})
 	})
 
+	Context("Config is created for when SNR CR already exists", func() {
```
slintes, line 285: nit: "for" can be deleted

pkg/reboot/calculator_test.go
```diff
@@ -38,13 +38,9 @@ var _ = Describe("Calculator tests", func() {
 	})
 
 	JustBeforeEach(func() {
-		Expect(k8sClient.Create(context.Background(), snrConfig)).To(Succeed())
+		createConfig(snrConfig)
```
slintes, line 41: can you please explain what the value of this change (introducing functions which do much more than needed IMHO) is? The tests are testing the config, they fail if it's not created, there is no need to test existence when creating and even less when deleting it :shrug:

slintes (commented):
some nits and comments

pkg/reboot/calculator_test.go
mshitrit: Sure. 
"GetRebootTime should return correct value" was flaky.
IMO it was because the config didn't create fast enough to be set in the calculator.

For me it makes sense to verify setup steps that takes before the test before the test starts, I find it easier to troubleshoot later.

I don't mind reverting this change, and you can introduce a different fix in a separate PR if that's something you prefer.

slintes: if it's flaky, than increasing the test timeout is an easier fix but with the same effect IMHO. WDYT?

mshitrit: I agree it's easier.
But I think the alternative is a better fix,  for the following reasons:
- it'll reduce the flakiness of all the tests that has a prerequisite of config and not just that specific one
- we have can have easier time troubleshooting in case  test fail - in case config isn't created we have an early indication instead of trying to to figure it out from the test failing (for example in this case we need to figure it out because GetRebootDuration doesn't return the expected value)
- I think it simplify the test workflow: since this test is not about testing etcd that config is created, but to see how the config creation affects the calculator I think verifying the config is created should not be part of the test but part of the setup.

slintes: > and not just that specific one

You need to increase the timeout on both tests of course

> in case config isn't created we have an early indication

To my best knowledge this never happened in unit tests so far, it's just more or less slow depending on the host. In the end the only effect of the added code is additional timeout. Or did you see any other issues than running into a timeout?

> we need to figure it out because GetRebootDuration doesn't return the expected value

The error message is pretty clear in that case.

And all that still doesn't explain the value of the existence test in the cleanup... the test would have failed in setup or in the actual test without config already, why fail it in cleanup as well?
Actually I see the old version would fail as well, which is unneeded, failures in the delete call can be ignored. The important part is that the config doesn't exist when the test finishes.

But ok, we won't agree anyway, not worth further discussion 🤷🏼‍♂️ 

/lgtm

slintes:
/hold

not sure if other threads are resolved

clobrano (approved):

clobrano:
My observations have been addressed. Lgtm too

pkg/reboot/calculator_test.go
mshitrit: >In the end the only effect of the added code is additional timeout. Or did you see any other issues than running into a timeout?

Not sure if you wanted a reply or not so feel free to ignore (I assumed you might want a reply because it was a question).

I'll try to explain myself better with an example.
I've simulated configuration not created on time for both use cases.

In the first use case we'll get the following error (see below), so we still needs to do some research as to the reason it failed. 
This is something I generally prefer to avoid by making sure that a test only tests what it's suppose to (in this case the value of   `GetRebootDuration` and not the creation of the configuration)
```
 SelfNodeRemediationConfig not set yet, can't calculate minimum reboot duration
      {
          msg: "SelfNodeRemediationConfig not set yet, can't calculate minimum reboot duration",
         ...
      }
```      

In the second use case (when create is verified before the test) we'll get the following error, which cuts down the required investigation and since this is done in a shared block the value applied to all current and future tests.

```
  The function passed to Eventually failed at /home/mshitrit/gitRepos/forked/medik8s/self-node/pkg/reboot/calculator_test.go:148 with:
  Expected success, but got an error:
      <*errors.StatusError | 0xc00055f040>: 
      SelfNodeRemediationConfig.self-node-remediation.medik8s.io "self-node-remediation-config" not found
      {
         ...
              Message: "SelfNodeRemediationConfig.self-node-remediation.medik8s.io \"self-node-remediation-config\" not found",
              ...
      }
```

PR #214, opened 2024-06-11
Improve safe time to reboot handling and more

slintes:
#### Why we need this PR
There are several issues in the current implementation of handling the safe time to reboot.
Context: the safe time reboot determines how long we wait until we assume that the node rebooted and no workloads are running anymore, before continuing with the remediation process and accelerating workload rescheduling by deleting pods, or using the ungraceful node shutdown feature by applying the OutOfService taint.
That time can be set in the SNRConfig (`SafeTimeToAssumeNodeRebootedSeconds`). We also calculate a minimum, which depends  on other values in the config, as well as cluster size and watchdog timeout.
The minimum time is calculated by all SNR agents on pod start, stored by unhealthy node's agent on the SNR CR during remediation, and then used by both the unhealthy node's agent and the manager for further remediation.

The issues are:

a) the agents crashloop in case the calculated time is lower than the specified time.
b) our own default value can be too low.
c) the calculated time might not be accurate anymore during remediation because of cluster size change.
d) the unhealthy node's agent might not be able to store it's calculated time during remediation, so the manager won't use it it and fallback to the specified value.
e) even when the agent would be able to store the value, there is a race condition with the manager which wants to store the specified value.
f) both unhealthy agent and the manager are running most of the remediation code in parallel, which leads to unneeded conflicts on resource updates, and hard to understand and maintain code execution flows. 

#### Changes made

These changes fix a+c+d+e:
- don't let the agent calculate the reboot duration on start. 
- instead, just store the watchdog timeout on the agent's node, and let the manager calculate the reboot time during remediation using the watchdog timeout from the unhealthy node, and the current cluster size.

This fixes a+b:
- make the specified value optional and don't set a default anymore. Don't fail in case it is given and lower than the calculated value, just use the calculated value instead.

This fixes f:
- make a clear separation in the SN config reconcile (=remediation) code flow between agent and manager code flow. The agent just reboots when needed. Everything else is handled by the manager alone.

More fixes done during development:
- fix reboot calculation by using `MaxTimeForNoPeersResponse` correctly
- deduplicate batch size calculation
- add unit tests for both reboot duration and batch size calculation

Fixes [ECOPROJECT-1875](https://issues.redhat.com//browse/ECOPROJECT-1875)
and more
Supersedes https://github.com/medik8s/self-node-remediation/pull/197

#### Test plan
- Unit tests are updated as needed
- TODO: add missing unit test for minimum reboot duration calculation
- e2e tests should not need any change, since this PR doesn't introduce any feature change


pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to calculate minimum reboot duration")
 	}
-	return s.timeToAssumeNodeRebooted
-}
 
-func (s *safeTimeCalculator) SetTimeToAssumeNodeRebooted(timeToAssumeNodeRebooted time.Duration) {
-	s.timeToAssumeNodeRebooted = timeToAssumeNodeRebooted
+	specRebootDurationSeconds := config.Spec.SafeTimeToAssumeNodeRebootedSeconds
+	if specRebootDurationSeconds != nil {
```
clobrano, line 74: Nit: Inverting the logic here should make it more readable:

```
if specRebootDurationSeconds == nil {
    log...
    return...
}
// continue with the rest
```

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
 }
 
-func (s *safeTimeCalculator) Start(_ context.Context) error {
-	return s.calcMinTimeAssumeRebooted()
-}
+func (r *rebootDurationCalculator) calculateMinimumRebootDuration(k8sClient client.Client, ctx context.Context, cfg *v1alpha1.SelfNodeRemediationConfig, watchdogTimeout time.Duration) (time.Duration, error) {
+
+	spec := cfg.Spec
+
+	// The reboot duration needs be at least the time we know we need for determining a node issue and trigger the reboot!
```
clobrano, line 93: Maybe clearer like: The reboot duration should be long enough to identify a node issue and trigger the reboot.

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
+	minTime += MaxTimeForNoPeersResponse
+
+	// 2. plus time for asking peers (10% batches + 1st smaller batch)
+	numBatches, err := r.calcNumOfBatches(k8sClient, ctx)
+	if err != nil {
+		return 0, errors.Wrap(err, "failed to calculate number of batches")
 	}
-	// 4. some buffer
-	minTime += 15 * time.Second
-	s.log.Info("calculated minTimeToAssumeNodeRebooted is:", "minTimeToAssumeNodeRebooted", minTime)
```
clobrano, line 106: keeping the log with the calculated time might be useful for debugging issues

clobrano (changes requested):
I left some comments, but didn't finish the review. I'll come back tomorrow

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
+
+	log := ctrl.Log.WithName("rebootDurationCalculator")
+
+	var config *v1alpha1.SelfNodeRemediationConfig
+	if r.snrConfig != nil {
+		config = r.snrConfig
+	} else {
+		// just in case we reconcile a SNR CR sooner than the SNRConfig CR
+		var err error
+		if config, err = r.getConfig(k8sClient, ctx, operatorNs); err != nil {
```
mshitrit, line 62: IIUC in case config fetch is successful we can set the config in the calculator

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
+	SetConfig(config *v1alpha1.SelfNodeRemediationConfig)
 }
 
-type safeTimeCalculator struct {
-	timeToAssumeNodeRebooted, minTimeToAssumeNodeRebooted                   time.Duration
-	wd                                                                      watchdog.Watchdog
-	maxErrorThreshold                                                       int
-	apiCheckInterval, apiServerTimeout, peerDialTimeout, peerRequestTimeout time.Duration
-	log                                                                     logr.Logger
-	k8sClient                                                               client.Client
```
mshitrit, line 43: What is the advantage of removing the client ? 
IIUC the downside is that it needs to be filtered down in multiple methods now.

controllers/selfnoderemediation_controller.go
```diff
@@ -146,7 +148,14 @@ func (r *SelfNodeRemediationReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		return r.ReconcileAgent(ctx, req)
+	}
+	return r.ReconcileManager(ctx, req)
+}
+
+func (r *SelfNodeRemediationReconciler) ReconcileAgent(ctx context.Context, req ctrl.Request) (returnResult ctrl.Result, returnErr error) {
+	r.logger = r.Log.WithValues("pod", "agent", "selfnoderemediation", req.NamespacedName)
 
 	snr := &v1alpha1.SelfNodeRemediation{}
 	if err := r.Get(ctx, req.NamespacedName, snr); err != nil {
```
mshitrit, line 161: I think it'll also make sense here to check for DeletionTimestamp != nil

controllers/selfnoderemediation_controller.go
```diff
@@ -600,11 +628,20 @@ func (r *SelfNodeRemediationReconciler) updateSnrStatus(ctx context.Context, snr
...
-	//we assume the unhealthy node will be rebooted by maxTimeNodeHasRebooted
-	maxTimeNodeHasRebooted := metav1.NewTime(metav1.Now().Add(r.SafeTimeCalculator.GetTimeToAssumeNodeRebooted()))
-	snr.Status.TimeAssumedRebooted = &maxTimeNodeHasRebooted
+func (r *SelfNodeRemediationReconciler) setTimeAssumedRebooted(ctx context.Context, node *v1.Node, snr *v1alpha1.SelfNodeRemediation) {
+	if snr.Status.TimeAssumedRebooted != nil {
+		// timeAssumedRebooted already set, nothing to do
+		return
+	}
+	// get reboot duration
+	rebootDuration, err := r.RebootDurationCalculator.GetRebootDuration(r.Client, ctx, node, r.MyNamespace)
```
mshitrit, line 637: I don't think it makes sense to ignore the error.

I suggest to either:
 - modify GetRebootDuration not to return an error (it can return the Fallback time in case it can't calculate the time) 
 - or address the returned error.

controllers/selfnoderemediationconfig_controller.go
```diff
@@ -49,11 +52,11 @@ const (
...
-	Log                       logr.Logger
-	Scheme                    *runtime.Scheme
-	InstallFileFolder         string
-	Namespace                 string
-	ManagerSafeTimeCalculator reboot.SafeTimeCalculator
+	Log                      logr.Logger
+	Scheme                   *runtime.Scheme
+	InstallFileFolder        string
+	Namespace                string
+	RebootDurationCalculator reboot.RebootDurationCalculator
```
mshitrit, line 59: double reboot in pkg name and interface

pkg/utils/annotations.go
```diff
@@ -61,3 +71,16 @@ func IsSoftwareRebootEnabled() (bool, error) {
...
+	if node.Annotations == nil {
+		return 0
+	}
+
+	timeout, err := strconv.Atoi(node.Annotations[WatchdogTimeoutSecondsAnnotation])
+	if err != nil {
+		return 0
+	}
+
+	return time.Duration(timeout) * time.Second
```
mshitrit, line 85: IIUC a use case where we can't get the value of the label isn't the expected flow.
As such, I think we should address it as an error.
Returning a 0 for this might be a risk of calculating a too low Safe time value.

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -25,102 +27,109 @@ import (
...
 
 const (
-	ConfigCRName                         = "self-node-remediation-config"
-	defaultWatchdogPath                  = "/dev/watchdog"
-	DefaultSafeToAssumeNodeRebootTimeout = 180
-	defaultIsSoftwareRebootEnabled       = true
+	ConfigCRName        = "self-node-remediation-config"
+	defaultWatchdogPath = "/dev/watchdog"
+	// FallbackRebootDuration is used by the manager in case it fails to calculate the minimum safe reboot time
+	FallbackRebootDuration         = 300 * time.Second
```
beekhof, line 33: Even the existence of this const scares me, but what on what basis do we have confidence that 5min is sufficient?
Seems like that should at least be documented here.

If we agree this const should exist, the agent/system should arguably consider it an error if they or the watchdog are configured with values that would result in the agent taking longer.  I don't see any checks for this.

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -25,102 +27,109 @@ import (
...
 	// SafeTimeToAssumeNodeRebootedSeconds is the time after which the healthy self node remediation
 	// agents will assume the unhealthy node has been rebooted, and it is safe to recover affected workloads.
 	// This is extremely important as starting replacement Pods while they are still running on the failed
 	// node will likely lead to data corruption and violation of run-once semantics.
 	// In an effort to prevent this, the operator ignores values lower than a minimum calculated from the
-	// ApiCheckInterval, ApiServerTimeout, MaxApiErrorThreshold, PeerDialTimeout, and PeerRequestTimeout fields.
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:default=180
-	SafeTimeToAssumeNodeRebootedSeconds int `json:"safeTimeToAssumeNodeRebootedSeconds,omitempty"`
+	// ApiCheckInterval, ApiServerTimeout, MaxApiErrorThreshold, PeerDialTimeout, and PeerRequestTimeout fields,
```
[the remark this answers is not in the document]
slintes: I think it is? It's like a list of fields, and something different.
Like: "the green, red, and small apple, and the yellow banana". Should not be "the green, red, small apple, and the yellow banana", should it? 🤔

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -25,102 +27,109 @@ import (
...
+	// and the unhealthy node's individual watchdog timeout.
+	// +optional
+	SafeTimeToAssumeNodeRebootedSeconds *int `json:"safeTimeToAssumeNodeRebootedSeconds,omitempty"`
 
+	// The timeout for api-server connectivity check.
 	// Valid time units are "ms", "s", "m", "h".
-	// +optional
 	// +kubebuilder:default:="5s"
-	// +kubebuilder:validation:Pattern="^(0|([0-9]+(\\.[0-9]+)?(ms|s|m|h)))$"
+	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
```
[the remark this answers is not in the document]
slintes: well..... since the very small units are technically allowed for duration, in practice there is no use case to use them IMHO. Documenting them might be more confusing than helpful. Having `ms` in the doc is strange already IMHO. WDYT?

pkg/reboot/calculator.go
slintes: agree, done

pkg/reboot/calculator.go
slintes: it's logged at the caller in every case

pkg/reboot/calculator.go
slintes: wfm, done

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,100 @@ const (
...
+		var err error
+		if config, err = r.getConfig(k8sClient, ctx, operatorNs); err != nil {
+			return 0, errors.Wrap(err, "failed to get SelfNodeRemediationConfig")
+		}
 	}
 
-	if s.timeToAssumeNodeRebooted < s.minTimeToAssumeNodeRebooted {
-		return s.minTimeToAssumeNodeRebooted
+	watchdogTimeout := utils.GetWatchdogTimeout(node)
+	minimumCalculatedRebootDuration, err := r.calculateMinimumRebootDuration(k8sClient, ctx, config, watchdogTimeout)
```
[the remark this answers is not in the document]
slintes: yes and no: in case the config wasn't set yet, the config which was got within this method, wasn't set to the field (which didn't make that much sense...).

However, after removing that part as suggested in your other comment, there only is yes to this 😉 Done

pkg/reboot/calculator.go
slintes: leftover from an earlier version, where I didn't even initialize the calculator in main, will revert 👍🏼

controllers/selfnoderemediation_controller.go
slintes: good catch 👍🏼

controllers/selfnoderemediation_controller.go
slintes: yes, in the context of returning an error when the config isn't set yet in the calculator, I already modified this to log and return the error, in order to always retry. There is a good chance that every possible error is temporary only. That way we also don't need the fallback.

controllers/selfnoderemediationconfig_controller.go
slintes: fixed

pkg/utils/annotations.go
slintes: Yes, it's not the expected flow. But returning an error and retrying doesn't make sense IMHO, because the chance that anyone is setting the annotation on the unhealthy node later on is near zero. So we would never go on with remediation.
I agree though that 0 isn't the best choice. Do you have a suggestion for a safe value?

api/v1alpha1/selfnoderemediationconfig_types.go
slintes: >  I don't see any checks for this.

there can't be any checks because this was the fallback in case we run into issues when calculating the minimum reboot duration. However, this fallback is removed now, we retry now (see https://github.com/medik8s/self-node-remediation/pull/214#discussion_r1643184005)

pkg/utils/annotations.go
slintes: I went with 60s, that's the maximum default timeout according to https://docs.kernel.org/watchdog/watchdog-parameters.html

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,106 @@ const (
...
-	}
+func (r *calculator) SetConfig(config *v1alpha1.SelfNodeRemediationConfig) {
+	r.snrConfig = config
 }
 
-func (s *safeTimeCalculator) GetTimeToAssumeNodeRebooted() time.Duration {
-	if !s.isAgent {
-		return s.timeToAssumeNodeRebooted
+// TODO add unit test!
+func (r *calculator) GetRebootDuration(k8sClient client.Client, ctx context.Context, node *v1.Node) (time.Duration, error) {
```
mshitrit, line 57: please remove k8sClient from method params

pkg/reboot/calculator.go
```diff
@@ -23,102 +23,106 @@ const (
...
+		"specified time in seconds", specRebootDuration.Seconds(), "calculated minimum time in seconds", minimumCalculatedRebootDuration.Seconds())
+	return specRebootDuration, nil
 }
 
-func (s *safeTimeCalculator) calcMinTimeAssumeRebooted() error {
-	if !s.isAgent {
-		return nil
-	}
-	// The reboot time needs be at least the time we know we need for determining a node issue and trigger the reboot!
+func (r *calculator) calculateMinimumRebootDuration(k8sClient client.Client, ctx context.Context, watchdogTimeout time.Duration) (time.Duration, error) {
```
mshitrit, line 95: please remove k8sClient from method params

pkg/utils/annotations.go
mshitrit: >But returning an error and retrying doesn't make sense IMHO

I Just wanted the error addressed and not ignored, I see you already did that by logging it :+1: 

>that's the maximum default timeout

That's what I had in mind as well, wasn't sure what's the exact number is though.

pkg/reboot/calculator.go
```diff
@@ -145,10 +148,11 @@ func (s *safeTimeCalculator) calcNumOfBatches() int {
...
 	}
-	//In order to stay on the safe side taking the largest calculated batch number (capped at 11)
-	if s.highestCalculatedBatchNumber < numberOfBatches {
-		s.highestCalculatedBatchNumber = numberOfBatches
-	}
-	return s.highestCalculatedBatchNumber
+	return numberOfBatches, nil
+}
 
+func (r *calculator) getConfig(k8sClient client.Client, ctx context.Context, namespace string) (*v1alpha1.SelfNodeRemediationConfig, error) {
```
mshitrit, line 154: please remove k8sClient from method params

pkg/reboot/calculator.go
slintes: hm, Goland's unused var / function warning used to work better... thx, fixed

pkg/reboot/calculator.go
slintes: the complete method can be removed 🙈

slintes:
/retest

is one of the new tests flaky...?

/test 4.12-test 
/test 4.13-test 
/test 4.14-test 
/test 4.15-test 
/test 4.16-test

slintes:
crap, retest was wrong the the github workflow of course 🙈

pkg/reboot/calculator.go
mshitrit: NP.
in this case I don't think it was considered an unused var because the method used the variable passed to it and not the one encapsulated in the struct.

controllers/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -177,25 +170,6 @@ var _ = Describe("SNR Config Test", func() {
 				}, 10*time.Second, 250*time.Millisecond).Should(Succeed())
 			})
 		})
-		When("SafeTimeToAssumeNodeRebootedSeconds is modified in Configuration", func() {
```
mshitrit, line 180: I think this test should be replaced and not just removed.
Instead of setting `SafeTimeToAssumeNodeRebootedSeconds` in the calculator when the configuration is changed we are setting the whole configuration which should be verified.

pkg/reboot/calculator_test.go
```diff
@@ -0,0 +1,128 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/medik8s/self-node-remediation/api/v1alpha1"
+	"github.com/medik8s/self-node-remediation/pkg/utils"
+)
+
+var _ = Describe("Calculator tests", func() {
```
mshitrit, line 21: Nice Tests !

pkg/utils/peers_test.go
```diff
@@ -0,0 +1,62 @@
...
+		})
+		It("GetNextBatchSize should return correct first batch size", func() {
+			Expect(GetNextBatchSize(nrOfNodes, nrOfNodes)).To(Equal(3))
+		})
+		It("GetNextBatchSize should return correct next batch size", func() {
+			Expect(GetNextBatchSize(nrOfNodes, nrOfNodes-3)).To(Equal(3))
+		})
+		It("GetNextBatchSize should return correct last batch size", func() {
+			Expect(GetNextBatchSize(nrOfNodes, 2)).To(Equal(2))
+		})
```
mshitrit, line 37: Nit: I think it'll be nicer to replace those tests with something like:
```
nodesLeft := nrOfNodes
expectedBatchesSize := []int{3, 3, 3, 3, 3, 3, 2, 0}
for _, expectedBatchSize := range expectedBatchesSize {
	actualBatchSize := GetNextBatchSize(nrOfNodes, nodesLeft)
	Expect(actualBatchSize).To(Equal(expectedBatchSize))
	nodesLeft -= actualBatchSize
}
```
it can even be simplified farther by initializing nodesLeft in a JustBeforeEach block shared for all the tests in this file.

controllers/tests/config/selfnoderemediationconfig_controller_test.go
slintes: I think that's covered in reboot_test.go, no?

pkg/utils/peers_test.go
slintes: good idea, done

controllers/tests/config/selfnoderemediationconfig_controller_test.go
mshitrit: I don't think so.
The idea behind this test is that both the Config reconciler and the Manager reconciler share the same instance of the calculator.
So we are verifying that a change done by one is reflected in the instance encapsulated by the other (because it's the same instance).

Feel free to point me to the relevant test in case it is covered and I've missed that.

slintes: sorry, I meant reboot/calculator_test.go
There we create 2 different configs, and ensure that GetRebootDuration() return the correct value. The configs are reconciled and set into the calculator, they are not set manually, see suite_test.go. The only thing which is manually done is setting the config to nil in the cleanup, just to be sure that the next test doesn't use old values.

api/v1alpha1/selfnoderemediationconfig_types.go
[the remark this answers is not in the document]
clobrano: I didn't see the were different groups :+1:

clobrano:
/lgtm
Keeping the "hold" because there are other threads open
/hold

slintes:
all discussions resolved

/hold cancel

PR #215, opened 2024-06-11
Trying to fix tide by adding explicit job name

slintes:
#### Why we need this PR
Tide isn't working

#### Changes made
Adding explicit job name to github worklow... just a try 🤷🏼‍♂️


clobrano (approved):

slintes:
will probably only have effect after merge

PR #216, opened 2024-06-11
Increase timeout on deleteAndWait in e2e

slintes:
#### Why we need this PR
Flaky e2e

#### Changes made
Increase timeout on deleteAndWait in e2e.
This used for deleting SNR CRs, which only works when remediation is done, which can take longer than 2 min.


clobrano (approved):

slintes:
failed test is known issue, will merge this because it reduces flakiness of e2e

PR #217, opened 2024-06-11
try difference in implicit and explicit job name...

slintes:
trying to get tide running


slintes:
status still shows the long `Pre Submit / Build (pull_request)`...

/close

PR #220, opened 2024-06-13
Use Pod IP for peer communication

clobrano:
#### Why we need this PR

SNR Peers communication uses hostnetwork (Node IP), which exposes a HTTP/2 endpoint.
Using the Pod IP will make the port harder to attack.

#### Changes made

In place of looking for Nodes' IP, we look for other agents' Pod IP.

https://issues.redhat.com/browse/ECOPROJECT-1879


pkg/apicheck/check.go
```diff
@@ -214,7 +213,7 @@ func (c *ApiConnectivityCheck) canOtherControlPlanesBeReached() bool {
 	return (healthyResponses + unhealthyResponses + apiErrorsResponses) > 0
 }
 
-func (c *ApiConnectivityCheck) popNodes(nodes *[][]v1.NodeAddress, count int) []string {
+func (c *ApiConnectivityCheck) popNodes(nodes *[]string, count int) []string {
```
mshitrit, line 216: IIUC the `nodes` variable here is actually the peer's address

pkg/apicheck/check.go
```diff
@@ -227,12 +226,13 @@ func (c *ApiConnectivityCheck) popNodes(nodes *[][]v1.NodeAddress, count int) []
 	//todo maybe we should pick nodes randomly rather than relying on the order returned from api-server
 	addresses := make([]string, count)
 	for i := 0; i < count; i++ {
-		nodeAddresses := (*nodes)[i]
-		if len(nodeAddresses) == 0 || nodeAddresses[0].Address == "" {
+		// TODO: shall we need to get "count" addresses anyway, replacing empty IP with another Node?
+		address := (*nodes)[i]
+		if address == "" {
 			c.config.Log.Info("ignoring node without IP address")
```
mshitrit, line 232: I think this log need to be edited

pkg/apicheck/check.go
```diff
@@ -227,12 +226,13 @@ func (c *ApiConnectivityCheck) popNodes(nodes *[][]v1.NodeAddress, count int) []
 	//todo maybe we should pick nodes randomly rather than relying on the order returned from api-server
 	addresses := make([]string, count)
 	for i := 0; i < count; i++ {
-		nodeAddresses := (*nodes)[i]
-		if len(nodeAddresses) == 0 || nodeAddresses[0].Address == "" {
+		// TODO: shall we need to get "count" addresses anyway, replacing empty IP with another Node?
+		address := (*nodes)[i]
```
mshitrit, line 230: Nit: maybe use something more specific than `address` (i.e for example podIp or podIpAddress)

pkg/peers/peers.go
```diff
@@ -111,37 +111,49 @@ func (p *Peers) updatePeers(ctx context.Context, getSelector func() labels.Selec
...
+	}
+
 	nodesCount := len(nodes.Items)
-	addresses := make([][]v1.NodeAddress, nodesCount)
+	addresses := make([]string, nodesCount)
 	for i, node := range nodes.Items {
-		addresses[i] = node.Status.Addresses
+		for _, pod := range pods.Items {
+			if pod.Spec.NodeName == node.Name {
+				addresses[i] = pod.Status.PodIP
```
slintes, line 136: I was wondering why we have a string type now, and indeed there is a better choice IMHO, what about using `pod.Status.PodIPs[0]`?

pkg/apicheck/check.go
```diff
@@ -130,35 +129,35 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
 		// but do not ask more than we have
-		if len(nodesToAsk) < nodesBatchCount {
-			nodesBatchCount = len(nodesToAsk)
+		if len(peersToAsk) < nodesBatchCount {
+			nodesBatchCount = len(peersToAsk)
 		}
 
-		chosenNodesAddresses := c.popNodes(&nodesToAsk, nodesBatchCount)
-		healthyResponses, unhealthyResponses, apiErrorsResponses, _ := c.getHealthStatusFromPeers(chosenNodesAddresses)
+		chosenPodIPs := c.popPeerIPs(&peersToAsk, nodesBatchCount)
```
slintes, line 159: nit: at other places we use `peer` instead of `node` or `pod`, what about naming this var `chosenPeerIPs` as well?

clobrano: you're right, just a typo here

pkg/peers/peers.go
clobrano: Do you mean moving around this data as [PodIP](https://github.com/medik8s/self-node-remediation/blob/c6a8cd32b8191d88020fd637cc9ca23de69b1078/vendor/k8s.io/api/core/v1/types.go#L3815-L3818) and then let [popPeersIP](https://github.com/medik8s/self-node-remediation/blob/c6a8cd32b8191d88020fd637cc9ca23de69b1078/pkg/apicheck/check.go#L216) deal with it returning the `string[]` of IPs?

clobrano: I mean, at which point of the chain would it be better to use the underlining `PodIP.IP`?
IIUC, the only interface requiring the `string` is [grpc.DialContext](https://pkg.go.dev/google.golang.org/grpc@v1.56.3#DialContext)

slintes: I would use PodIP everywhere where we used NodeAddress before.
But: oh, it's just a wrapper around a string, I expected a more IP-ish thing 😁  
And: oh, we did not even check the type of the NodeAddress in the old version of popNodes, and just took the first one 🙈

clobrano: >  But: oh, it's just a wrapper around a string, I expected a more IP-ish thing 😁

:grin: yep, moreover at the end of the day we use the `string`, so not sure it's worth it

mshitrit:
/lgtm
/hold
Holding since not sure if others threads are resolved - feel free to unhold in this is the case.

slintes:
code lgtm, but I would prefer to have enabled peer check e2e tests before merging this

slintes (approved):
/hold

wait for #226 being merged

clobrano:
> could not run steps: step [input:ocp-4.12-upi-installer] failed: failed to wait for importing imagestreamtag

4.12?

clobrano:
It doesn't seem something related to our test.

/retest

slintes:
> > could not run steps: step [input:ocp-4.12-upi-installer] failed: failed to wait for importing imagestreamtag
> 
> 4.12?

looks very unrelated, upi is also wrong, should be ipi IIUC

slintes:
meh, reopening and fixing #234

PR #225, opened 2024-07-03
Fix calculator lock and unit test

slintes:
#### Why we need this PR

calculator unit test times out sometimes because config wasn't set yet

#### Changes made

- fix lock when reading only (will probably have no significant effect though)
- increase unit test timeout to wait for config being set in calculator

### DIFF
--- a/pkg/reboot/calculator.go
+++ b/pkg/reboot/calculator.go
@@ -55,11 +55,10 @@ func (r *calculator) SetConfig(config *v1alpha1.SelfNodeRemediationConfig) {
 	r.snrConfig = config
 }
 
-// TODO add unit test!
 func (r *calculator) GetRebootDuration(ctx context.Context, node *v1.Node) (time.Duration, error) {
 
-	r.snrConfigLock.Lock()
-	defer r.snrConfigLock.Unlock()
+	r.snrConfigLock.RLock()
+	defer r.snrConfigLock.RUnlock()
 	if r.snrConfig == nil {
 		return 0, errors.New("SelfNodeRemediationConfig not set yet, can't calculate minimum reboot duration")
 	}
--- a/pkg/reboot/calculator_test.go
+++ b/pkg/reboot/calculator_test.go
@@ -88,7 +88,7 @@ var _ = Describe("Calculator tests", func() {
 		It("GetRebootTime should return correct value", func() {
 			Eventually(func() (time.Duration, error) {
 				return calculator.GetRebootDuration(context.Background(), unhealthyNode)
-			}, "5s", "200ms").Should(Equal(time.Duration(expectedRebootDurationSeconds) * time.Second))
+			}, "15s", "200ms").Should(Equal(time.Duration(expectedRebootDurationSeconds) * time.Second))
 		})
 	})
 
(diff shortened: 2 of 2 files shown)


clobrano (approved):

pkg/reboot/calculator.go
```diff
@@ -55,11 +55,10 @@ func (r *calculator) SetConfig(config *v1alpha1.SelfNodeRemediationConfig) {
 	r.snrConfig = config
 }
 
-// TODO add unit test!
 func (r *calculator) GetRebootDuration(ctx context.Context, node *v1.Node) (time.Duration, error) {
 
-	r.snrConfigLock.Lock()
-	defer r.snrConfigLock.Unlock()
+	r.snrConfigLock.RLock()
```
mshitrit, line 60: Nit: I missed the fact that this is a R/W Mutex lock.
IIUC there isn't a use case were we will apply multiple Read Locks,  so in practice we are using this as a regular lock.  I suggest using Mutex instead.

slintes: The RW lock is what we need: don't read while someone writes, and let write one tread only at a time. No need to lock reads while no one writes. No matter if this used today or not, it's the correct thing, not?

mshitrit: My point wasn't that it's not doing what it needs to do, but that it basically functions as a regular lock so we might as well use a regular lock for better readability/simplicity.

The idea behind RW locks is to allow multiple reads when there isn't a write lock which IIUC isn't the case here.

Consider the following example:

Participants
**A** - Config Reconciler (triggering the lock with  `SetConfig`)
**B** - Remediation Reconciler (Triggering the lock with `GetRebootDuration`)

**With RW Lock**
Use case 1 -> A gets the lock (Write lock) first  -> **result**: B is blocked until lock is released
Use case 2 -> B gets the lock (Read lock) first  -> **result**: A is blocked until lock is released

**With Regular Lock**
Use case 1 -> A gets the lock first  -> **result**: B is blocked until lock is released
Use case 2 -> B gets the lock first  -> **result**: A is blocked until lock is released

So we would have the exact same behavior with a regular lock.

RW lock would have made sense in case we would have had another participant : **C** -  Remediation Reconciler  2

In which case:

**RW Lock**
Use case 1 -> A gets the lock (Write lock) first  -> **result**: B and C are blocked until lock is released
use case 2 -> B gets the lock (Read lock) first  -> **result**: A is blocked until lock is released, but C can still get a Read lock.
use case 3 -> C gets the lock (Read lock) first  -> **result**: A is blocked until lock is released, but B can still get a Read lock.

slintes: > RW lock would have made sense in case we would have had another participant : C - Remediation Reconciler 2

yes, but why not allowing concurrent reads, even when today there is no other participant, and the reconciler is configured to use 1 thread only? Both can change 🤷🏼‍♂️

mshitrit: The way I see it, it's a simple case were we just not using the best tool for the job.
As it goes with such cases it creates unnecessary complexity and ambiguity (when I'd read this code I'll automatically assume that there are more than two parties involved because of the usage of RW lock) .

I don't think that arguing that future changes will make this a more compatible choice is a valid argument, unless there are concrete plans for such changes in the near future.

In this case, since the code is still functional and I think the added complexity and ambiguity are minimal I categorized this as a Nit.

PR #226, opened 2024-07-03
Some fixes and e2e test improvements

slintes:
- re-enable and fix api check log tests in e2e test
  - use service IP for killing API connection
  - kill API connection on SNR DS pod
  - add peer check server logs and use them for test which can't get logs from unhealthy node's SNR agent pod
  - wait for pod deletion only, not restart (restart is caused by reboot, not SNR)
- refactor / cleanup e2e tests
- fix owner / node name / machine name check in peer check server and agent reconciler
- update sort-imports, which ignores generated files now


controllers/owner_and_name.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, ownerRef := range snr.OwnerReferences {
+		if ownerRef.Kind == "NodeHealthCheck" {
+			return true, &ownerRef
+		}
+	}
+	return false, nil
+}
+
+// IsOwnedByMachine checks if the SNR CR is owned by a Machine CR.
+func IsOwnedByMachine(snr *v1alpha1.SelfNodeRemediation) (bool, *metav1.OwnerReference) {
```
mshitrit, line 53: Nit: this could be made private

controllers/owner_and_name.go
```diff
@@ -0,0 +1,96 @@
...
+func getNodeNameDirect(snr *v1alpha1.SelfNodeRemediation) string {
+	nodeName, isNodeNameAnnotationExist := snr.GetAnnotations()[commonAnnotations.NodeNameAnnotation]
+	if isNodeNameAnnotationExist {
+		return nodeName
+	}
+	return snr.GetName()
+}
+
+// IsOwnedByNHC checks if the SNR CR is owned by a NodeHealthCheck CR.
+func IsOwnedByNHC(snr *v1alpha1.SelfNodeRemediation) (bool, *metav1.OwnerReference) {
```
mshitrit, line 43: Nit: this could be made private

controllers/owner_and_name.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/openshift/api/machine/v1beta1"
+
+	"github.com/medik8s/self-node-remediation/api/v1alpha1"
+)
+
+// GetNodeName gets the node name:
+// - if owned by NHC, or as fallback, from annotation or CR name
+// - if owned by a Machine, from the Machine's node reference
+func GetNodeName(ctx context.Context, c client.Client, snr *v1alpha1.SelfNodeRemediation, log logr.Logger) (string, error) {
```
mshitrit, line 21: Nit: this could be made private

pkg/peerhealth/server.go
```diff
@@ -111,7 +101,8 @@ func (s *Server) Start(ctx context.Context) error {
 }
 
 // IsHealthy checks if the given node is healthy
-func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthResponse, error) {
+func (s *Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthResponse, error) {
+	s.log.Info("IsHealthy", "node", request.GetNodeName(), "machine", request.GetMachineName())
```
clobrano, line 105: do we want to keep this log?
If yes, we might improve it a little ("Another peer asking check", or something like that)

pkg/peerhealth/server.go
```diff
@@ -111,7 +101,8 @@ func (s *Server) Start(ctx context.Context) error {
 }
 
 // IsHealthy checks if the given node is healthy
-func (s Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthResponse, error) {
+func (s *Server) IsHealthy(ctx context.Context, request *HealthRequest) (*HealthResponse, error) {
```
clobrano, line 104: I made this change in my test PR, but I don't think it is really needed anymore

clobrano (changes requested):

pkg/peerhealth/server.go
slintes: you mean the pointer receiver? I made it today because some receivers were pointers, some not

pkg/peerhealth/server.go
slintes: yes I kept this on purpose, but good idea to improve the message

pkg/peerhealth/server.go
clobrano: Ah ok, so just a coincidence :grin: :+1:

controllers/owner_and_name.go
slintes: leftover from an earlier version, when it also was in a different package

clobrano (approved):
/lgtm
keeping the hold for the other threads
/hold

slintes:
strange error for ds pod creation in 4.13:
```
"container create failed: time="2024-07-09T17:26:51Z" level=warning msg="cgroup: subsystem does not exist" time="2024-07-09T17:26:51Z" level=warning msg="cgroup: subsystem does not exist" time="2024-07-09T17:26:51Z" level=warning msg="cgroup: subsystem does not exist" time="2024-07-09T17:26:51Z" level=error msg="runc create failed: unable to start container process: exec: \"/manager\": stat /manager: no such file or directory"
```

since 4.12 and 4.14 are green I will override. Let's keep an eye on it in #220 

/cherry-pick release-0.9

PR #235, opened 2024-07-14
Retry setting Out Of Service flags

mshitrit:
#### Why we need this PR
During release testing of SNR we've noticed one case were SNR template with Out Of Service taint strategy could not be created.
The reason for that was that because when SNR was initialized,  it failed to fetch the k8s version therefore SNR couldn't verify support for Out Of Service taint.
We couldn't reproduce this, but we suspect this occurred due to a temporary network issue. 
In order to minimize similar occurrences in the future a retry mechanism is added.

Adding links to our slack discussion on that subject. [[1]](https://redhat-internal.slack.com/archives/C03M5GKJNBA/p1720704425398739) [[2]](https://redhat-internal.slack.com/archives/C03M5GKJNBA/p1720809806541209)

#### Changes made
In case Out Of Service flags can't be set at the first attempt (due to failing getting the k8s version or any other reason) , several retries will be made before deciding Out Of Service flags isn't supported.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,7 +35,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -142,22 +141,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	interval := 2 * time.Second // retry every 2 seconds
-	timeout := 10 * time.Second // for a period of 10 seconds
-
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
-	defer cancel()
-
-	// Using wait.PollUntilContextTimeout to retry InitOutOfServiceTaintFlags in case there is a temporary network issue.
-	// Since the last internal error returned by InitOutOfServiceTaintFlags also indicates whether polling succeed or not, there is no need to also keep the context error returned by PollUntilContextTimeout.
-	_ = wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
-		if err = utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
-			return false, nil // Keep retrying
-		}
-		return true, nil // Success
-	})
-
-	if err != nil {
+	if err := utils.InitOutOfServiceTaintFlagsWithRetry(context.Background(), mgr.GetConfig()); err != nil {
 		setupLog.Error(err, "unable to verify out-of-service taint support. out-of-service taint isn't supported")
 	}
 
--- a/main.go
+++ b/main.go
@@ -35,6 +35,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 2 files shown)


beekhof:
How much longer is an OCP/k8s version without the taint still supported for?
Arguably we should restrict the retry to just this error case, but perhaps we can already kill the whole check?

Unrelated, the handling of `err` in the `else if` block of `InitOutOfServiceTaintFlags` seems likely to overwrite `err` if already set.

mshitrit:
> How much longer is an OCP/k8s version without the taint still supported for?

We use the taint as default since ocp 4.14 (we currently support 4.12+)

> Arguably we should restrict the retry to just this error case, but perhaps we can already kill the whole check?

The main purpose of the retry was to overcome a network issue which prevents fetching the k8s version, I think it would be tricky to assume that we can get the version needed to decide whether to apply the retry logic when there are network issues that would potentially trigger the retry.

> Unrelated, the handling of `err` in the `else if` block of `InitOutOfServiceTaintFlags` seems likely to overwrite `err` if already set.

In case the err is already set it'll be returned in the previous `if` block. I think it's a fairly common practice to reuse err var (under the assumption that it will only be overridden when error did not occur in the previous statement)

main.go
```diff
@@ -141,7 +142,22 @@ func main() {
...
+	interval := 2 * time.Second // retry every 2 seconds
+	timeout := 10 * time.Second // for a period of 10 seconds
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// Using wait.PollUntilContextTimeout to retry InitOutOfServiceTaintFlags in case there is a temporary network issue.
+	// Since the last internal error returned by InitOutOfServiceTaintFlags also indicates whether polling succeed or not, there is no need to also keep the context error returned by PollUntilContextTimeout.
+	_ = wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
+		if err = utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
```
clobrano, line 154: I understand that this is unlikely, however, IIUC, the poll might retry even if the error is
returned by `setOutOfTaintFlags`, which is not what we intend, isn'it?

I would suggest to apply the retry only to the part that gets the version from the cluster, either
moving it inside `InitOutOfServiceTaintFlags`, or rework the function into two separate parts
- first get the version (with retry)
- second call `setOutOfTaintFlags`

clobrano (changes requested):

main.go
mshitrit: > IIUC, the poll might retry even if the error is returned by setOutOfTaintFlags

I agree, but I don't think it's a problem.

Since InitOutOfServiceTaintFlags failing isn't a blocker I think it's fine to retry the whole method.
  - Best case, we will recover other use cases we haven't considered (another very unlikely use case is: we'll get a corrupted version due to network issue which will cause  setOutOfTaintFlags to fail and retrying the whole method will fix that )
  - Worst case, we'll run some unnecessary retries during SNR setup which will delay the setup by couple of seconds

clobrano:
/lgtm
/hold

to get more reviews and close the other threads

pkg/peerhealth/peerhealth.pb.go
```diff
@@ -7,10 +7,11 @@
 package peerhealth
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
```
slintes, line 12: This is very probably caused by an old `sort-import` binary on your machine, and since CI doesn't check for modified files (they are modified by the `generate` target), it isn't catched. Same for the other file. Why is no one else wondering about this unrelated change? 🤔

slintes (changes requested):

main.go
```diff
@@ -141,7 +142,22 @@ func main() {
...
-	if err := utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
+	interval := 2 * time.Second // retry every 2 seconds
+	timeout := 10 * time.Second // for a period of 10 seconds
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// Using wait.PollUntilContextTimeout to retry InitOutOfServiceTaintFlags in case there is a temporary network issue.
+	// Since the last internal error returned by InitOutOfServiceTaintFlags also indicates whether polling succeed or not, there is no need to also keep the context error returned by PollUntilContextTimeout.
+	_ = wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
```
slintes, line 153: I would prefer to keep main clean and do this inside InitOutOfServiceTaintFlags. WDYT?

main.go
```diff
@@ -141,7 +142,22 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
+	interval := 2 * time.Second // retry every 2 seconds
+	timeout := 10 * time.Second // for a period of 10 seconds
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
slintes, line 148: you don't need this, it's done within PollUntilContextTimeout...

slintes (changes requested):

pkg/peerhealth/peerhealth.pb.go
mshitrit: Cool thanks, I thought it's the other way around (i.e that the change was due to the new sort-import version).

> Why is no one else wondering about this unrelated change? 🤔

Probably others assumed the same as myself - I assume that the commit name contributed to that notion as well :sweat_smile:

main.go
mshitrit: I have no strong preference either way so that's fine by me.

pkg/utils/taints.go
```diff
@@ -72,6 +75,24 @@ func InitOutOfServiceTaintFlags(config *rest.Config) error {
 	}
 }
 
+// InitOutOfServiceTaintFlagsWithRetry tries to initialize the OutOfService flags based on k8s version, in case it fails (potentially due to network issues) it will retry for a limited number of times
+func InitOutOfServiceTaintFlagsWithRetry(ctx context.Context, config *rest.Config) error {
```
slintes, line 79: nit: public functions usually are above the private ones

slintes (approved):

slintes:
/hold 

in case you want to address the nit

main.go
```diff
@@ -141,7 +141,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := utils.InitOutOfServiceTaintFlags(mgr.GetConfig()); err != nil {
+	if err := utils.InitOutOfServiceTaintFlagsWithRetry(context.Background(), mgr.GetConfig()); err != nil {
```
beekhof, line 144: Nice improvement

PR #238, opened 2024-10-02
Configurable minimum worker nodecount

### LINKED ISSUE #213: Support for remediation on single worker node configurations
Existing code requires there to be at least one other peer worker node before remediation can occur, precluding SNR from remediating on a configuration with 3 control plane nodes + 1 worker node, which is a scenario that we support for bare minimum deployments.

novasbc:
#### Why we need this PR

Existing code requires there to be at least one other peer worker node before remediation can occur, precluding SNR from remediating on a configuration with 3 control plane nodes + 1 worker node, which is a scenario that we support for bare minimum deployments.

#### Changes made

- Add minPeersForRemediation configuration value.  It defaults to 1, which maintains backward compatibility with existing deployments
- Update getWorkerPeersResponse to take into account the new configuration value and not fail when there isn't another peer, and the user has configured the minimum to zero

Fixes #213

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 .PHONY: generate
 generate: controller-gen protoc ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations. Also generate protoc / gRPC code
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
-	PATH=$(PATH):$(shell pwd)/bin/proto/bin && $(PROTOC) --go_out=. --go-grpc_out=. pkg/peerhealth/peerhealth.proto
+	PATH='$(PATH)':$(shell pwd)/bin/proto/bin && $(PROTOC) --go_out=. --go-grpc_out=. pkg/peerhealth/peerhealth.proto
 
 .PHONY: fmt
 fmt: ## Run go fmt against code.
--- a/api/v1alpha1/selfnoderemediationconfig_types.go
+++ b/api/v1alpha1/selfnoderemediationconfig_types.go
@@ -28,6 +28,7 @@ const (
 	ConfigCRName                   = "self-node-remediation-config"
 	defaultWatchdogPath            = "/dev/watchdog"
 	defaultIsSoftwareRebootEnabled = true
+	defaultMinPeersForRemediation  = 1
 )
 
 // SelfNodeRemediationConfigSpec defines the desired state of SelfNodeRemediationConfig
@@ -127,6 +128,15 @@ type SelfNodeRemediationConfigSpec struct {
 	// CustomDsTolerations allows to add custom tolerations snr agents that are running on the ds in order to support remediation for different types of nodes.
 	// +optional
 	CustomDsTolerations []v1.Toleration `json:"customDsTolerations,omitempty"`
+
+	// Minimum number of peer workers/control nodes to attempt to contact before deciding if node is unhealthy or not
+	//	if set to zero, no other peers will be required to be present for remediation action to occur when this
+	//	node has lost API server access.  If an insufficient number of peers are found, we will not attempt to ask
+	//	any peer nodes (if present) whether they see that the current node has been marked unhealthy with a
+	//	SelfNodeRemediation CR
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=0
+	MinPeersForRemediation int `json:"minPeersForRemediation,omitempty"`
 }
 
 // SelfNodeRemediationConfigStatus defines the observed state of SelfNodeRemediationConfig
--- a/controllers/tests/controller/selfnoderemediation_controller_test.go
+++ b/controllers/tests/controller/selfnoderemediation_controller_test.go
@@ -66,6 +66,10 @@ var _ = Describe("SNR Controller", func() {
 		k8sClient.ShouldSimulateFailure = false
 		k8sClient.ShouldSimulatePodDeleteFailure = false
 		isAdditionalSetupNeeded = false
+
+		By("Restore default settings for api connectivity check")
+		apiConnectivityCheckConfig.MinPeersForRemediation = shared.MinPeersForRemediation
+
 		deleteRemediations()
 		deleteSelfNodeRemediationPod()
 		//clear node's state, this is important to remove taints, label etc.
@@ -451,6 +455,19 @@ var _ = Describe("SNR Controller", func() {
 				verifyWatchdogNotTriggered()
 			})
 		})
+
+		Context("no peer found and MinPeersForRemediation is configured to 0", func() {
+			BeforeEach(func() {
+				By("Set MinPeersForRemedation to zero which should trigger the watchdog before the test")
+				apiConnectivityCheckConfig.MinPeersForRemediation = 0
+			})
+
+			It("Does not receive peer communication and since configured to need zero peers, initiates a reboot",
+				func() {
+					verifyWatchdogTriggered()
+				})
+		})
+
 	})
 
 	Context("Configuration is missing", func() {
--- a/controllers/tests/controller/suite_test.go
+++ b/controllers/tests/controller/suite_test.go
@@ -53,13 +53,14 @@ import (
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
 var (
-	testEnv                 *envtest.Environment
-	dummyDog                watchdog.FakeWatchdog
-	unhealthyNode, peerNode = &v1.Node{}, &v1.Node{}
-	cancelFunc              context.CancelFunc
-	k8sClient               *shared.K8sClientWrapper
-	fakeRecorder            *record.FakeRecorder
-	snrConfig               *selfnoderemediationv1alpha1.SelfNodeRemediationConfig
+	testEnv                    *envtest.Environment
+	dummyDog                   watchdog.FakeWatchdog
+	unhealthyNode, peerNode    = &v1.Node{}, &v1.Node{}
+	cancelFunc                 context.CancelFunc
+	k8sClient                  *shared.K8sClientWrapper
+	fakeRecorder               *record.FakeRecorder
+	snrConfig                  *selfnoderemediationv1alpha1.SelfNodeRemediationConfig
+	apiConnectivityCheckConfig *apicheck.ApiConnectivityCheckConfig
 )
 
 var unhealthyNodeNamespacedName = client.ObjectKey{
@@ -152,14 +153,15 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 
 	rebooter := reboot.NewWatchdogRebooter(dummyDog, ctrl.Log.WithName("rebooter"))
-	apiConnectivityCheckConfig := &apicheck.ApiConnectivityCheckConfig{
-		Log:                ctrl.Log.WithName("api-check"),
-		MyNodeName:         shared.UnhealthyNodeName,
-		CheckInterval:      shared.ApiCheckInterval,
-		MaxErrorsThreshold: shared.MaxErrorThreshold,
-		Peers:              peers,
-		Rebooter:           rebooter,
-		Cfg:                cfg,
+	apiConnectivityCheckConfig = &apicheck.ApiConnectivityCheckConfig{
+		Log:                    ctrl.Log.WithName("api-check"),
+		MyNodeName:             shared.UnhealthyNodeName,
+		CheckInterval:          shared.ApiCheckInterval,
+		MaxErrorsThreshold:     shared.MaxErrorThreshold,
+		Peers:                  peers,
+		Rebooter:               rebooter,
+		Cfg:                    cfg,
+		MinPeersForRemediation: shared.MinPeersForRemediation,
 	}
 	apiCheck := apicheck.New(apiConnectivityCheckConfig, nil)
 	err = k8sManager.Add(apiCheck)
--- a/pkg/apicheck/check.go
+++ b/pkg/apicheck/check.go
@@ -50,6 +50,7 @@ type ApiConnectivityCheckConfig struct {
 	PeerRequestTimeout        time.Duration
 	PeerHealthPort            int
 	MaxTimeForNoPeersResponse time.Duration
+	MinPeersForRemediation    int
 }
 
 func New(config *ApiConnectivityCheckConfig, controlPlaneManager *controlplane.Manager) *ApiConnectivityCheck {
@@ -129,10 +130,29 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 
 	c.config.Log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
 	peersToAsk := c.config.Peers.GetPeersAddresses(peers.Worker)
-	if peersToAsk == nil || len(peersToAsk) == 0 {
-		c.config.Log.Info("Peers list is empty and / or couldn't be retrieved from server, nothing we can do, so consider the node being healthy")
+
+	// We check to see if we have at least the number of peers that the user has configured as required.
+	//  If we don't have this many peers (for instance there are zero peers, and the default value is set
+	//  which requires at least one peer), we don't want to remediate. In this case we have some confusion
+	//  and don't want to remediate a node when we shouldn't.  Note: It would be unusual for MinPeersForRemediation
+	//  to be greater than 1 unless the environment has specific requirements.
+	if len(peersToAsk) < c.config.MinPeersForRemediation {
+		c.config.Log.Info("Ignoring api-server error as we have an insufficient number of peers found, "+
+			"so we aren't going to attempt to contact any to check for a SelfNodeRemediation CR"+
+			" - we will consider it as if there was no CR present & as healthy.", "minPeersRequired",
+			c.config.MinPeersForRemediation, "actualNumPeersFound", len(peersToAsk))
+
 		// TODO: maybe we need to check if this happens too much and reboot
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersWereFound}
+	}
+
+	// If we make it here and there are no peers, we can't proceed because we need at least one peer
+	//  to check.  So it doesn't make sense to continue on - we'll mark as unhealthy and exit fast
+	if len(peersToAsk) == 0 {
+		c.config.Log.Info("Marking node as unhealthy due to being isolated.  We don't have any peers to ask "+
+			"and MinPeersForRemediation must be greater than zero", "minPeersRequired",
+			c.config.MinPeersForRemediation, "actualNumPeersFound", len(peersToAsk))
+		return peers.Response{IsHealthy: false, Reason: peers.UnHealthyBecauseNodeIsIsolated}
 	}
 
 	apiErrorsResponsesSum := 0
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 		if healthyResponses+unhealthyResponses+apiErrorsResponses > 0 {
 			c.timeOfLastPeerResponse = time.Now()
 		}
+		c.config.Log.Info("Aggregate peer health responses", "healthyResponses", healthyResponses,
+			"unhealthyResponses", unhealthyResponses, "apiErrorsResponses", apiErrorsResponses)
 
 		if healthyResponses > 0 {
-			c.config.Log.Info("Peer told me I'm healthy.")
+			c.config.Log.Info("There is at least one peer who thinks this node healthy, so we'll respond "+
+				"with a healthy status", "healthyResponses", healthyResponses, "reason",
+				"peers.HealthyBecauseCRNotFound")
 			c.errorCount = 0
 			return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseCRNotFound}
 		}
 
 		if unhealthyResponses > 0 {
-			c.config.Log.Info("Peer told me I'm unhealthy!")
+			c.config.Log.Info("I got at least one peer who thinks I'm unhealthy, so we'll respond "+
+				"with unhealthy", "unhealthyResponses", unhealthyResponses, "reason",
+				"peers.UnHealthyBecausePeersResponse")
 			return peers.Response{IsHealthy: false, Reason: peers.UnHealthyBecausePeersResponse}
 		}
 
 		if apiErrorsResponses > 0 {
-			c.config.Log.Info("Peer can't access the api-server")
+			c.config.Log.Info("If you see this, I didn't get any healthy or unhealthy peer responses, "+
+				"instead they told me they can't access the API server either", "apiErrorsResponses",
+				apiErrorsResponses)
 			apiErrorsResponsesSum += apiErrorsResponses
 			// TODO: consider using [m|n]hc.spec.maxUnhealthy instead of 50%
 			if apiErrorsResponsesSum > nrAllPeers/2 { // already reached more than 50% of the peers and all of them returned api error
 				// assuming this is a control plane failure as others can't access api-server as well
-				c.config.Log.Info("More than 50% of the nodes couldn't access the api-server, assuming this is a control plane failure")
+				c.config.Log.Info("More than 50% of the nodes couldn't access the api-server, assuming "+
+					"this is a control plane failure, so we are going to return healthy in that case",
+					"reason", "HealthyBecauseMostPeersCantAccessAPIServer")
 				return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseMostPeersCantAccessAPIServer}
 			}
 		}
 
 	}
 
+	c.config.Log.Info("We have attempted communication with all known peers, and haven't gotten either: " +
+		"a peer that believes we are healthy, a peer that believes we are unhealthy, or we haven't decided that " +
+		"there is a control plane failure")
+
 	//we asked all peers
 	now := time.Now()
 	// MaxTimeForNoPeersResponse check prevents the node from being considered unhealthy in case of short network outages
 	if now.After(c.timeOfLastPeerResponse.Add(c.config.MaxTimeForNoPeersResponse)) {
-		c.config.Log.Error(fmt.Errorf("failed health check"), "Failed to get health status peers. Assuming unhealthy")
+		c.config.Log.Error(fmt.Errorf("failed health check"), "Failed to get health status peers. "+
+			"Assuming unhealthy", "reason", "UnHealthyBecauseNodeIsIsolated")
 		return peers.Response{IsHealthy: false, Reason: peers.UnHealthyBecauseNodeIsIsolated}
 	} else {
-		c.config.Log.Info("Ignoring no peers response error, time is below threshold for no peers response", "time without peers response (seconds)", now.Sub(c.timeOfLastPeerResponse).Seconds(), "threshold (seconds)", c.config.MaxTimeForNoPeersResponse.Seconds())
+		c.config.Log.Info("Ignoring no peers response error, time is below threshold for no peers response",
+			"time without peers response (seconds)", now.Sub(c.timeOfLastPeerResponse).Seconds(),
+			"threshold (seconds)", c.config.MaxTimeForNoPeersResponse.Seconds(),
+			"reason", "HealthyBecauseNoPeersResponseNotReachedTimeout")
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersResponseNotReachedTimeout}
 	}
 
--- a/pkg/controlplane/manager.go
+++ b/pkg/controlplane/manager.go
@@ -61,17 +61,34 @@ func (manager *Manager) IsControlPlaneHealthy(workerPeerResponse peers.Response,
 	switch workerPeerResponse.Reason {
 	//reported unhealthy by worker peers
 	case peers.UnHealthyBecausePeersResponse:
+		manager.log.Info("We are deciding the control plane is not healthy because the peer response was UnHealthyBecausePeersResponse")
 		return false
 	case peers.UnHealthyBecauseNodeIsIsolated:
+		manager.log.Info("While trying to determine if the control plane is healthy, the peer response was "+
+			"UnHealthyBecauseNodeIsIsolated, so we are returning true if we could reach other control plane nodes",
+			"canOtherControlPlanesBeReached", canOtherControlPlanesBeReached)
 		return canOtherControlPlanesBeReached
 	//reported healthy by worker peers
 	case peers.HealthyBecauseErrorsThresholdNotReached, peers.HealthyBecauseCRNotFound, peers.HealthyBecauseNoPeersResponseNotReachedTimeout:
+		manager.log.Info("We are deciding that the control plane is healthy because either: "+
+			"HealthyBecauseErrorsThresholdNotReached "+
+			", HealthyBecauseCRNotFound,  or HealthyBecauseNoPeersResponseNotReachedTimeout",
+			"reason", workerPeerResponse.Reason)
 		return true
 	//controlPlane node has connection to most workers, we assume it's not isolated (or at least that the controlPlane node that does not have worker peers quorum will reboot)
 	case peers.HealthyBecauseMostPeersCantAccessAPIServer:
-		return manager.isDiagnosticsPassed()
+		didDiagnosticsPass := manager.isDiagnosticsPassed()
+		manager.log.Info("The peers couldn't access the API server, so we are returning whether "+
+			"diagnostics passed", "didDiagnosticsPass", didDiagnosticsPass)
+		return didDiagnosticsPass
 	case peers.HealthyBecauseNoPeersWereFound:
-		return manager.isDiagnosticsPassed() && canOtherControlPlanesBeReached
+		didDiagnosticsPass := manager.isDiagnosticsPassed()
+
+		manager.log.Info("We couldn't find any peers so we are returning didDiagnosticsPass && "+
+			"canOtherControlPlanesBeReached", "didDiagnosticsPass", didDiagnosticsPass,
+			"canOtherControlPlanesBeReached", canOtherControlPlanesBeReached)
+
+		return didDiagnosticsPass && canOtherControlPlanesBeReached
 
 	default:
 		errorText := "node is considered unhealthy by worker peers for an unknown reason"
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ bin
 
 # editor and IDE paraphernalia
 .idea
+*.iml
 *.swp
 *.swo
 *~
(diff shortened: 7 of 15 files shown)


slintes:
Hi, do you mind extending the description please? What's the issue, how do you fix it, how do you test the changes...
Also, please check the failed test.
Thanks

novasbc:
> Hi, do you mind extending the description please? What's the issue, how do you fix it, how do you test the changes... Also, please check the failed test. Thanks

Updated the description, included the issue # as well.

Also, fixed the build which was failing with 'make verify-bundle', because the bundle hadn't been updated.

slintes:
Thanks!

/test 4.16-openshift-e2e

novasbc:
@razo7 @mshitrit 

I looked into the e2e failures reported over the past few days and realized that it was due to temporary/environmental issues.  When I re-ran they started passing better.  We can't run the tests in an openshift environment, so weren't seeing the same things locally.

Anyhow, I believe this is ready for review.

Thanks!

pkg/apicheck/check.go
```diff
@@ -129,12 +130,17 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 
 	c.config.Log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
 	peersToAsk := c.config.Peers.GetPeersAddresses(peers.Worker)
-	if peersToAsk == nil || len(peersToAsk) == 0 {
-		c.config.Log.Info("Peers list is empty and / or couldn't be retrieved from server, nothing we can do, so consider the node being healthy")
-		// TODO: maybe we need to check if this happens too much and reboot
+	if peersToAsk == nil && c.config.MinPeersForRemediation != 0 || len(peersToAsk) < c.config.MinPeersForRemediation {
```
clobrano, line 133: It's a bit tricky, but as `len(peersToAsk)` is zero if `peersToAsk` is `nil`, I think you can get rid of the first part and just use `len(peersToAsk) < c.config.MinPeersForRemediation`.


* if `peersToAsk == nil` (and so `len(...) == 0`) and `c.config.MinPeersForRemediation != 0`, then also `len(peersToAsk) < c.config.MinPeersForRemediation` is True.
* For all the other combinations, we always need to evaluate the part after `||` anyway

pkg/apicheck/check.go
```diff
@@ -129,12 +130,17 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+		c.config.Log.Info("Peers list is empty and / or less than the minimum required peers for remediation, so consider the node being healthy")
+		//todo maybe we need to check if this happens too much and reboot
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersWereFound}
 	}
 
+	//if MinPeersForRemediation == 0 and there are no peers to contact, assume node is unhealthy
+	if peersToAsk == nil || len(peersToAsk) == 0 {
```
clobrano, line 140: I wonder if it is meaningful to allow `MinPeersForRemediation` to be zero

clobrano (changes requested):

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -127,6 +128,11 @@ type SelfNodeRemediationConfigSpec struct {
 	// CustomDsTolerations allows to add custom tolerations snr agents that are running on the ds in order to support remediation for different types of nodes.
 	// +optional
 	CustomDsTolerations []v1.Toleration `json:"customDsTolerations,omitempty"`
+
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=0
+	// Minimum number of peer workers/control nodes to attempt to contact before deciding if node is unhealthy or not
```
clobrano, line 134: I might understand the minimum value for workers, but for CP? Do we want to limit the peers to contact for both workers and CP, even if we have more than one?

slintes, line 134: please add what happens when this is set 0
nit: please move comment above the kubebuilder markers

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -170,6 +176,7 @@ func NewDefaultSelfNodeRemediationConfig() SelfNodeRemediationConfig {
 		Spec: SelfNodeRemediationConfigSpec{
 			WatchdogFilePath:        defaultWatchdogPath,
 			IsSoftwareRebootEnabled: defaultIsSoftwareRebootEnabled,
+			MinPeersForRemediation:  defaultMinPeersForRemediation,
```
slintes, line 179: this is redundant, will be set by api server because of the default defined in the kubebuilder marker

api/v1alpha1/selfnoderemediationconfig_types.go
slintes: also: does it make sense allow a higher value than 1? If so, what happens when we have less workers available? :thinking:

pkg/apicheck/check.go
slintes: this is purpose of the PR: change the behaviour from "we are healthy" to "we are unhealthy" in case we have no peer, in a non breaking way :)

slintes (changes requested):

Makefile
```diff
@@ -160,7 +160,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 .PHONY: generate
 generate: controller-gen protoc ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations. Also generate protoc / gRPC code
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
-	PATH=$(PATH):$(shell pwd)/bin/proto/bin && $(PROTOC) --go_out=. --go-grpc_out=. pkg/peerhealth/peerhealth.proto
+	PATH='$(PATH)':$(shell pwd)/bin/proto/bin && $(PROTOC) --go_out=. --go-grpc_out=. pkg/peerhealth/peerhealth.proto
```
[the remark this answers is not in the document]
novasbc: When some of our engineers build the solution in wsl2 based distributions, there ends up being spaces in the PATH variable

api/v1alpha1/selfnoderemediationconfig_types.go
mshitrit: In case higher value does not make sense, I suggest switching to a boolean value

pkg/apicheck/check.go
mshitrit, line 133: For me it's a bit tricky to read.
Hopefully I didn't miss anything (so feel free to correct me if I did :sweat_smile: )

I suggest something along these lines:
```
if  len(peersToAsk) == 0 {
   //if MinPeersForRemediation == 0 ...
   if c.config.MinPeersForRemediation == 0{
     return peers.Response{IsHealthy: false, Reason: peers.UnHealthyBecauseNodeIsIsolated}
   } 
  //TODO ...
   c.config.Log.Info("Peers list is empty and / or less than the minimum required peers for remediation, so consider the node being healthy")
 }
    
```

I think that this way it's clear we are addressing 2 different use cases where no peers where found (i.e peersToAsk is empty)

api/v1alpha1/selfnoderemediationconfig_types.go
novasbc: > also: does it make sense allow a higher value than 1? If so, what happens when we have less workers available? 🤔

We wanted to make this configurable such that for different configs that we have this being used in - different amounts of nodes and differing use cases, we gave flexibility of choice.

I do believe that in general it's going to be either zero or one, but didn't want to make that change when one of our configurations drive it.

pkg/apicheck/check.go
```diff
@@ -129,12 +130,17 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 
 	c.config.Log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
 	peersToAsk := c.config.Peers.GetPeersAddresses(peers.Worker)
-	if peersToAsk == nil || len(peersToAsk) == 0 {
-		c.config.Log.Info("Peers list is empty and / or couldn't be retrieved from server, nothing we can do, so consider the node being healthy")
-		// TODO: maybe we need to check if this happens too much and reboot
+	if peersToAsk == nil && c.config.MinPeersForRemediation != 0 || len(peersToAsk) < c.config.MinPeersForRemediation {
+		c.config.Log.Info("Peers list is empty and / or less than the minimum required peers for remediation, so consider the node being healthy")
+		//todo maybe we need to check if this happens too much and reboot
```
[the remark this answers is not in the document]
novasbc: > Please, keep the older "TODO:" format, some tools look for it to count the open items :)

Apologies, some automated tooling we are using made that change 🤷‍♂️

api/v1alpha1/selfnoderemediationconfig_types.go
novasbc: We do support single node cluster install for some configs that go out the door

novasbc: > please add what happens when this is set 0 nit: please move comment above the kubebuilder markers

I added more details and moved the comments

api/v1alpha1/selfnoderemediationconfig_types.go
novasbc: Removed the redundant line

pkg/apicheck/check.go
novasbc: I broke it up into the two cases, and removed checking for nil, and added some more comments to hopefully make it more understandable.

I didn't do the original code, but definitely can see where you are coming from.  Hope this makes it more clear.

pkg/apicheck/check.go
```diff
@@ -130,14 +130,23 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
-		c.config.Log.Info("Peers list is empty and / or less than the minimum required peers for remediation, so consider the node being healthy")
-		//todo maybe we need to check if this happens too much and reboot
+
+	// We check to see if we have at least the number of peers that the user has configured as required.
+	//	If we don't have this many peers (for instance there are zero peers, and the default value is set
+	//	which requires at least one peer), we don't want to remediate. In this case we have some confusion
+	//	and don't want to remediate a node when we shouldn't.  Note: It would be unusual for MinPeersForRemediation
+	//	to be greater than 1 unless the environment has specific requirements.
+	if len(peersToAsk) < c.config.MinPeersForRemediation {
+		c.config.Log.Info("Peers list is empty and / or less than the minimum required peers for remediation, " +
```
mshitrit, line 140: Log message can be confusing for this use case.
Assuming we don't use only 1 or 0 values for `MinPeersForRemediation ` Peer List might not be empty.

api/v1alpha1/selfnoderemediationconfig_types.go
novasbc: @clobrano checking to see if this can be resolved based on the latest changes.  Thanks!

pkg/apicheck/check.go
novasbc: @clobrano  checking to see if this can be resolved based on the latest changes.  Thanks!

api/v1alpha1/selfnoderemediationconfig_types.go
novasbc: @slintes checking to see if this can be resolved based on the latest changes.  Thanks!

slintes: You need to run make bundle for updating CRD files with the updated comment. Other than that lgtm 👍🏼

clobrano: Yes, after updating CRD files as Marc asked, lgtm

novasbc:
> hm, a unit test for this would be nice. I'm wondering if it's possible to add a test case similar to this one, but without peer and minPeers configured?
> 
> https://github.com/medik8s/self-node-remediation/blob/main/controllers/tests/controller/selfnoderemediation_controller_test.go#L438-L464

We have been unable to properly run e2e tests like the one you referenced in our infrastructure (not openshift) - we spent some decent cycles trying to get it to work.

Are you actually suggesting a unit test, or e2e? We did update one of the config unit tests, but that was super trivial & minor.

pkg/apicheck/check.go
novasbc: We did update this to try and make it more clear, and added more comments - and removed an unnecessary check.

novasbc:
If it's the case of an e2e test, we can try and write it and submit and run it through your infrastructure and some iterations - I just expect that to take quite some time to get right w/o the ability to run and debug locally.

slintes:
> Are you actually suggesting a unit test

I wrote unit test, and pointed to a unit test, so this is a yes :)

> but would love to potentially get this set of changes pushed upstream

And we would like to have a test that verifies that at least setting `minPeersForRemediation` to `0` has the desired effect.

novasbc:
> > Are you actually suggesting a unit test
> 
> I wrote unit test, and pointed to a unit test, so this is a yes :)
> 
> > but would love to potentially get this set of changes pushed upstream
> 
> And we would like to have a test that verifies that at least setting `minPeersForRemediation` to `0` has the desired effect.

Apologies, when I looked at the linked code, I thought it was actually executing as an e2e test, will look more closely.  Locally, they were not running with the unit tests with make test

novasbc:
Confirmed, I can get a unit test running in that context, it was misunderstanding of what was actually going on with the linked context.

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -445,22 +445,88 @@ var _ = Describe("SNR Controller", func() {
 			remediationStrategy = v1alpha1.ResourceDeletionRemediationStrategy
 		})
 
-		It("Verify that watchdog is not receiving food after some time", func() {
```
mshitrit, line 448: Hi, any particular reason this test was removed ?

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -445,22 +445,88 @@ var _ = Describe("SNR Controller", func() {
...
-				missedFeed := newTime.Before(lastFoodTime.Add(timeout))
-				if timeoutPassed && missedFeed {
-					return true
-				}
-				lastFoodTime = newTime
-				return false
-			}, 10*shared.PeerUpdateInterval, timeout).Should(BeTrue())
+		AfterEach(func() {
+			By("Restore default settings")
+			apiConnectivityCheckConfig.MinPeersForRemediation = 1
```
mshitrit, line 450: Nit: I think this makes more sense in the general After Each block

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -445,22 +445,88 @@ var _ = Describe("SNR Controller", func() {
...
+
+			// sleep so config can update
+			time.Sleep(time.Second * 2)
```
mshitrit, line 453: I don't think this is needed, IIUC config isn't being updated async here

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -445,22 +445,88 @@ var _ = Describe("SNR Controller", func() {
...
+		})
+
+		Context("no peer found, and using default setting for MinPeersForRemediation", func() {
+			BeforeEach(func() {
+				apiConnectivityCheckConfig.MinPeersForRemediation = 1
```
mshitrit, line 458: I think this is redundant because was already updated in the AfterEach block

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -445,22 +445,88 @@ var _ = Describe("SNR Controller", func() {
...
+				snrConfig.Spec.MinPeersForRemediation = 1
```
mshitrit, line 459: IIUC this is the default value, so I think it would make more sense to add it in `GenerateTestConfig`

controllers/tests/controller/suite_test.go
```diff
@@ -203,6 +205,134 @@ var _ = BeforeSuite(func() {
 
 })
 
+//var _ = BeforeEach(func() {
```
mshitrit, line 208: Leftovers ? probably can be removed

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: I verified that if I didn't put the sleep often times the config was not updated when the other routines ran.  It took me quite some time to track this particular problem down 😭

It's even worse if I disable some tests so things run in different orders

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: it wasn't removed, I changed the text to what I believed possibly described it better.  It's now in a separate context under the general context of "unhealthy node without api-server access" since both have similar requirements.

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: Done, moved to the general block up at the top.

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: Done, moved to GenerateTestConfig

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: done, removed - it was an artifact from before I had sleeps in place.

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: I was able to remove the sleep when putting it in beforeeach because there was already a sleep there

controllers/tests/controller/suite_test.go
```diff
@@ -152,14 +153,15 @@ var _ = BeforeSuite(func() {
...
-		Cfg:                cfg,
+	apiConnectivityCheckConfig = &apicheck.ApiConnectivityCheckConfig{
+		Log:                    ctrl.Log.WithName("api-check"),
+		MyNodeName:             shared.UnhealthyNodeName,
+		CheckInterval:          shared.ApiCheckInterval,
+		MaxErrorsThreshold:     shared.MaxErrorThreshold,
+		Peers:                  peers,
+		Rebooter:               rebooter,
+		Cfg:                    cfg,
+		MinPeersForRemediation: 1,
```
clobrano, line 164: Can we use the `MinPeersForRemediation` defined in `shared`?

pkg/apicheck/check.go
```diff
@@ -129,12 +130,28 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
 	c.config.Log.Info("Error count exceeds threshold, trying to ask other nodes if I'm healthy")
 	peersToAsk := c.config.Peers.GetPeersAddresses(peers.Worker)
-	if peersToAsk == nil || len(peersToAsk) == 0 {
-		c.config.Log.Info("Peers list is empty and / or couldn't be retrieved from server, nothing we can do, so consider the node being healthy")
+
+	// We check to see if we have at least the number of peers that the user has configured as required.
+	//	If we don't have this many peers (for instance there are zero peers, and the default value is set
+	//	which requires at least one peer), we don't want to remediate. In this case we have some confusion
+	//	and don't want to remediate a node when we shouldn't.  Note: It would be unusual for MinPeersForRemediation
+	//	to be greater than 1 unless the environment has specific requirements.
```
clobrano, line 138: In my editor it looks like there's a <TAB> at the beginning of each line of text (after //). Could you double check, please?

pkg/apicheck/check.go
```diff
@@ -129,12 +130,28 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+			"so we aren't going to attempt to contact any to check for a SelfNodeRemediation CR"+
+			" - we will consider it as if there was no CR present & as healthy.", "minPeersRequired",
+			c.config.MinPeersForRemediation, "actualNumPeersFound", len(peersToAsk))
+
 		// TODO: maybe we need to check if this happens too much and reboot
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersWereFound}
 	}
 
+	// If we make it here and there are no peers, we can't proceed because we need at least one peer
+	//	to check.  So it doesn't make sense to continue on - we'll mark as unhealthy and exit fast
```
clobrano, line 150: Same <TAB> appears in this comment section as above. Could you please check?

clobrano (changes requested):

controllers/tests/controller/suite_test.go
novasbc: Done

pkg/apicheck/check.go
novasbc: Yes, it was tabbed.  Replaced with spaces.

pkg/apicheck/check.go
novasbc: Yes, it was tabbed.  Replaced with spaces.

novasbc:
@mshitrit cleaned up leftovers, and moved things around per requests.  Looked at the latest updated commit with fresh eyes, and does seem pretty clean.

pretty sure the last time i looked at it my eyes were bleeding from maintaining a local version which still has the debug changes and some other stuff that can't go in yet 😔

slintes:
/hold

I'm having a hard to spot the difference between the old existing test and the new one, and either I'm misunderstanding something, the old test is broken, or the new doesn't add any value? Atm I tend to the 2nd option (why should it trigger a reboot / stop the the watchdog?). Let me do some tests on my machine...

novasbc:
My local branch has additions on the original test to validate that the watchdog isn't triggered. 

In the new test it validated that it WAS triggered.

slintes:
I think the old test succeeded, even though it shouldn't, because of this change: https://github.com/medik8s/self-node-remediation/pull/220/commits/cda2f3f9a5b93e592a88eb6b1fc7b0c85a250623#diff-9a251a7c17883073c468864af09499f7005e3dd80e4624c1c63c53868450b087R86-R92
When I remove the `cancel()` calls, the test fails (as I would expect). Those cancel calls stop the k8s manager, which stops its components, and the watchdog is one of them, so the watchdog disarms, and the test succeeds (no new feed found).

However, why did the test not fail before above change?
And why was the test even added like this? Without peers, and with default config, nothing should happen.
And obviously when running in a pod, it works like expected, that's the issue you actually have, right? No reboot ever without peers 🤷🏼‍♂️ 

I will discuss this with the team, and continue investigation tomorrow, and hopefully provide a fix asap...

novasbc:
> However, why did the test not fail before above change? And why was the test even added like this? Without peers, and with default config, nothing should happen. And obviously when running in a pod, it works like expected, that's the issue you actually have, right? No reboot ever without peers 🤷🏼‍♂️
> 
> I will discuss this with the team, and continue investigation tomorrow, and hopefully provide a fix asap...

I really struggled trying to make another test based on the original because I couldn't understand what it was intended to do precisely.

Spent quite a many hours walking through what it was doing, and tried to infer it - and like I said, updated the test to something that appeared to be working, and validating an actual useful behavior.

If it would help, I can post the branch to my private repo and you can take a look.

I was able to get it such that the watchdog was reset at the beginning of the tests each time - I had to add a reset method and a few other things to accomplish that, but I ended up with some other problems with the k8sclient and had to abandon ship at the moment due to time constraints, and really desiring to get the core change pushed someway somehow.

When I'm back to my corporate laptop I'll give you a bit more detail

slintes:
> When I remove the cancel() calls, the test fails

but only when it's focused... 🤔

slintes:
This fixes the issues with the existing unit tests: https://github.com/medik8s/self-node-remediation/pull/250

When you rebase your PR, you just need to call `verifyWatchdogTriggered()` in your new test.

novasbc:
> This fixes the issues with the existing unit tests: #250
> 
> When you rebase your PR, you just need to call `verifyWatchdogTriggered()` in your new test. And sorry for the time consuming trouble this caused :/

Awesome, glad to see i wasn't imagining things!  

I think this will fix some ordering issues I found with the tests as well when I would comment some tests out to decrease execution time while developing my new test 😁

controllers/tests/controller/selfnoderemediation_controller_test.go
novasbc: the test is completely changed now with the latest updates I rebased on top of, @mshitrit I believe this can be closed

controllers/tests/controller/suite_test.go
novasbc: new test update is significantly smaller, including leftovers 😂.  Believe this can be re-evaluated as well!

pkg/apicheck/check.go
```diff
@@ -129,12 +130,28 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+			c.config.MinPeersForRemediation, "actualNumPeersFound", len(peersToAsk))
+
 		// TODO: maybe we need to check if this happens too much and reboot
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersWereFound}
 	}
 
+	// If we make it here and there are no peers, we can't proceed because we need at least one peer
+	//  to check.  So it doesn't make sense to continue on - we'll mark as unhealthy and exit fast
+	if len(peersToAsk) == 0 {
+		return peers.Response{IsHealthy: false, Reason: peers.UnHealthyBecauseNodeIsIsolated}
```
slintes, line 152: please add a log statement before returning here, similar to the other returns

slintes (changes requested):
2 nits in the unit test, but I wouldn't block for them.
A log statement is missing though.

heads up, #250 needs be be merged first

slintes:
#250 is merged, but it got a little change, I'm afraid you need to rebase again. Also note my review above. Thanks!

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -451,6 +455,18 @@ var _ = Describe("SNR Controller", func() {
 				verifyWatchdogNotTriggered()
 			})
 		})
+
+		Context("no peer found and MinPeersForRemediation is configured to 0", func() {
+			It("Does not receive peer communication and since configured to need zero peers, initiates a reboot",
+				func() {
+					verifyWatchdogNotTriggered()
+
+					apiConnectivityCheckConfig.MinPeersForRemediation = 0
```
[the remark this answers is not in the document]
novasbc: thanks - I originally had it in a BeforeEach but due to other context I had thought that it was preferable by the team for it to be moved into the test - since it was only for this one individual test.

will move it back and submit after I'm able to get the tests running again.

controllers/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -451,6 +455,18 @@ var _ = Describe("SNR Controller", func() {
 				verifyWatchdogNotTriggered()
 			})
 		})
+
+		Context("no peer found and MinPeersForRemediation is configured to 0", func() {
+			It("Does not receive peer communication and since configured to need zero peers, initiates a reboot",
+				func() {
+					verifyWatchdogNotTriggered()
```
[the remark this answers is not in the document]
novasbc: So, I included this because I need to ensure that the watchdog hasn't already been triggered.  If the watchdog was already triggered, then the test would succeed as a false positive.

The other test may have tested it individually but that doesn't guarantee that by the time this test executes that it was in that state prior to execution.  Test ordering could have a dramatic effect on this.

I can remove it to be sure if you really want that, I just wanted to avoid possible future false positives.

controllers/tests/controller/selfnoderemediation_controller_test.go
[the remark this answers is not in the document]
slintes: it's a bit matter of taste, whether it's worth for a single test to have a BeforeEach() block... I think it's cleaner, but again, a nit only and no blocker 🙂

pkg/apicheck/check.go
novasbc: Done

controllers/tests/controller/selfnoderemediation_controller_test.go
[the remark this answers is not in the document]
novasbc: Removed the additional call to verify watchdog not triggered.

controllers/tests/controller/selfnoderemediation_controller_test.go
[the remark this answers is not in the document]
novasbc: Done, moved back to BeforeEach.  I happen to also prefer it there in general anyhow.

novasbc:
I've updated code associated to currently open requests.

I added a few more logging statements to cover a scenario we are finding in a specific customer configuration that would have been nice to have in the base build as well.

slintes:
the modified logs broke this e2e test: https://github.com/medik8s/self-node-remediation/blob/main/e2e/self_node_remediation_test.go#L155

novasbc:
> the modified logs broke this e2e test: https://github.com/medik8s/self-node-remediation/blob/main/e2e/self_node_remediation_test.go#L155

Thanks for the heads up, I'm used to the e2e tests failing randomly in the past so wouldn't have immediately come to that.

I'll take a look at the code this morning.  We were unable to run e2e locally unfortunately.

slintes:
>  I'm used to the e2e tests failing randomly in the past 

Unfortunately that's correct.
But when ALL e2e tests fail, probably something is wrong with the code 😉

pkg/apicheck/check.go
```diff
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 		if healthyResponses+unhealthyResponses+apiErrorsResponses > 0 {
 			c.timeOfLastPeerResponse = time.Now()
 		}
+		c.config.Log.Info("Aggregate peer health responses", "healthyResponses", healthyResponses,
+			"unhealthyResponses", unhealthyResponses, "apiErrorsResponses", apiErrorsResponses)
 
 		if healthyResponses > 0 {
-			c.config.Log.Info("Peer told me I'm healthy.")
+			c.config.Log.Info("There is at least one peer who thinks this node healthy, so we'll respond"+
```
slintes, line 173: missing space at end of string

pkg/apicheck/check.go
```diff
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+				"with a healthy status", "healthyResponses", healthyResponses, "reason",
+				"peers.HealthyBecauseCRNotFound")
 			c.errorCount = 0
 			return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseCRNotFound}
 		}
 
 		if unhealthyResponses > 0 {
-			c.config.Log.Info("Peer told me I'm unhealthy!")
+			c.config.Log.Info("I got at least one peer who thinks I'm unhealthy, so we'll respond"+
```
slintes, line 181: missing space at end of string

pkg/apicheck/check.go
```diff
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
-			c.config.Log.Info("Peer can't access the api-server")
+			c.config.Log.Info("If you see this, I didn't get any healthy or unhealthy peer responses, "+
+				"instead they told me they can't access the API server either", "apiErrorsResponses",
+				apiErrorsResponses)
 			apiErrorsResponsesSum += apiErrorsResponses
 			// TODO: consider using [m|n]hc.spec.maxUnhealthy instead of 50%
 			if apiErrorsResponsesSum > nrAllPeers/2 { // already reached more than 50% of the peers and all of them returned api error
 				// assuming this is a control plane failure as others can't access api-server as well
-				c.config.Log.Info("More than 50% of the nodes couldn't access the api-server, assuming this is a control plane failure")
+				c.config.Log.Info("More than 50% of the nodes couldn't access the api-server, assuming"+
```
slintes, line 195: missing space at end of string

pkg/apicheck/check.go
```diff
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+					"this is a control plane failure, so we are going to return healthy in that case",
+					"reason", "HealthyBecauseMostPeersCantAccessAPIServer")
 				return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseMostPeersCantAccessAPIServer}
 			}
 		}
 
 	}
 
+	c.config.Log.Info("We have attempted communication with all known peers, and haven't gotten either:" +
```
slintes, line 204: missing space at end of string

pkg/apicheck/check.go
```diff
@@ -146,39 +166,57 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+		"a peer that believes we are healthy, a peer that believes we are unhealthy, or we haven't decided that" +
```
slintes, line 205: missing space at end of string

pkg/controlplane/manager.go
```diff
@@ -61,17 +61,34 @@ func (manager *Manager) IsControlPlaneHealthy(workerPeerResponse peers.Response,
 	switch workerPeerResponse.Reason {
 	//reported unhealthy by worker peers
 	case peers.UnHealthyBecausePeersResponse:
+		manager.log.Info("We are deciding the control plane is not healthy because the peer response was UnHealthyBecausePeersResponse")
 		return false
 	case peers.UnHealthyBecauseNodeIsIsolated:
+		manager.log.Info("While trying to determine if the control plane is healthy, the peer response was"+
```
slintes, line 67: missing space at end of string

pkg/apicheck/check.go
```diff
@@ -129,12 +130,31 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
...
+			c.config.MinPeersForRemediation, "actualNumPeersFound", len(peersToAsk))
+
 		// TODO: maybe we need to check if this happens too much and reboot
 		return peers.Response{IsHealthy: true, Reason: peers.HealthyBecauseNoPeersWereFound}
 	}
 
+	// If we make it here and there are no peers, we can't proceed because we need at least one peer
+	//  to check.  So it doesn't make sense to continue on - we'll mark as unhealthy and exit fast
+	if len(peersToAsk) == 0 {
+		c.config.Log.Info("Marking node as unhealthy due to being isolated.  We don't have any peers to ask"+
```
slintes, line 152: missing space at end of string

slintes (changes requested):
sorry, missed it before, there a some spaces missing in concatinated log statements...

@mshitrit do you want to review as well? Beside those missing spaces I think the PR is good to go...

novasbc:
@slintes corrected spacing issues.  I have to stop multitasking with 800 other things going on 😂

pkg/apicheck/check.go
novasbc: Marking closed as I just double checked, I'm seeing spaces, not tabs as requested.

Wanting to have an accurate look to be sure there are no actual open defects still

pkg/apicheck/check.go
novasbc: Marking closed as I just double checked, we now have spaces instead of tabs

Wanting to have an accurate look to be sure there are no actual open defects still

slintes:
took a while, but:

/lgtm
/hold cancel

Thanks for your patience and good communication, and sorry again for the trouble with the unit test!

slintes (approved):

PR #239, opened 2024-10-31
Remove golang.org/x/crypto package Dependency

razo7:
#### Why we need this PR

Remove the `golang.org/x/crypto`package dependency when `github.com/Masterminds/sprig` is used. It is coming from running the `go mod why` command.
```
go mod why -m golang.org/x/crypto
# golang.org/x/crypto
github.com/medik8s/self-node-remediation/pkg/render
github.com/Masterminds/sprig
golang.org/x/crypto/scrypt
```
#### Changes made

1. Remove Sprig functions and `golang.org/x/crypto` package dependency. No longer add Sprig (200+) universal functions when rendering the template. Add dummy function as it is needed for 1 unit test.
2. Cleaning up unused modules using `go mod ...` commands.
3. Remove deprecated [function ReadFile](https://pkg.go.dev/io/ioutil#ReadFile) and package `io/ioutil` and using [`os.ReadFile`](https://pkg.go.dev/os#ReadFile)


razo7:
The changes in the PR are insufficient.
Before the `renderTemplate` function, the Sprig package was used to import (230) [universal functions](https://github.com/Masterminds/sprig/blob/master/functions.go#L96-L385) when some of the functions were implemented locally by Sprig. 
There is no clear (1-1) alternative function, and if that still a requirement to have all the "old" functions then we need to look for more options

razo7:
>The changes in the PR are insufficient.
Before the renderTemplate function, the Sprig package was used to import (230) [universal functions](https://github.com/Masterminds/sprig/blob/master/functions.go#L96-L385) when some of the functions were implemented locally by Sprig.
There is no clear (1-1) alternative function, and if that still a requirement to have all the "old" functions then we need to look for more options

The universal functions are redundant, so they are no longer needed.

pkg/render/render_test.go
```diff
@@ -89,7 +89,8 @@ func TestDir(t *testing.T) {
 	g := NewGomegaWithT(t)
 
 	d := MakeRenderData()
-	d.Funcs["fname"] = func(s string) string { return s }
+	d.Funcs["fname"] = func(s string) string { return "test-" + s }
```
slintes, line 92: this looks like a significant change to me, can you explain please?

razo7: Not sure why it was changed. I reverted it

slintes (approved):

PR #249, opened 2025-01-22
Fix Deletion timestamp check

mshitrit:
#### Why we need this PR
SNR should allow removal of redundant remediations - a remediation of a node that no longer exist since MDR deleted the node and triggered a provisioning of a new one.
Currently there is a bug in the code which prevents SNR reaching that logic, this PR fixes it.

#### Changes made
When SNR detects a remediation that is both deleted and marked as Timeout by NHC it should remove the finilizer. 
This PR removed an unnecessary check of the CR deletion timestamp that preceded this logic and prevented reaching this logic and the removal of the finilizer.

[ECOPROJECT-2489](https://issues.redhat.com//browse/ECOPROJECT-2489)

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -807,7 +807,7 @@ func (r *SelfNodeRemediationReconciler) removeNoExecuteTaint(node *v1.Node) erro
 }
 
 func (r *SelfNodeRemediationReconciler) isStoppedByNHC(snr *v1alpha1.SelfNodeRemediation) bool {
-	if snr != nil && snr.Annotations != nil && snr.DeletionTimestamp == nil {
+	if snr != nil && snr.Annotations != nil {
 		_, isTimeoutIssued := snr.Annotations[nhcTimeOutAnnotation]
 		return isTimeoutIssued
 	}


controllers/selfnoderemediation_controller.go
```diff
@@ -807,7 +807,7 @@ func (r *SelfNodeRemediationReconciler) removeNoExecuteTaint(node *v1.Node) erro
 }
 
 func (r *SelfNodeRemediationReconciler) isStoppedByNHC(snr *v1alpha1.SelfNodeRemediation) bool {
-	if snr != nil && snr.Annotations != nil && snr.DeletionTimestamp == nil {
+	if snr != nil && snr.Annotations != nil {
```
slintes, line 810: what about the check for deletion timestamp in line 245 (annotation is set, finalizer should be removed). Why is that needed? 🤔

mshitrit: The way I understand it the trigger for removing the finalizer should be removed when the node is healthy which in this case is indicated by the deletion of the CR.
In case we'll remove the check at 245, adding the timeout annotation will trigger the removal of the finalizer before the node is healthy.

Removing the remediation only based on the timeout annotation is also an option, IMO I think I prefer the first option though, since it aligns better with how we manage the finalizer in all other use cases.
I also like it better because I think breaking it into stages makes it easier to track :
- Step 1 (CR timeout) annotation added: triggering an event and updating condition
- Step 2 (Node is healthy) delete CR: remove finalizer

slintes: follow up after investigating a NHC e2e issue:

- keeping the finalizer on a timed out CR is correct actually!
- but when such a CR is deleted, we do not only need to remove the finalizer, but also cleanup taints

@mshitrit

PR #250, opened 2025-01-24
Fix watchdog usage in controller unit tests

slintes:
#### Why we need this PR
Controller unit tests are broken. They have false positives because they all use the same watchdog.
Also, one unit test is testing for the wrong result.

#### Changes made
- fix error handling in peers component
- add unit test for watchdog
- fixed issues in controller unit test

For details see commit messages.

Fixes issues with #238 

#### Test plan
There are new and fixed tests

### DIFF
--- a/pkg/peers/peers.go
+++ b/pkg/peers/peers.go
@@ -27,6 +27,9 @@ type Role int8
 const (
 	Worker Role = iota
 	ControlPlane
+
+	// this is used instead of "peerUpdateInterval" when peer update fails
+	quickPeerUpdateInterval = 2 * time.Minute
 )
 
 type Peers struct {
@@ -77,22 +80,24 @@ func (p *Peers) Start(ctx context.Context) error {
 		p.controlPlanePeerSelector = createSelector(hostname, getControlPlaneLabel(myNode))
 	}
 
-	var updatePeersError error
-	cancellableCtx, cancel := context.WithCancel(ctx)
-
 	p.log.Info("peer starting", "name", p.myNodeName)
-	wait.UntilWithContext(cancellableCtx, func(ctx context.Context) {
-		updatePeersError = p.updateWorkerPeers(ctx)
-		if updatePeersError != nil {
-			cancel()
-		}
-		updatePeersError = p.updateControlPlanePeers(ctx)
-		if updatePeersError != nil {
-			cancel()
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
+		updateWorkerPeersError := p.updateWorkerPeers(ctx)
+		updateControlPlanePeersError := p.updateControlPlanePeers(ctx)
+		if updateWorkerPeersError != nil || updateControlPlanePeersError != nil {
+			// the default update interval is quite long, in case of an error we want to retry quicker
+			quickCtx, quickCancel := context.WithCancel(ctx)
+			wait.UntilWithContext(quickCtx, func(ctx context.Context) {
+				quickUpdateWorkerPeersError := p.updateWorkerPeers(ctx)
+				quickUpdateControlPlanePeersError := p.updateControlPlanePeers(ctx)
+				if quickUpdateWorkerPeersError == nil && quickUpdateControlPlanePeersError == nil {
+					quickCancel()
+				}
+			}, quickPeerUpdateInterval)
 		}
 	}, p.peerUpdateInterval)
 
-	return updatePeersError
+	return nil
 }
 
 func (p *Peers) updateWorkerPeers(ctx context.Context) error {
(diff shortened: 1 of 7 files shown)


novasbc:
Looking at this briefly on my phone and it looks very familiar to what I was working on trying to accomplish on my local branch, and was having issues getting fully correct!

Thanks for putting this in, I should be able to rebase off this and I think it will do the trick.

Regards,
Mark

pkg/peers/peers.go
```diff
@@ -77,22 +80,24 @@ func (p *Peers) Start(ctx context.Context) error {
...
+		updateWorkerPeersError := p.updateWorkerPeers(ctx)
+		updateControlPlanePeersError := p.updateControlPlanePeers(ctx)
+		if updateWorkerPeersError != nil || updateControlPlanePeersError != nil {
+			// the default update interval is quite long, in case of an error we want to retry quicker
+			quickCtx, quickCancel := context.WithCancel(ctx)
+			wait.UntilWithContext(quickCtx, func(ctx context.Context) {
+				quickUpdateWorkerPeersError := p.updateWorkerPeers(ctx)
+				quickUpdateControlPlanePeersError := p.updateControlPlanePeers(ctx)
+				if quickUpdateWorkerPeersError == nil && quickUpdateControlPlanePeersError == nil {
+					quickCancel()
```
slintes, line 94: hm, does this cancel the parent context as well? That would be bad!

slintes: It doesn't, we're good here. Would be terrifying if a child can cancel the parent's and with that all sibling's contexts, right? 🙈

pkg/watchdog/watchdog_test.go
```diff
@@ -0,0 +1,96 @@
...
+			g.Expect(wd.Status()).To(Equal(watchdog.Armed))
+		}, 1*time.Second, 100*time.Millisecond).Should(Succeed(), "watchdog should be armed")
+	})
+
+	AfterEach(func() {
+		cancel()
+	})
+
+	Context("Watchdog started", func() {
+		It("should be feeded", func() {
```
clobrano, line 38: Maybe it's old code, but I think it's a typo: s/feeded/fed. Same below

pkg/watchdog/watchdog_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+		It("should be disarmed and and not be fed anymore", func() {
+			Eventually(func(g Gomega) {
+				g.Expect(wd.Status()).To(Equal(watchdog.Disarmed))
+			}, 1*time.Second, 100*time.Millisecond).Should(Succeed(), "watchdog should be disarmed")
+			verifyNoWatchdogFood(wd)
+		})
+	})
+
+	Context("Triggered watchdog resetted", func() {
```
clobrano, line 69: typo: resetted/reset

pkg/watchdog/watchdog_test.go
```diff
@@ -0,0 +1,96 @@
...
+		BeforeEach(func() {
+			wd.Stop()
+			wd.Reset()
+		})
+
+		It("should be armed and and be feeded", func() {
```
clobrano, line 75: typo: "add" repeated twice

clobrano (changes requested):
Just few typos (in old code probably, but we moved to a new file, so it is worth fixing it I think)

slintes:
/hold

@mshitrit in case you want to review

pkg/watchdog/interface.go
```diff
@@ -17,6 +17,9 @@ type Watchdog interface {
 	GetTimeout() time.Duration
 	// LastFoodTime return the last time the watchdog was fed
 	LastFoodTime() time.Time
+	// Reset resets the watchdog
+	// This very probably should never be used in production code, but for tests only!
+	Reset()
```
mshitrit, line 22: When possible, I think it's better not to update production code in order to accommodate test code.
This use case is a bit tricky, but we can probably wrap up the Fake dog with another layer that would have the `Reset` and `startFeeding` so those methods would be isolated in test code.
[Here](https://github.com/mshitrit/self-node/commit/e149b6975cf15bc2e5a9648977c00362848a04d0) is my attempt doing so.

slintes: fair enough, and thanks for your suggestion with code. I took a slightly different approach though (see updated last commit), by

- using a new FakeWatchdog interface instead of struct, and extending the original fake impl struct, I think that's a bit clearer than having 2 fake watchdog types
- keeping startFeeding() in the sync impl for preventing code duplication

WDYT?

mshitrit: looks great !
/lgtm

PR #252, opened 2025-01-29
Build file based catalog

Shai1-Levi:
#### Why we need this PR

Currently SNR Makefile creates a db-based catalog, which was deprecated

#### Changes made

This change create a file-based catalog

https://issues.redhat.com/browse/ECOPROJECT-2583

#### Test plan

* Create the file-based catalog and push it to my quay.io repository via Makefile
* Create a CatalogSource pointing to the new catalog
* Install SNR via WebUI, verify that everything works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,8 @@ YQ_VERSION = v4.44.1
 OPERATOR_NAME ?= self-node-remediation
 OPERATOR_NAMESPACE ?= openshift-workload-availability
 
+BLUE_ICON_PATH = "./config/assets/snr_icon_blue.png"
+
 # VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -455,15 +457,43 @@ ifeq (,$(wildcard $(OPM)))
...
+
 # Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
 # This recipe invokes 'opm' in 'semver' bundle add mode. For more information on add modes, see:
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-build
-catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMG) $(FROM_INDEX_OPT)
+catalog-build: opm ## Build a file-based catalog image.
+	# Remove the catalog directory and Dockerfile
+	-rm -r ${CATALOG_DIR} ${CATALOG_DOCKERFILE}
```
razo7, line 482: If I am not mistaken there is no such thing `-rm` it should be just `rm`,
```suggestion
	rm -r ${CATALOG_DIR} ${CATALOG_DOCKERFILE}
```

razo7:
/hold
The last commit should be avoided as I have missed the usage of hyphen

Makefile
clobrano: Yes, in a similar change (other operator) I added the hyphen to ignore the case the directory to remove does not exist already (which is ok)

PR #262, opened 2025-06-18
Fix health server timeout

slintes:
#### Why we need this PR
We observed cases when the k8s client code had some delay somewhere, which exceeded the configured k8s client timeout, and even the health client timeout on the caller side, causing unneeded fencing.

#### Changes made
- the health server monitors the timeout itself now, instead of relying on library code
- new unit tests cover the API server issue use cases (delay and error) 

[ECOPROJECT-2642](https://issues.redhat.com//browse/ECOPROJECT-2642)

## Summary by CodeRabbit

- **New Features**
  - Improved peer health checks to better handle backend API delays and errors, ensuring more accurate status reporting.
- **Bug Fixes**
  - Enhanced test coverage for scenarios involving API timeouts and errors, reducing the risk of incorrect health check results.
- **Refactor**
  - Streamlined internal handling of API call timeouts for more robust server behavior.
- **Tests**
  - Added new test cases simulating API timeouts and errors to validate health check responses.
  - Introduced thread-safe delay and error injection in test suite to simulate backend conditions more accurately.

### DIFF
--- a/pkg/peerhealth/suite_test.go
+++ b/pkg/peerhealth/suite_test.go
@@ -3,6 +3,7 @@ package peerhealth
 import (
 	"context"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
@@ -32,7 +33,7 @@ const nodeName = "somenode"
 
 var cfg *rest.Config
 var k8sClient client.Client
-var reader client.Reader
+var reader *ReaderWrapper
 var testEnv *envtest.Environment
 var snrReconciler *controllers.SelfNodeRemediationReconciler
 var cancelFunc context.CancelFunc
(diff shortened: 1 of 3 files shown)


pkg/peerhealth/suite_test.go
```diff
@@ -101,3 +106,25 @@ var _ = AfterSuite(func() {
 	cancelFunc()
 	Expect(testEnv.Stop()).To(Succeed())
 })
+
+type ReaderWrapper struct {
+	client.Reader
+	mu    sync.RWMutex
+	delay *time.Duration
+	err   error
```
beekhof, line 114: Is this only used for testing? I don't see any functional usage.

clobrano (approved):
I added a minor comment to the test, but it is ok for me

/lgtm

PR #264, opened 2025-06-22
Update peer timeout logic

mshitrit:
#### Why we need this PR
This PR address one of the issues found in [ECOPROJECT-2642](https://issues.redhat.com/browse/ECOPROJECT-2642)

The following in particular:
ATM apiServerTimeout is hard coded to 3 s
The reason that it is hardcoded to begin with is because both APIServerTimeout and PeerTimeout defaulted to 5 seconds. This might have created a scenario where Node A might time out waiting for a peer health check from Node B, while Node B is simultaneously experiencing a slow API server response. 
In that scenario Node A might have initiating remediation prematurely, assuming a network issue, when the actual problem is API server slowness.

The focus of the fix is:
- removing the hard coded apiServerTimeout value
- making sure we avoid the described scenario by updating the PeerTimeout value and overriding it in case an invalid value is used.

#### Changes made
- updated the default PeerTimeout value to a higher, safer value (e.g., 7 seconds).
- added a validating admission webhook that issues a warning if the configured PeerTimeout is less than APIServerTimeout + MinimumBuffer (e.g., 2 seconds).
- if the configured PeerTimeout is less than APIServerTimeout + MinimumBuffer, the operator will internally use APIServerTimeout + MinimumBuffer for the peer check.

[ECOPROJECT-2642](https://issues.redhat.com//browse/ECOPROJECT-2642)

## Summary by CodeRabbit

- **New Features**
  - Added safety validation and warnings if the peer request timeout is set too low compared to the API server timeout, helping prevent premature remediation actions.
  - Improved event recording and logging for timeout adjustments to enhance observability.

- **Bug Fixes**
  - Ensured peer request timeout is never set below a safe minimum relative to the API server timeout.

- **Chores**
  - Updated default peer request timeout from 5 seconds to 7 seconds across configuration, documentation, and tests to improve reliability.

### DIFF
--- a/api/v1alpha1/selfnoderemediationconfig_types.go
+++ b/api/v1alpha1/selfnoderemediationconfig_types.go
@@ -94,7 +94,7 @@ type SelfNodeRemediationConfigSpec struct {
 
 	// Timeout for each peer request.
 	// Valid time units are "ms", "s", "m", "h".
-	// +kubebuilder:default:="5s"
+	// +kubebuilder:default:="7s"
 	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$"
 	// +kubebuilder:validation:Type:=string
 	// +optional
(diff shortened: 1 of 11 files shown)


mshitrit:
@coderabbitai full review

PR #270, opened 2025-07-31
[WIP] Add an expiry mechanism to out-of-service taint.

mshitrit:
#### Why we need this PR
We've encountered some rare use cases where stale workload stays on healthy nodes and isn't removed by the out of service taint.
This was preventing the removal of the taint and usage of the node.

#### Changes made
Adding an expiry mechanism to the Out Of Service taint, taint will be removed if expiry time was reached.

RHWA-216

### DIFF
--- a/controllers/selfnoderemediation_controller.go
+++ b/controllers/selfnoderemediation_controller.go
@@ -44,24 +44,26 @@ import (
 )
 
 const (
-	SNRFinalizer            = "self-node-remediation.medik8s.io/snr-finalizer"
-	nhcTimeOutAnnotation    = "remediation.medik8s.io/nhc-timed-out"
-	excludeRemediationLabel = "remediation.medik8s.io/exclude-from-remediation"
+	SNRFinalizer                    = "self-node-remediation.medik8s.io/snr-finalizer"
+	nhcTimeOutAnnotation            = "remediation.medik8s.io/nhc-timed-out"
+	excludeRemediationLabel         = "remediation.medik8s.io/exclude-from-remediation"
+	outOfServiceTimestampAnnotation = "remediation.medik8s.io/oos-timestamp"
 
 	eventReasonRemediationSkipped = "RemediationSkipped"
 
 	// remediation
-	eventReasonAddFinalizer              = "AddFinalizer"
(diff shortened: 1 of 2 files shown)


mshitrit:
>Removing it on a unhealthy node does not make that much sense to me... 🤔

Hmm interesting point.
The way I look at it this taint in `OutOfService` strategy is the equivalent of SNR deleting the pods in  `ResourceDeletion` strategy (in which this taint does not exist at all).
So the action of deleting the resources (in either startegy) is dependent on the node being fenced and not healthy.
What's the value in keeping the taint ?

mshitrit:
>Before adding the taint node.kubernetes.io/out-of-service, it should be verified that the node is already in shutdown or power off state (not in the middle of restarting).

We already don't do that.

>The user is required to manually remove the out-of-service taint after the pods are moved to a new node

We don't plan to do that either (let me know if I'm misinterpreting your intention here)

>user has checked that the shutdown node has been recovered

I don't think that blindly following k8s guidelines in this particular case is a good argument here, especially since in all other cases we don't (I was looking for a rational that I might have missed as to why this approach creates an additional risk).

In any case definitely not a deal breaker as far as I'm concerned, so I'm fine with removing the taint from a healthy node.

controllers/selfnoderemediation_controller.go
```diff
@@ -81,6 +83,9 @@ var (
 		Value:  "nodeshutdown",
 		Effect: v1.TaintEffectNoExecute,
 	}
+
+	// OutOfServiceTimeoutDuration - time after which out-of-service taint is automatically removed
+	OutOfServiceTimeoutDuration = 1 * time.Minute
```
razo7, line 88: Shouldn't it be a configurable value?

mshitrit:
@razo7 , @clobrano let me know if you want to further discuss [this](https://github.com/medik8s/self-node-remediation/pull/270#discussion_r2272396469) or if it's clear to merge.

controllers/selfnoderemediation_controller.go
clobrano: No strong opinion, we can expose the configuration later, eventually 👍

clobrano (approved):

PR #273, opened 2025-11-18
Change Deployment UpdateStrategy

kattz-kawa:
#### Why we need this PR

- After new placement rule introduced in #180, the manager pods can only be scheduled on two specific nodes. Under these conditions, the default strategy may cause rollout deadlocks during node recovery.
- Certificate rotation managed by OLM can trigger Deployment rollouts by changing the `olmcahash` annotation. With the current strategy, this may lead to deadlocks if a node is down during recovery, causing rollout failures. To improve reliability and ensure safe updates under these conditions, we need a more controlled approach.

#### Changes made
- Added `RollingUpdate` strategy with `maxSurge: 0` and `maxUnavailable: 1` for safer deployment updates.

- Before (default): When a rollout starts, the Deployment tries to keep the old pod until the new one is ready. On clusters with only 2 nodes, this can fail because there is no room to schedule the new pod, leading to a stuck rollout.

- After: Pods are updated one at a time. The controller deletes one pod first, then creates a new one, ensuring that rollout can proceed safely even when only 2 nodes are available.

[RHWA-366](https://issues.redhat.com/browse/RHWA-366)

## Summary by CodeRabbit

* **Chores**
  * Updated deployment strategy configuration for controller-manager components to implement RollingUpdate strategy with specified constraints on pod surges and allowed unavailability. These operational improvements ensure more controlled and stable application updates while maintaining better service availability throughout upgrade processes.

### DIFF
--- a/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
+++ b/bundle/manifests/self-node-remediation.clusterserviceversion.yaml
@@ -273,7 +273,11 @@ spec:
             matchLabels:
               control-plane: controller-manager
               self-node-remediation-operator: ""
-          strategy: {}
+          strategy:
+            rollingUpdate:
+              maxSurge: 0
+              maxUnavailable: 1
+            type: RollingUpdate
           template:
             metadata:
               labels:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -17,6 +17,11 @@ spec:
     matchLabels:
       control-plane: controller-manager
   replicas: 2
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 0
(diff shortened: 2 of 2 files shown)


slintes:
probably not related to this PR, but we need to investigate test failures...

slintes:
the test is failing too often, we need to have a look at it:

```
  [FAIL] Self Node Remediation E2E Workers Remediation Without API connectivity All nodes (no API connection for all) [It] should not remediate
  /go/src/github.com/medik8s/self-node-remediation/e2e/self_node_remediation_test.go:201
```

overriding it here in order to get this PR finally merged, changes can't be related IMHO, it succeeded on 4.19 and 4.21

/override ci/prow/4.20-openshift-e2e

PR #275, opened 2025-12-01
Configurable E2E Namespace by Environment Variable

razo7:
#### Why we need this PR

- Testing the E2E code would fail when the SNR is installed in another namespace other than the hardcoded namespace `self-node-remediation` .

See below the error from testing.
```
  [FAILED] Timed out after 300.009s.
  Could not get boot time on target node
  Unexpected error:
      <*errors.StatusError | 0xc00025eaa0>: 
      namespaces "self-node-remediation" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "namespaces \"self-node-remediation\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "self-node-remediation",
                  Group: "",
                  Kind: "namespaces",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
  occurred
```
- Provide a meaningful error on missing kubeconfig
#### Changes made

- Set and fetch the operator-installed namespace by the env var `OPERATOR_NS`, and when it is missing, use the default value `self-node-remediation`.
- Use a text to describe a missing kubeconfig before running the e2e tests
## Summary by CodeRabbit

* **Tests**
  * Test suite now supports runtime selection of the test namespace via an environment variable with a configurable default, and logs which value is used.
  * Improved initialization checks and clearer error handling when acquiring cluster configuration during test setup.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/e2e/suite_test.go
+++ b/e2e/suite_test.go
@@ -70,7 +70,7 @@ var _ = BeforeSuite(func(ctx SpecContext) {
 	Expect(err).NotTo(HaveOccurred())
 
 	cfg, err = config.GetConfig()
-	Expect(err).NotTo(HaveOccurred())
+	Expect(err).NotTo(HaveOccurred(), "Couldn't get kubeconfig")
 	Expect(cfg).ToNot(BeNil())
 
 	k8sClient, err = client.New(cfg, client.Options{})


slintes (commented):
the "Why we need this PR" section does not provide a why 🤷🏼‍♂️ 🤔

razo7:
> the "Why we need this PR" section does not provide a why 🤷🏼‍♂️ 🤔

I agree, I have added the error that led to this issue in the description. 

I have kept the original value as the default value, but I am not sure it is needed.
Other e2e tests of our operators fail on missing the installed namespace (NMO/NHC), or they use a common namespace (e.g., [openshift-operators](https://github.com/medik8s/machine-deletion-remediation/blob/main/e2e/e2e_test.go#L27)).
I think failing on missing the installed namespace makes sense and is easy to fix (good signal to modify the env var before the tests begin).

PR #287, opened 2026-02-25
Remove tag trigger from post-submit workflow

razo7:
#### Why we need this PR

The **post-submit** workflow should only run on pushes to `main` branch.
Tag pushes should not trigger unversioned image builds, as the `.github/workflows/release.yml` workflow handles versioned releases properly.

Similar fix applied in https://github.com/medik8s/node-maintenance-operator/pull/156

This issue was found after the recent git tag push of the latest operator version
#### Changes made

Remove the trigger of post-submit workflow for git tags which begins with `v`.

## Summary by CodeRabbit

* **Chores**
  * Updated workflow configuration to trigger only on main branch pushes.

### DIFF
--- a/.github/workflows/post-submit.yaml
+++ b/.github/workflows/post-submit.yaml
@@ -3,8 +3,6 @@ on:
   push:
     branches:
       - main
-    tags:
-      - 'v*'
 
 jobs:
   push_to_registry:


slintes:
huh, this is old, feel free to ping me for review next time :)

obsolete now:
https://github.com/medik8s/self-node-remediation/pull/292

/close

PR #296, opened 2026-04-15
Detect host architecture for GOARCH in build script

razo7:
#### Why we need this PR

`hack/build.sh` hardcodes `GOARCH=amd64`, which prevents building the operator binary on non-x86_64 hosts (e.g. arm64, s390x, ppc64le). Anyone building on those platforms must manually override GOARCH.

Similarly, the Makefile `operator-sdk` and `opm` download targets hardcode `OS=linux && ARCH=amd64`, which downloads amd64 binaries regardless of the host architecture. This causes `Exec format error` when CI runs on a non-amd64 node (observed on NHC CI with OCP 4.20 arm64 nodes — same latent bug exists here).

#### Changes made

- `hack/build.sh`: Detect host architecture at build time using `go env GOARCH`, fall back to `amd64` if empty
- `Makefile`: Use `$(shell go env GOOS)` and `$(shell go env GOARCH)` for `operator-sdk` and `opm` downloads, matching the pattern already used by MDR, FAR, and NMO
- No behavior change on x86_64 hosts (`go env GOARCH` returns `amd64`, same as before)

Related to [RHWA-514](https://redhat.atlassian.net/browse/RHWA-514). Enables the Go binary build and tool downloads to work on non-amd64 hosts.

#### Test plan

- Existing CI runs on amd64 and produces identical behavior (no regression)
- `go env GOARCH` returns the correct architecture on all supported platforms
- `operator-sdk` and `opm` binaries match the host architecture after download

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -419,7 +419,7 @@ ifeq (,$(wildcard $(OPERATOR_SDK)))
 	set -e ;\
 	rm -rf $(OPERATOR_SDK_BIN_FOLDER) ;\
 	mkdir -p $(dir $(OPERATOR_SDK)) ;\
-	OS=linux && ARCH=amd64 && \
+	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
 	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_$${OS}_$${ARCH} ;\
 	chmod +x $(OPERATOR_SDK) ;\
 	}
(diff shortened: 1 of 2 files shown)


mshitrit (dismissed):

razo7:
Added a second commit to fix arch-hardcoded tool downloads in the Makefile.

This is the same latent bug that caused NHC CI to fail on OCP 4.20 with `Exec format error` — the `operator-sdk` and `opm` targets hardcode `OS=linux && ARCH=amd64`, downloading amd64 binaries even when CI runs on an arm64 node. The fix uses `$(shell go env GOOS)` and `$(shell go env GOARCH)` to detect the host platform, matching the pattern already used by MDR, FAR, and NMO.

mshitrit (approved):

Issue #251, opened 2025-01-28
Remediation doesn't occur when node can contact peer control plane nodes even if they consider it unhealthy


novasbc:
### Issue

In some configurations, despite a node being marked as unhealthy, a control plane node that still has access to peer control plane nodes but has lost API server access will not remediate.

### Scenario:

- 3 control plane nodes
- 3 worker nodes
- control plane node partitioned across two disks, one physical, another served by removable RAID based storage
- (not sure it matters but for full context) /var is mapped onto the removable storage as well as the root.  /etc and /usr and other critical sections are on the OS disk
    - In this case, we happen to be running rke2 services.  with the configuration on our nodes, that puts a bunch of k8s and related binaries into the removable storage such as:
        - containerd
        - crictl
        - kubectl
        - runc

### Steps to reproduce

- Provision environment with working cluster
- Configure NHC & SNR - in this case default configuration should be fine.  Our goal was to reboot the control plane node to ensure all workloads were off of it
- On one of the control plane nodes (we'll call it control1), remove or disable the removable storage
- After this occurs, a significant number of key services go offline due to removed storage.  rke2-server, a variety of pods that rely on storage
- We did find that the self-node-remediation pod stayed living in this case (have to access it by looking up the host PID for the process and getting detail from it's process info in /proc/n/*
- We observe that SNR continues to operate, and never remediates (log snippets reproduced below).

### Detail
We believe that SNR *should* have remediated, but likely there is a bug in the logic flow.  What we are seeing is that SNR is:

1. Eventually timing out on API server access
2. Getting health statuses from peer.  All indicating unhealthy
3. Deciding on "peers did not confirm that we are unhealthy, ignoring error"
    - I have no clue how it got from number 2 above to number 3 without additional logging statements being emitted

It seems like to get to number 3, it must go through [manager.go:66](https://github.com/medik8s/self-node-remediation/blob/16052be84bbfe9b1c9e791d6e05a9eae45d9655a/pkg/controlplane/manager.go#L66)  which returns true if the control plane was reachable.  But if that were the case, it would have had to go through [check.go:178](https://github.com/medik8s/self-node-remediation/blob/16052be84bbfe9b1c9e791d6e05a9eae45d9655a/pkg/apicheck/check.go#L178) which I should have seen in the log output.

[...]

PR #307, opened 2026-05-28
fix: remediate CP node when peer control planes consider it unhealthy

mshitrit:
This PR leans on the groundwork and analysis made by   @mark-dell-usa in https://github.com/medik8s/self-node-remediation/pull/255 , but is using a different solution approach. 

#### Why we need this PR
When a control-plane node loses API server access and cannot reach worker peers (UnHealthyBecauseNodeIsIsolated), SNR checks whether other CP nodes are reachable before deciding to reboot. The reachability check (canOtherControlPlanesBeReached) collapsed all peer responses — including explicit "unhealthy" verdicts — into a single boolean. A peer CP responding "this node is unhealthy" was treated identically to "I'm healthy, network is fine", preventing remediation even when every peer CP had a SNR CR for the broken node.

#### Changes made
Changes made

- internal/apicheck/check.go: Renamed canOtherControlPlanesBeReached() → getControlPlanePeersStatus(), which now returns (canBeReached bool, cpUnhealthy bool). The second value is unhealthyResponses > 0, preserving the peer CPs' health verdict instead of discarding it.
- internal/controlplane/manager.go: IsControlPlaneHealthy gains a third parameter cpUnhealthy bool and short-circuits to false (remediate) at the top of the function, before the worker-response switch, when peer CPs explicitly report this node as unhealthy.
- internal/controller/tests/controller/: unit test that cover this use case.

[RHWA-384](https://redhat.atlassian.net/browse/RHWA-384)

#### Test plan
Added regression test coverage validating remediation/watchdog behavior for control-plane node scenarios and updated test-suite setup to support it.

## Release Notes

* **Bug Fixes**
  * Improved control-plane node remediation logic to correctly handle scenarios where control-plane peers report a node as unhealthy.

* **Tests**
  * Added regression test coverage for control-plane node health check and remediation scenarios.

### DIFF
--- a/internal/apicheck/check.go
+++ b/internal/apicheck/check.go
@@ -125,10 +125,9 @@ func (c *ApiConnectivityCheck) isConsideredHealthy() bool {
 	isWorkerNode := c.controlPlaneManager == nil || !c.controlPlaneManager.IsControlPlane()
 	if isWorkerNode {
 		return workerPeersResponse.IsHealthy
-	} else {
-		return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, c.canOtherControlPlanesBeReached())
 	}
-
+	canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
+	return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, canBeReached, cpUnhealthy)
 }
 
 func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
@@ -232,19 +231,24 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 
 }
 
-func (c *ApiConnectivityCheck) canOtherControlPlanesBeReached() bool {
+// getControlPlanePeersStatus contacts peer control plane nodes and returns whether they can be reached
+// and whether they consider this node unhealthy (i.e. a SNR CR exists for it).
+func (c *ApiConnectivityCheck) getControlPlanePeersStatus() (canBeReached bool, consideredUnhealthy bool) {
 	peersToAsk := c.config.Peers.GetPeersAddresses(peers.ControlPlane)
 	numOfControlPlanePeers := len(peersToAsk)
 	if numOfControlPlanePeers == 0 {
 		c.config.Log.Info("Peers list is empty and / or couldn't be retrieved from server, other control planes can't be reached")
-		return false
+		return false, false
 	}
 
 	chosenPeersIPs := c.popPeerIPs(&peersToAsk, numOfControlPlanePeers)
 	healthyResponses, unhealthyResponses, apiErrorsResponses, _ := c.getHealthStatusFromPeers(chosenPeersIPs)
 
+	c.config.Log.Info("Control plane peers status", "healthyResponses", healthyResponses,
+		"unhealthyResponses", unhealthyResponses, "apiErrorsResponses", apiErrorsResponses)
+
 	// Any response is an indication of communication with a peer
-	return (healthyResponses + unhealthyResponses + apiErrorsResponses) > 0
+	return (healthyResponses + unhealthyResponses + apiErrorsResponses) > 0, unhealthyResponses > 0
 }
 
 func (c *ApiConnectivityCheck) popPeerIPs(peersIPs *[]corev1.PodIP, count int) []corev1.PodIP {
--- a/internal/apicheck/check.go
+++ b/internal/apicheck/check.go
@@ -127,7 +127,12 @@ func (c *ApiConnectivityCheck) isConsideredHealthy() bool {
 		return workerPeersResponse.IsHealthy
 	}
 	canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
-	return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, canBeReached, cpUnhealthy)
+	if cpUnhealthy {
+		c.config.Log.Info("Peer control plane nodes reported this node as unhealthy, triggering remediation")
+		return false
+	}
+
+	return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, canBeReached)
 }
 
 func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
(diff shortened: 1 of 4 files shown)


internal/apicheck/check.go
```diff
@@ -125,10 +125,9 @@ func (c *ApiConnectivityCheck) isConsideredHealthy() bool {
 	isWorkerNode := c.controlPlaneManager == nil || !c.controlPlaneManager.IsControlPlane()
 	if isWorkerNode {
 		return workerPeersResponse.IsHealthy
-	} else {
-		return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, c.canOtherControlPlanesBeReached())
 	}
-
+	canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
+	return c.controlPlaneManager.IsControlPlaneHealthy(workerPeersResponse, canBeReached, cpUnhealthy)
```
slintes: what's the value of passing the new `cpUnhealthy` var to `IsControlPlaneHealthy` instead of just evaluating it here?

slintes (commented):
I think this is a reasonable improvement... not sure though yet if it fixes #251 🤔

internal/apicheck/check.go
mshitrit: good point TBH I didn't think of it.
I guess in mind `IsControlPlaneHealthy` is "responsible for that particular decision and needs to have all the info to make it, and it feels like returning here kind of compromise this concept.

On the flip side the value of returning here, is simplifying IsControlPlaneHealthy and saving sending to it an unnecessary variable and acting immediately on that value , which is also quite nice.

Either way is fine by me, let me which do you prefer.

internal/apicheck/check.go
```diff
@@ -232,19 +231,24 @@ func (c *ApiConnectivityCheck) getWorkerPeersResponse() peers.Response {
 
 }
 
-func (c *ApiConnectivityCheck) canOtherControlPlanesBeReached() bool {
+// getControlPlanePeersStatus contacts peer control plane nodes and returns whether they can be reached
+// and whether they consider this node unhealthy (i.e. a SNR CR exists for it).
+func (c *ApiConnectivityCheck) getControlPlanePeersStatus() (canBeReached bool, consideredUnhealthy bool) {
```
clobrano, line 241: This new function name is unclear to me. What status is reporting? What about `controlPlanePeersReachableAndUnhealthy`? I did consider to suggest splitting the function, but it would be mostly duplication.

internal/apicheck/check.go
clobrano: I also agree on evaluate `cpUnhealthy` here. The meaning of the variable is clear enough to take decision on it. Also `IsControlPlaneHealthy` doesn't seem to take big decisions.

clobrano (changes requested):

internal/apicheck/check.go
mshitrit: TBH I prefer `Status` over `ReachableAndUnhealthy` the latter seems a bit of a mouthful, in general I'm not a huge fan of method names using "And" (i.e  `somethingAndAnotherSomething` )

> What status is reporting?

I think it's fairly intuitive - the health status of the control peers agents, but it's also explicitly explained in the method comment for good measure.

Let me know if you consider this a blocker or a Nit.

razo7 (commented):
Looks like a good fix, even though I understand the broader concern about temporary false API-server-issue reports (1/3 service endpoint routing chance).

A few additional items beyond what @slintes and @clobrano already raised:

This PR effectively supersedes https://github.com/medik8s/self-node-remediation/pull/255 from @mark-dell-usa, who spent months investigating, writing analysis, and proposing a fix. Would be good to reference it in the description and acknowledge their contribution.

The test plan section in the PR description is empty. Worth filling in with the new regression test and E2E status to catch this unusual problem.

internal/controller/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -500,6 +508,152 @@ var _ = Describe("SNR Controller", func() {
...
+			go func() { _ = cpPeers.Start(cpPeersCtx) }()
+			DeferCleanup(cancelCpPeers)
+
+			// 7. Initialise the controlplane.Manager for the unhealthy CP node and call Start
+			//    while the API is still reachable.
+			cpManager := controlplane.NewManager(shared.UnhealthyNodeName, k8sClient.Client)
+			Expect(cpManager.Start(context.Background())).To(Succeed())
+
+			// 8. Give Peers time to run its first update and populate controlPlanePeersAddresses.
+			time.Sleep(2 * time.Second)
```
razo7: NIT: Hardcoded sleep is fragile under CI load. An `Eventually` polling `cpPeers.GetPeersAddresses(peers.ControlPlane)` for a non-empty result would be more robust, if the API allows concurrent access during `Start()`.

internal/controller/tests/controller/selfnoderemediation_controller_test.go
```diff
@@ -500,6 +508,152 @@ var _ = Describe("SNR Controller", func() {
...
+
+		BeforeEach(func() {
+			// 1. Label both nodes as control-plane so the Peers selector works correctly.
+			for _, nodeName := range []string{shared.UnhealthyNodeName, shared.PeerNodeName} {
+				node := &corev1.Node{}
+				Expect(k8sClient.Client.Get(context.Background(), client.ObjectKey{Name: nodeName}, node)).To(Succeed())
+				patch := client.MergeFrom(node.DeepCopy())
+				node.Labels[commonlabels.ControlPlaneRole] = ""
+				Expect(k8sClient.Client.Patch(context.Background(), node, patch)).To(Succeed())
+				nodeName := nodeName
```
razo7: NIT: `nodeName := nodeName` is a pre-Go 1.22 loop variable capture pattern. With Go 1.25 this is a no-op and can be removed. This line copies a variable to itself, so it's dead code.

internal/apicheck/check.go
clobrano: > in general I'm not a huge fan of method names using "And"

Me too, but because it's an indicator of a function doing "too much" :)

Not a blocker though

razo7 (approved):
LGTM and holding the PR in case others have any open matters to be addressed or need more time to review
/hold

clobrano (approved):
/lgtm

Leaving a chance to get more reviews/close other threads

/hold

mshitrit:
/override ci/prow/4.23-openshift-e2e
failure unrelated to PR

PR #309, opened 2026-06-08
Set readOnlyRootFilesystem on DaemonSet agent container

razo7:
## Why we need this PR

The SNR DaemonSet agent container should have `readOnlyRootFilesystem: true` set in its securityContext. The agent performs no writes to its container filesystem, so all I/O goes through hostPath mounts, nsenter, or the Kubernetes API — so this is a safe hardening measure.

## Changes made

- Added `readOnlyRootFilesystem: true` to the container-level securityContext in
  `install/self-node-remediation-deamonset.yaml`

## Which issue(s) this PR fixes

N/A

## Test plan

- [x] Verified the agent has no container-local filesystem writes
- [ ] CI passes (unit + e2e)
- [ ] DaemonSet pods start and operate normally with the new setting

## Summary by CodeRabbit

* **Security**
  * Hardened the self-node remediation daemon by enabling `readOnlyRootFilesystem` at the container level.

* **Tests**
  * Expanded end-to-end coverage to validate daemon security settings and pod identity constraints.
  * Improved controller test cleanup and deletion verification to more reliably confirm resource removal.

### DIFF
--- a/e2e/self_node_remediation_test.go
+++ b/e2e/self_node_remediation_test.go
@@ -449,15 +449,28 @@ func ensureSnrRunning(nodes *v1.NodeList) {
 			defer wg.Done()
 			pod := findSnrPod(node)
 			utils.WaitForPodReady(k8sClient, pod)
-			verifyNoHostPort(pod)
+			verifyPodSecurity(pod)
 		}()
 	}
 	wg.Wait()
 }
 
-func verifyNoHostPort(pod *v1.Pod) {
+func verifyPodSecurity(pod *v1.Pod) {
 	ExpectWithOffset(1, pod.Spec.Containers).ToNot(BeEmpty())
-	for _, port := range pod.Spec.Containers[0].Ports {
+	container := pod.Spec.Containers[0]
+
+	for _, port := range container.Ports {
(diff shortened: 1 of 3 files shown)


weshayutin (commented):
LGTM, will the e2e be able to tell us if this broke something?  If not, we should add a test :)

razo7:
infra CI failure- re run..
/test 4.22-openshift-e2e 4.23-openshift-e2e 5.0-openshift-e2e

e2e/self_node_remediation_test.go
```diff
@@ -95,6 +95,33 @@ var _ = Describe("Self Node Remediation E2E", func() {
 			ensureSnrRunning(workerNodes)
 		})
 
+		Describe("DaemonSet Security Configuration", func() {
```
mshitrit: I think that adding 3 new e2e tests might be a big overhead.
I suggest integrating those verification in existing tests.

Maybe to ensureSnrRunning  ? (seems like a good logical choice)

mshitrit (dismissed):

mshitrit (approved):

PR #310, opened 2026-06-18
Add node-scoped caching for SNR agents

mshitrit:
#### Why we need this PR
In large clusters (1000+ nodes), each SNR agent pod caches cluster-wide resources, leading to significant memory consumption per agent. This PR reduces per-agent memory footprint by scoping the cache to only node-local resources.

#### Changes made
- Configured node-scoped caching for agent pods to watch only resources on their assigned node
- Limited cache scope to: pods on the agent's node, the agent's own node object, and namespace-scoped secrets
- Updated peers component to use API reader for cluster-wide node access, bypassing the scoped cache
- Manager mode continues using cluster-wide cache as before

[RHWA-1026](https://redhat.atlassian.net/browse/RHWA-1026)

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent mode now properly scopes controller caching and data access to node-local and deployment-namespace resources, improving efficiency and isolation.
  * Peer initialization now uses an uncached API reader so peers can reliably observe cluster-wide node data.
* **Observability**
  * Improved structured logging around node reboot timing and reboot-check decisions, including clearer status details when reboot timeouts occur.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -33,6 +33,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	pkgruntime "k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -42,6 +43,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/cache"
 	ctrl "sigs.k8s.io/controller-runtime"
+	runtimecache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -128,6 +130,64 @@ func main() {
 		setupLog.Info("HTTP/2 for metrics and webhook server enabled")
 	}
 
+	// Configure cache options based on manager vs agent mode
+	// Motivation: In large clusters (e.g., 1000+ nodes), each agent pod's cache can consume
+	// significant memory by watching and storing cluster-wide resources. Node-scoped caching
+	// reduces per-agent memory footprint by limiting the cache to only the resources needed
+	// for this specific node, preventing memory bloat as cluster size grows.
+	var cacheOpts runtimecache.Options
+	if !isManager {
+		// Agent mode: scope cache to node-local resources only
+		myNodeName := os.Getenv(nodeNameEnvVar)
+		if myNodeName == "" {
+			setupLog.Error(errors.New("failed to get own node name"), "node name was empty",
+				"env var name", nodeNameEnvVar)
+			os.Exit(1)
+		}
+
+		ns, err := utils.GetDeploymentNamespace()
+		if err != nil {
+			setupLog.Error(err, "unable to get the deployment namespace for cache scoping")
+			os.Exit(1)
+		}
+
+		setupLog.Info("Configuring node-scoped cache for agent mode", "nodeName", myNodeName, "namespace", ns)
+		cacheOpts = runtimecache.Options{
+			ByObject: map[client.Object]runtimecache.ByObject{
+				// Scope Pods to this node only
+				// Without scoping: cache would watch all pods in the cluster
+				// With scoping: cache only stores pods scheduled on this node
+				&corev1.Pod{}: {
+					Field: fields.SelectorFromSet(fields.Set{
+						"spec.nodeName": myNodeName,
+					}),
+				},
+				// Scope Nodes to this node only
+				// Note: Peers component needs all nodes, so it uses GetAPIReader() to bypass cache
+				&corev1.Node{}: {
+					Field: fields.SelectorFromSet(fields.Set{
+						"metadata.name": myNodeName,
+					}),
+				},
+				// Scope Secrets to SNR namespace only
+				// Prevents caching secrets from all namespaces cluster-wide
+				&corev1.Secret{}: {
+					Namespaces: map[string]runtimecache.Config{
+						ns: {},
+					},
+				},
+				// VolumeAttachment: No cache scoping available
+				// - API server doesn't support spec.nodeName field selector (only metadata.name/namespace)
+			},
+		}
+	}
+	// Manager mode uses default (cluster-wide) cache
+
+	// Note: In agent mode with scoped cache, reads of out-of-scope resources
+	// (e.g., Pods on other nodes) will return NotFound from the cache.
+	// This is intentional - agents should only need node-local resources.
+	// If out-of-scope reads are needed, use mgr.GetAPIReader() for direct API access.
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme: scheme,
 		Metrics: server.Options{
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -128,6 +130,64 @@ func main() {
...
+				// Prevents caching secrets from all namespaces cluster-wide
+				&corev1.Secret{}: {
+					Namespaces: map[string]runtimecache.Config{
+						ns: {},
+					},
+				},
+				// VolumeAttachment: No cache scoping available
+				// - API server doesn't support spec.nodeName field selector (only metadata.name/namespace)
+			},
+		}
```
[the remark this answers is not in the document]
mshitrit: This concern is not valid. The cache scoping only applies to agent mode (`if !isManager` at line 139), not manager mode.

  Manager mode uses the default cluster-wide cache (line 184 comment), so when `isResourceDeletionCompleted` is called from the manager's reconcile path, `r.Client.List(pods)` returns all pods cluster-wide, not just pods on the manager's own node.

  The analysis missed that the cache scoping is conditional - agents get node-scoped caches (which is correct since they only remediate their own node), while managers retain cluster-wide visibility.

mshitrit:
verified e2e tests on local cluster, and did an initial analysis which doesn't show failure is related to this PR 
Giving it another go before a deeper analysis
/test 4.23-openshift-e2e

slintes:
4.23 was never green yet :/ 
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-medik8s-self-node-remediation-main-4.23-openshift-e2e

I'd suggest to use an earlier version for this PR and investigate the 4.23 issue separately...

mshitrit:
> I'd suggest to use an earlier version for this PR and investigate the 4.23 issue separately...

That's the only version that was running the prow e2e (which never was green).

I tested with an earlier version e2e test on my local aws cluster successfully. 
I don't think though that it passed because of different content (since diff was very minor), my bet is on something difference between the CI cluster and my own.

I will do more tests and update.

mshitrit:
1. Verified image e2e tests pass on my local aws cluster.
2. Couldn't find a RCA (pointing to failure been unrelated to PR changes which does not make sense IMO).
3. added some more logs for potential better diagnostics

/test 4.22-openshift-e2e

mshitrit:
test passed successfully without any changes (expect additional logs).
looks like analysis was correct and issue indeed was unrelated to the PR.
/test 4.22-openshift-e2e

cmd/main.go
```diff
@@ -309,7 +370,13 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
...
 	peerApiServerTimeout := getDurEnvVarOrDie("PEER_API_SERVER_TIMEOUT")
 
-	myPeers := peers.New(myNodeName, peerUpdateInterval, mgr.GetClient(), ctrl.Log.WithName("peers"), peerApiServerTimeout)
+	// Use GetAPIReader() instead of GetClient() for peers component
+	// Reason: The agent's cache is node-scoped (only caches this node's data), but the peers
+	// component needs to read ALL nodes in the cluster to monitor peer health. GetAPIReader()
+	// bypasses the scoped cache and reads directly from the API server, allowing peers to access
+	// cluster-wide node data. This is an intentional trade-off: accept some direct API calls for
+	// peers rather than caching all nodes on every agent pod (which would defeat the memory savings).
+	myPeers := peers.New(myNodeName, peerUpdateInterval, mgr.GetAPIReader(), ctrl.Log.WithName("peers"), peerApiServerTimeout)
```
razo7, line 379: With `GetAPIReader()`, each agent makes 5 direct (uncached) API calls per update cycle — 1 Get + 2 List for worker peers + 2 List for control-plane peers.

Steady state (15min interval): 1000 agents × 5 calls / 900s ≈ 5.5 req/s — fine.

Error retry (2min interval, `quickPeerUpdateInterval`): 1000 agents × 5 calls / 120s ≈ 42 req/s of uncached calls, precisely when the API server is already stressed. This creates a feedback loop.

Consider adding jitter to `quickPeerUpdateInterval` in a follow-up, or exponential backoff on repeated failures. Not a blocker for this PR since the peers polling logic is pre-existing.

cmd/main.go
```diff
@@ -128,6 +130,64 @@ func main() {
...
+					}),
+				},
+				// Scope Secrets to SNR namespace only
+				// Prevents caching secrets from all namespaces cluster-wide
+				&corev1.Secret{}: {
+					Namespaces: map[string]runtimecache.Config{
+						ns: {},
+					},
+				},
+				// VolumeAttachment: No cache scoping available
```
razo7, line 179: Since `isResourceDeletionCompleted()` (which reads VolumeAttachments) only runs in manager mode and manager mode retains the cluster-wide cache, then this is gentle. Agent pods will cache VolumeAttachments they never read, which is wasted memory but not a correctness issue.

Future optimization: consider disabling the VolumeAttachment informer in agent mode entirely, or using `mgr.GetAPIReader()` for VolumeAttachment access in any agent-mode code paths.

cmd/main.go
```diff
@@ -128,6 +130,64 @@ func main() {
...
 
+	// Configure cache options based on manager vs agent mode
+	// Motivation: In large clusters (e.g., 1000+ nodes), each agent pod's cache can consume
+	// significant memory by watching and storing cluster-wide resources. Node-scoped caching
+	// reduces per-agent memory footprint by limiting the cache to only the resources needed
+	// for this specific node, preventing memory bloat as cluster size grows.
+	var cacheOpts runtimecache.Options
+	if !isManager {
+		// Agent mode: scope cache to node-local resources only
+		myNodeName := os.Getenv(nodeNameEnvVar)
```
razo7, line 141: nit: `myNodeName` and deployment namespace are retrieved again at lines 335-346 in `initSelfNodeRemediationAgent()` with identical logic. Could extract once at a higher scope and pass as parameters.

razo7 (approved):
I like the implementation, and I left small NITs
That can be done now or in a follow up commit, so I am holding
/hold

cmd/main.go
mshitrit: >or using mgr.GetAPIReader() for VolumeAttachment access in any agent-mode code paths.

I don't think that would work.
a. It does not prevent the core issue of VolumeAttachment still being cached
b. In any case per your previous statement VolumeAttachment aren't accessed in agent mode so nowhere to apply this

>consider disabling the VolumeAttachment informer in agent mode entirely

That may work, it would look something like this:


```
 cacheOpts = runtimecache.Options{
      ByObject: map[client.Object]runtimecache.ByObject{
          &corev1.Pod{}: {
              Field: fields.SelectorFromSet(fields.Set{
                  "spec.nodeName": myNodeName,
              }),
          },
          &corev1.Node{}: {
              Field: fields.SelectorFromSet(fields.Set{
                  "metadata.name": myNodeName,
              }),
          },
          &corev1.Secret{}: {
              Namespaces: map[string]runtimecache.Config{
                  ns: {},
              },
          },
      },
      DisableFor: []client.Object{
          &storagev1.VolumeAttachment{},
      },
  }
```

I'm not keen to apply it yet though, as it'll block any usage or registration of VolumeAttachment in the node and I'm not sure yet regarding the saving value.

cmd/main.go
mshitrit: Duplicated part is in a different method.
IMO "messing" the functions with receiving additional params overweight the value of saving that minor duplication in this particular case.

cmd/main.go
mshitrit: Nice idea.
Not sure if it's important enough to create a ticket for or something we'd revisit in case the need arise.
Let me know if you have a preference

razo7: > Not sure if it's important enough to create a ticket for or something we'd revisit in case the need arise.

I do see value in creating the ticket with a low effort to track it.  Can you create a follow-up ticket to add jitter/backoff to the peer update interval to prevent the API server from becoming a thundering herd under stress?

mshitrit: Done :+1: 
https://redhat.atlassian.net/browse/RHWA-1306

PR #321, opened 2026-07-16
Add regression test for deployment rollout strategy

mpryc:
## Summary
- Adds a unit test that validates `config/manager/manager.yaml` has `maxSurge: 0` and `maxUnavailable: 1` on the controller-manager deployment
- Prevents regression of the upgrade deadlock fixed in PR #273

## Context
Validation for [RHWA-1041](https://issues.redhat.com/browse/RHWA-1041): SNR v0.11.0 introduced `topologySpreadConstraints` with `DoNotSchedule` and scaled to 2 replicas, but used the default `maxSurge` (25%). On 2-worker clusters, this caused OLM upgrade deadlocks because the new pod couldn't satisfy the spread constraint while old pods occupied all scheduling slots.

## Test plan
- [x] Test passes on current main (`maxSurge: 0`)
- [x] Test fails when `maxSurge` is changed to 1 (reproduces the regression)
- [x] Runs as part of `make test` (covers `./internal/...`)
- [x] No new dependencies added

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify the manager deployment uses a rolling update strategy.
  * Confirmed deployment updates allow no additional surge capacity while keeping one instance unavailable.

### DIFF
--- a/internal/deployment/deployment_test.go
+++ b/internal/deployment/deployment_test.go
@@ -1,0 +1,58 @@
+package deployment
+
+import (
+	"io"
+	"os"
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+const managerYAMLPath = "../../config/manager/manager.yaml"
+
+func parseManagerDeployment(t *testing.T) *appsv1.Deployment {
+	t.Helper()
+	f, err := os.Open(managerYAMLPath)
+	if err != nil {
+		t.Fatalf("failed to open manager.yaml: %v", err)
+	}
+	defer f.Close()
+
+	decoder := yaml.NewYAMLOrJSONDecoder(f, 4096)
+	for {
+		obj := &appsv1.Deployment{}
+		if err := decoder.Decode(obj); err != nil {
+			if err == io.EOF {
+				break
+			}
+			t.Fatalf("failed to decode document from manager.yaml: %v", err)
+		}
+		if obj.Kind == "Deployment" {
+			return obj
+		}
+	}
+	t.Fatal("no Deployment found in manager.yaml")
+	return nil
+}
+
+// TestManagerRolloutStrategyPreventsUpgradeDeadlock validates that the manager deployment
+// uses maxSurge=0 and maxUnavailable=1.
+//
+// With topologySpreadConstraints (DoNotSchedule) and 2 replicas, the default maxSurge
+// causes a 3rd pod to be created during rolling updates. On 2-worker clusters, this pod
+// cannot satisfy the spread constraint and remains Pending forever, blocking OLM upgrades.
+func TestManagerRolloutStrategyPreventsUpgradeDeadlock(t *testing.T) {
+	g := NewGomegaWithT(t)
+	dep := parseManagerDeployment(t)
+
+	g.Expect(dep.Spec.Strategy.Type).To(Equal(appsv1.RollingUpdateDeploymentStrategyType))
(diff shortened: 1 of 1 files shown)


internal/deployment/deployment_test.go
```diff
@@ -0,0 +1,58 @@
...
+// causes a 3rd pod to be created during rolling updates. On 2-worker clusters, this pod
+// cannot satisfy the spread constraint and remains Pending forever, blocking OLM upgrades.
+func TestManagerRolloutStrategyPreventsUpgradeDeadlock(t *testing.T) {
+	g := NewGomegaWithT(t)
+	dep := parseManagerDeployment(t)
+
+	g.Expect(dep.Spec.Strategy.Type).To(Equal(appsv1.RollingUpdateDeploymentStrategyType))
+	g.Expect(dep.Spec.Strategy.RollingUpdate).NotTo(BeNil())
+	g.Expect(*dep.Spec.Strategy.RollingUpdate.MaxSurge).To(Equal(intstr.FromInt(0)),
+		"maxSurge must be 0 to prevent upgrade deadlock on 2-worker clusters")
```
razo7, line 58: Nit: add nil guards before dereferencing `MaxSurge` and `MaxUnavailable` to get a clear assertion failure instead of a panic if someone removes these fields from the YAML:

```go
g.Expect(dep.Spec.Strategy.RollingUpdate.MaxSurge).NotTo(BeNil(), "maxSurge must be explicitly set")
g.Expect(*dep.Spec.Strategy.RollingUpdate.MaxSurge).To(Equal(intstr.FromInt32(0)),
    "maxSurge must be 0 to prevent upgrade deadlock on 2-worker clusters")
g.Expect(dep.Spec.Strategy.RollingUpdate.MaxUnavailable).NotTo(BeNil(), "maxUnavailable must be explicitly set")
g.Expect(*dep.Spec.Strategy.RollingUpdate.MaxUnavailable).To(Equal(intstr.FromInt32(1)),
    "maxUnavailable must be 1 to allow progress by terminating old pods first")
```

Also: `intstr.FromInt` is deprecated since apimachinery v0.28 — use `intstr.FromInt32`.

internal/deployment/deployment_test.go
```diff
@@ -0,0 +1,58 @@
...
+	g.Expect(*dep.Spec.Strategy.RollingUpdate.MaxUnavailable).To(Equal(intstr.FromInt(1)),
+		"maxUnavailable must be 1 to allow progress by terminating old pods first")
+}
```
razo7, line 62: The `parseManagerDeployment` helper is nice and can be reused. I would suggest extending its coverage to other fields that can regress silently:

**Security** (regression = pods fail to schedule on OCP / certification failure):
- `runAsNonRoot: true`
- `seccompProfile: RuntimeDefault`
- `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`
- `capabilities.drop: ALL`

**Operational** (tightly coupled to the rollout strategy this test guards):
- `replicas: 2` — if changed to 1, the rollout test still passes but HA is gone
- `priorityClassName: system-cluster-critical` — prevents eviction during node pressure
- `topologySpreadConstraints` — `DoNotSchedule` + `maxSkew: 1`

This pattern can be portable to FAR and NHC — they share the same deployment structure.

razo7 (commented):
I like the helper and I think we can extend it to avoid future regerssion. Other than that I left small NITs

internal/deployment/deployment_test.go
mpryc: Should be addressed.

razo7: Nice two functions (`TestManagerDeploymentSecurityContext`, `TestManagerDeploymentHAAndScheduling`), but I didn't see a call for these functions actually to test them :)

PR #325, opened 2026-07-24
Fix CP node isolation detection via peer reachability gate

mpryc:
## Summary

Fixes [RHWA-1384](https://redhat.atlassian.net/browse/RHWA-1384) — SNR agent cannot remediate isolated control plane nodes.

On CP nodes, the ClusterIP request can be routed to the local kube-apiserver endpoint. The `/readyz` checks do not provide a signal that detects loss of connectivity to the rest of the cluster, so the error count never increments, peer consultation is never triggered, and the watchdog is fed indefinitely.

### Changes

- **Peer reachability gate on CP nodes**: after readyz passes on a CP node, verify that at least one peer is reachable via lightweight TCP dial against a random sample of up to 3 peers. If no peer is reachable, treat the check as failed and enter the existing `isConsideredHealthy()` error threshold / peer consultation flow. Worker nodes are unaffected.
- **`PeerAddressProvider` interface**: changed `ApiConnectivityCheckConfig.Peers` from `*peers.Peers` to an interface. `*peers.Peers` satisfies it without modification — no changes needed in `cmd/main.go` or any consumer.
- **`PeerDialer` interface**: abstracts TCP dialing for testability.

### What does NOT change

- `/readyz?exclude=shutdown` endpoint and how it is called
- Worker node behavior
- `getWorkerPeersResponse()`, `getControlPlanePeersStatus()`, `IsControlPlaneHealthy()`, `isDiagnosticsPassed()`
- Watchdog feeding/stopping mechanics
- All existing configuration (no new CRD fields)
- `SafeTimeToAssumeNodeRebootedSeconds` calculation
- The OLM bundle

### Design

See `docs/design/cp-isolation-detection.md` for the full design document including root cause analysis, considered alternatives, edge cases, and time-to-remediation analysis.

## Test plan

- [x] 30 unit tests covering: `canReachAnyPeer`, `getAllPeerAddresses`, `isControlPlane`, full flow (CP healthy, CP isolated, worker unchanged, transient recovery), 2+1 arbiter cluster scenarios, compact cluster deduplication, interface compatibility, thread safety
- [x] All tests pass with `-race` flag
- [x] E2E: deployed on a 3x3 OCP cluster, isolated a CP node via EC2 security groups, verified remediation triggered (node rebooted, reached Fencing-Completed, recovered after isolation reverted) — [test notes](https://hackmd.io/@medik8s/SkB23VbBMg) by @weshayutin

### DIFF
--- a/docs/design/cp-isolation-detection.md
+++ b/docs/design/cp-isolation-detection.md
@@ -1,0 +1,125 @@
+# Fix: SNR Cannot Remediate Isolated Control Plane Nodes
+
+## The Problem
+
+SNR's health check loop calls `/readyz?exclude=shutdown` through the Kubernetes
+ClusterIP service to determine whether the API server is reachable. On worker
+nodes this works correctly — when a worker is network-isolated, the request
+fails, the error count climbs, peers are consulted, and remediation proceeds.
+
+On control plane nodes it does not work. When a CP node loses network
+connectivity, kube-proxy routes the ClusterIP request to the only API server
+it can still reach: the local one running on the same node. That local API
+server returns HTTP 200 because the `/readyz` sub-checks (informer sync,
+post-start hooks) do not verify etcd quorum. The result: `errorCount` never
+increments, peer consultation is never triggered, the watchdog is fed
+indefinitely, and the node never reboots.
+
+Switching to `/healthz` does not help. The etcd healthz check tests local
+etcd connectivity, not quorum — it would still pass on an isolated node.
+
+### Why This Matters
+
+In a 2+1 arbiter topology (or any cluster with 3+ CP nodes), a
+network-isolated CP node holds its etcd member hostage. The remaining CP nodes
+cannot form quorum until the isolated node's etcd member is removed or the node
+reboots. SNR exists precisely to handle this — but the bug prevents it from
+ever firing.
+
+## Root Cause
+
+The health check has a single signal: "can I reach the API server?" On CP
+nodes this signal is tautologically true because the API server is local.
+There is no second signal to distinguish "healthy and connected" from
+"healthy but isolated."
+
+## The Fix
+
+Add a second signal: **peer reachability**. On a CP node, after readyz
+returns 200, verify that at least one peer node is reachable via TCP. If no
+peer is reachable, treat the check as failed and enter the existing error
+threshold / peer consultation flow.
+
+This is the minimal change that breaks the tautology. It requires no new CRD
+fields, no configuration changes, and no modifications to worker node behavior.
+
+### What Changes
+
+**1. Peer reachability gate on CP nodes** (`Start()` in `internal/apicheck/check.go`)
+
+After readyz returns 200 on a CP node, call `canReachAnyPeer()`:
+- If any peer responds to a TCP dial → the node has network connectivity →
+  reset error count (existing behavior).
+- If no peer responds → treat this cycle as a failure → enter
+  `isConsideredHealthy()` (existing error/peer consultation flow).
+
+Worker nodes skip this check entirely — their behavior is unchanged.
+
+`canReachAnyPeer()` uses a lightweight TCP dial (not full gRPC) against up to
+3 peers from the combined worker + CP peer list. TCP is sufficient because we
+are testing "is the network up?", not "is the peer application healthy?" The
+full gRPC peer health check still runs later inside `isConsideredHealthy()`
+when the error threshold is reached.
+
+**2. CP isolation escalation** (`isConsideredHealthy()`)
+
+A secondary issue: when readyz *does* fail on a CP node (e.g., iptables DROP
+rather than NIC down), the existing code path enters `isConsideredHealthy()`
+but the CP peer isolation signal can be discarded if the worker error threshold
+has not yet been reached. This delays remediation unnecessarily.
+
+A new counter (`cpUnreachableCount`) tracks consecutive CP peer unreachability
+independently of the worker threshold. When it reaches `MaxErrorsThreshold`,
+the code bypasses the worker threshold and feeds an isolation signal directly
+to `IsControlPlaneHealthy()`.
+
+### What Does Not Change
+
+- The `/readyz?exclude=shutdown` endpoint and how it is called
+- Worker node behavior (no code path change for workers)
+- `getWorkerPeersResponse()`, `getControlPlanePeersStatus()`,
+  `IsControlPlaneHealthy()`, `isDiagnosticsPassed()`
+- Watchdog feeding/stopping mechanics
+- All existing configuration (no new CRD fields, no new env vars)
+- `SafeTimeToAssumeNodeRebootedSeconds` calculation
+- The OLM bundle
+
+### Upgrade Path
+
+- `ApiConnectivityCheckConfig.Peers` changes from `*peers.Peers` to a
+  `PeerAddressProvider` interface. `*peers.Peers` satisfies this interface
+  without modification — no changes needed in `cmd/main.go` or any consumer.
+- On upgrade, CP nodes gain the peer reachability check. A healthy,
+  well-connected CP node sees no behavioral difference because
+  `canReachAnyPeer()` succeeds immediately.
+
+## Time to Remediation
+
+With defaults (CheckInterval=15s, MaxErrorsThreshold=3, ApiServerTimeout=5s,
+peerDialTimeout=2s, MaxTimeForNoPeersResponse=30s):
+
+```
+3 × (15s + 5s + ~2s) + 30s + watchdog ≈ 96s + watchdog timeout
+```
+
+This is comparable to the existing worker-node remediation timeline.
+
+## Edge Cases
+
+| Scenario | Outcome |
+|----------|---------|
+| Single CP node, no peers | `canReachAnyPeer()` → true (no peers = cannot determine isolation), error count resets |
+| Brief network blip (1–2 cycles) | Error count stays below threshold, recovers next cycle |
+| Partial isolation (some peers reachable) | `canReachAnyPeer()` → true, error count resets |
+| Full isolation, NIC down | `canReachAnyPeer()` fails repeatedly → remediation triggers |
+| Full isolation, iptables DROP | readyz fails + `cpUnreachableCount` escalates → remediation triggers |
+| Compact 3-node cluster | Peers deduplicated across worker/CP lists, works correctly |
+
+## Out of Scope
+
+PR #255 identified a related issue: when peers are consulted, kube-proxy may
+route their API server connections back through the failing CP node (1/N
+probability), causing false "cluster-wide API issue" conclusions. This is
+pre-existing and orthogonal — with the fixes above, the isolated CP node no
+longer depends solely on the worker peer consultation flow. Recommend filing
+a separate issue for the kube-proxy routing concern.
--- a/docs/design/cp-isolation-detection.md
+++ b/docs/design/cp-isolation-detection.md
@@ -8,30 +8,31 @@ nodes this works correctly — when a worker is network-isolated, the request
 fails, the error count climbs, peers are consulted, and remediation proceeds.
 
 On control plane nodes it does not work. When a CP node loses network
-connectivity, kube-proxy routes the ClusterIP request to the only API server
-it can still reach: the local one running on the same node. That local API
-server returns HTTP 200 because the `/readyz` sub-checks (informer sync,
-post-start hooks) do not verify etcd quorum. The result: `errorCount` never
-increments, peer consultation is never triggered, the watchdog is fed
-indefinitely, and the node never reboots.
-
-Switching to `/healthz` does not help. The etcd healthz check tests local
-etcd connectivity, not quorum — it would still pass on an isolated node.
+connectivity, the ClusterIP request can be routed to the local kube-apiserver
+endpoint running on that same node. In this failure mode, the local API server
+continues responding even though the node is isolated from the rest of the
+cluster. That local API server returns HTTP 200 because the `/readyz` checks
+do not provide a signal that detects loss of connectivity to the rest of the
+cluster. The result: `errorCount` never increments, peer consultation is
+never triggered, the watchdog is fed indefinitely, and the node never reboots.
 
 ### Why This Matters
 
-In a 2+1 arbiter topology (or any cluster with 3+ CP nodes), a
-network-isolated CP node holds its etcd member hostage. The remaining CP nodes
-cannot form quorum until the isolated node's etcd member is removed or the node
-reboots. SNR exists precisely to handle this — but the bug prevents it from
-ever firing.
+In a 2+1 arbiter topology (or standard 3-CP clusters), when a CP node is
+network-isolated, the remaining nodes maintain etcd quorum and the cluster
+continues operating in degraded mode. However, workloads on the isolated node
+keep running with stale state and are not rescheduled, because SNR never
+detects the isolation and never fences the node.
 
 ## Root Cause
 
-The health check has a single signal: "can I reach the API server?" On CP
-nodes this signal is tautologically true because the API server is local.
-There is no second signal to distinguish "healthy and connected" from
-"healthy but isolated."
+SNR calls the kube-apiserver `/readyz` endpoint through the ClusterIP service.
+In the failure mode observed here, the local API server handles the request.
+The `/readyz` endpoint runs multiple sub-checks (informer sync, post-start
+hooks, etc.) but does not provide a signal that detects loss of connectivity
+to the rest of the cluster. The relevant `/readyz` sub-checks pass even on an
+isolated node, so SNR has no signal to distinguish "healthy and connected"
+from "healthy but isolated."
 
 ## The Fix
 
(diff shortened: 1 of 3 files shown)


docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+connectivity, kube-proxy routes the ClusterIP request to the only API server
+it can still reach: the local one running on the same node. That local API
+server returns HTTP 200 because the `/readyz` sub-checks (informer sync,
+post-start hooks) do not verify etcd quorum. The result: `errorCount` never
+increments, peer consultation is never triggered, the watchdog is fed
+indefinitely, and the node never reboots.
+
+Switching to `/healthz` does not help. The etcd healthz check tests local
+etcd connectivity, not quorum — it would still pass on an isolated node.
+
```
DanielFroehlich, line 18: That is not correct. See test result in this [comment](https://redhat.atlassian.net/browse/RHWA-1384?focusedCommentId=17702302), showin health endpoint failing on isolated node with etcd check.

docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+### Why This Matters
+
+In a 2+1 arbiter topology (or any cluster with 3+ CP nodes), a
+network-isolated CP node holds its etcd member hostage. The remaining CP nodes
+cannot form quorum until the isolated node's etcd member is removed or the node
+reboots. SNR exists precisely to handle this — but the bug prevents it from
```
DanielFroehlich: That is not correct. The remaining CP node (or in 2+1 arbiter the remaining CP node plus the arbiter) do have quorum and continue to work in degraded but operative mode. There is no hostage situation. 
Which means the controlplane is active, workload is scheduled and restarted, although it is still running on the failed/isolated node.

docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+ever firing.
+
+## Root Cause
+
+The health check has a single signal: "can I reach the API server?" On CP
+nodes this signal is tautologically true because the API server is local.
+There is no second signal to distinguish "healthy and connected" from
+"healthy but isolated."
+
```
DanielFroehlich, line 36: That statement is not correct. the health check does a lot more checks then just "can I reach the API server" - see all the checks done with "?verbose" output. The root cause is that the ready endpoint does not include the ectd check.

docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+## The Fix
+
+Add a second signal: **peer reachability**. On a CP node, after readyz
+returns 200, verify that at least one peer node is reachable via TCP. If no
+peer is reachable, treat the check as failed and enter the existing error
+threshold / peer consultation flow.
+
+This is the minimal change that breaks the tautology. It requires no new CRD
+fields, no configuration changes, and no modifications to worker node behavior.
+
```
DanielFroehlich, line 47: The bug has a bunch of other options on how to fix those. I would love to see them discussed with pros and cons. At least somerwhere below in "rejected alterantives" that those options have been considered, and why they have been rejected.

docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+3 × (15s + 5s + ~2s) + 30s + watchdog ≈ 96s + watchdog timeout
+```
+
+This is comparable to the existing worker-node remediation timeline.
+
+## Edge Cases
+
+| Scenario | Outcome |
+|----------|---------|
+| Single CP node, no peers | `canReachAnyPeer()` → true (no peers = cannot determine isolation), error count resets |
```
DanielFroehlich: this includes single CP but multiple worker node topologies i guess? Would love to see that stated explicitly, to make it clear that case has been considered.

docs/design/cp-isolation-detection.md
```diff
@@ -0,0 +1,125 @@
...
+when the error threshold is reached.
+
+**2. CP isolation escalation** (`isConsideredHealthy()`)
+
+A secondary issue: when readyz *does* fail on a CP node (e.g., iptables DROP
+rather than NIC down), the existing code path enters `isConsideredHealthy()`
+but the CP peer isolation signal can be discarded if the worker error threshold
+has not yet been reached. This delays remediation unnecessarily.
+
+A new counter (`cpUnreachableCount`) tracks consecutive CP peer unreachability
```
slintes: I don't understand how this change improves anything. You replace one error treshold with another one, not? IMHO this adds complexity with low value if any.

mpryc: You are right. I'll drop second counter. However @DanielFroehlich  shared test data showing /healthz already catches isolation via the etcd sub-check. If /livez does too, the fix might be as simple as switching the endpoint. In that case it would be almost a one-liner. wdyt ?

slintes: see https://redhat.atlassian.net/browse/RHWA-1384?focusedCommentId=17702903

slintes: My initial fav fix would be to not use the cluster ip for the API check, but the node IP of another CP node. 
But even that needs further considerations... what if the other node is down, how does peer check behave with that change, .... it's too easy to break things.

Always keep in mind why we do what we do:

- we want to reboot when a SNR CR is created
- that only works when we have a working API server connection
- that's why we introduced the API check: ensure we can watch for SNR CRs
- what if that connection is broken:
  - is it a problem with the API server? If so, nothing we can do
  - is it a problem with our own node? Reboot, the manager pod cares for everything else
  - in order to find out which of these 2 apply, we ask our peers

I think your proposed fix has the least chance for unwanted side effects...

internal/apicheck/check.go
```diff
@@ -110,14 +139,98 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
+			combined = append(combined, p)
+		}
+	}
+	for _, p := range cpPeers {
+		if _, exists := seen[p.IP]; !exists && p.IP != "" {
+			seen[p.IP] = struct{}{}
+			combined = append(combined, p)
+		}
+	}
+	return combined
```
slintes, line 234: you might want to shuffle peers... a deterministic first N selection could false-positive on larger clusters if first 3 peers are down

docs/design/cp-isolation-detection.md
DanielFroehlich: ```we want to reboot when a SNR CR is created``
That is a design flaw: On a network isolated node, there is no way to detect the SNR CR being created.
SNR can not depend on a failed node to detect that CR to be created.

There is also no way to contact peers. That is if API is detected faulty (e.g. via health endpoint), a peer check should also fail on the isolated node (connection timeout). If the peer check does not fail if it cant connect to any peer, that would also be a bug / design flaw.

weshayutin:
OK..  I have successfully recreated the issue AFAIK and to the best of my current ability.   Using ec2 security groups I completely isolated the networking of one the master nodes in a typical 3x3 cluster.   As documented in the bug that @DanielFroehlich opened, SelfNodeRemediation instance was created and no action is taken.. that's a bug imho.  Notes are here: https://hackmd.io/@medik8s/HyQdnN-Bzg    

Now today I pulled in @mpryc's fix and deployed it to the same cluster.  and ran the same scenario after ensuring all nodes were running etc..  This time a selfnoderemediation instance was created, the node rebooted and ended up in "Fencing-Complete"  .. I am not an expert but this seems right.. 

Then I reverted the change to the ec2 security groups and the node went back to "Ready" and the selfnoderemediation instance was deleted.   These notes are here: https://hackmd.io/@medik8s/SkB23VbBMg

The cr transitions can be seen here: 

```
oc get selfnoderemediations 
NAME                                              AGE
ip-10-0-76-226.us-west-2.compute.internal-6z94q   2m
whayutin@fedora:~/OPENSHIFT/git/MEDIK8S/node-healthcheck-operator$ oc get selfnoderemediations ip-10-0-76-226.us-west-2.compute.internal-6z94q -o yaml
apiVersion: self-node-remediation.medik8s.io/v1alpha1
kind: SelfNodeRemediation
metadata:
  annotations:
    remediation.medik8s.io/node-name: ip-10-0-76-226.us-west-2.compute.internal
    remediation.medik8s.io/template-name: self-node-remediation-automatic-strategy-template
  creationTimestamp: "2026-07-24T19:14:05Z"
  finalizers:
  - self-node-remediation.medik8s.io/snr-finalizer
  generateName: ip-10-0-76-226.us-west-2.compute.internal-
  generation: 1
  labels:
    app.kubernetes.io/part-of: node-healthcheck-controller
    remediation.medik8s.io/isControlPlaneNode: ""
  name: ip-10-0-76-226.us-west-2.compute.internal-6z94q
  namespace: openshift-workload-availability
  ownerReferences:
  - apiVersion: remediation.medik8s.io/v1alpha1
    controller: false
    kind: NodeHealthCheck
    name: nhc-control-plane
    uid: ba63397c-fb4b-4c1c-90a5-84234275180e
  resourceVersion: "96809"
  uid: 78cb0ce3-65b5-4c3d-8465-58338084796a
spec:
  remediationStrategy: Automatic
status:
  conditions:
  - lastTransitionTime: "2026-07-24T19:14:05Z"
    message: ""
    reason: RemediationStarted
    status: "True"
    type: Processing
  - lastTransitionTime: "2026-07-24T19:14:05Z"
    message: ""
    reason: RemediationStarted
    status: Unknown
    type: Succeeded
  phase: Reboot-Completed
  timeAssumedRebooted: "2026-07-24T19:16:05Z"
whayutin@fedora:~/OPENSHIFT/git/MEDIK8S/node-healthcheck-operator$ oc get selfnoderemediations ip-10-0-76-226.us-west-2.compute.internal-6z94q -o yaml
apiVersion: self-node-remediation.medik8s.io/v1alpha1
kind: SelfNodeRemediation
metadata:
  annotations:
    remediation.medik8s.io/node-name: ip-10-0-76-226.us-west-2.compute.internal
    remediation.medik8s.io/template-name: self-node-remediation-automatic-strategy-template
  creationTimestamp: "2026-07-24T19:14:05Z"
  finalizers:
  - self-node-remediation.medik8s.io/snr-finalizer
  generateName: ip-10-0-76-226.us-west-2.compute.internal-
  generation: 1
  labels:
    app.kubernetes.io/part-of: node-healthcheck-controller
    remediation.medik8s.io/isControlPlaneNode: ""
  name: ip-10-0-76-226.us-west-2.compute.internal-6z94q
  namespace: openshift-workload-availability
  ownerReferences:
  - apiVersion: remediation.medik8s.io/v1alpha1
    controller: false
    kind: NodeHealthCheck
    name: nhc-control-plane
    uid: ba63397c-fb4b-4c1c-90a5-84234275180e
  resourceVersion: "97147"
  uid: 78cb0ce3-65b5-4c3d-8465-58338084796a
spec:
  remediationStrategy: Automatic
status:
  conditions:
  - lastTransitionTime: "2026-07-24T19:16:26Z"
    message: ""
    reason: RemediationFinishedSuccessfully
    status: "False"
    type: Processing
  - lastTransitionTime: "2026-07-24T19:16:26Z"
    message: ""
    reason: RemediationFinishedSuccessfully
    status: "True"
    type: Succeeded
  phase: Fencing-Completed
  timeAssumedRebooted: "2026-07-24T19:16:05Z"
whayutin@fedora:~/OPENSHIFT/git/MEDIK8S/node-healthcheck-operator$ oc get selfnoderemediations ip-10-0-76-226.us-west-2.compute.internal-6z94q -o yaml
apiVersion: self-node-remediation.medik8s.io/v1alpha1
kind: SelfNodeRemediation
metadata:
  annotations:
    remediation.medik8s.io/node-name: ip-10-0-76-226.us-west-2.compute.internal
    remediation.medik8s.io/template-name: self-node-remediation-automatic-strategy-template
  creationTimestamp: "2026-07-24T19:14:05Z"
  finalizers:
  - self-node-remediation.medik8s.io/snr-finalizer
  generateName: ip-10-0-76-226.us-west-2.compute.internal-
  generation: 1
  labels:
    app.kubernetes.io/part-of: node-healthcheck-controller
    remediation.medik8s.io/isControlPlaneNode: ""
  name: ip-10-0-76-226.us-west-2.compute.internal-6z94q
  namespace: openshift-workload-availability
  ownerReferences:
  - apiVersion: remediation.medik8s.io/v1alpha1
    controller: false
    kind: NodeHealthCheck
    name: nhc-control-plane
    uid: ba63397c-fb4b-4c1c-90a5-84234275180e
  resourceVersion: "97147"
  uid: 78cb0ce3-65b5-4c3d-8465-58338084796a
spec:
  remediationStrategy: Automatic
status:
  conditions:
  - lastTransitionTime: "2026-07-24T19:16:26Z"
    message: ""
    reason: RemediationFinishedSuccessfully
    status: "False"
    type: Processing
  - lastTransitionTime: "2026-07-24T19:16:26Z"
    message: ""
    reason: RemediationFinishedSuccessfully
    status: "True"
    type: Succeeded
  phase: Fencing-Completed
  timeAssumedRebooted: "2026-07-24T19:16:05Z"

```

@slintes am I correct that your opinion is that the node should NOT reboot. which I think is reasonable as it didn't fix anything in this case ( networking issues ).  What is your suggested remediation, as I don't really know :) .. Thanks  in advance.

docs/design/cp-isolation-detection.md
slintes: What I wrote is a a very basic flow without going into technical details. I don't know if I explained it so badly, but I'm pretty sure there is no design flaw...

>  On a network isolated node, there is no way to detect the SNR CR being created.

that´s why we have all the other steps

> If the peer check does not fail if it can't connect to any peer

of course it does

slintes:
> @slintes am I correct that your opinion is that the node should NOT reboot

No, where do you see that? An isolated node needs to be fenced so that workloads can be rescheduled safely. Fencing happens by reboot. If the reboot fixes the network issue or not is not important. A reboot fixing any issues is a very welcome side effect but not the primary goal of what we do.

docs/design/cp-isolation-detection.md
mpryc: Fixed. The problem statement now says the /readyz checks "do not provide a signal that detects loss of connectivity to the rest of the cluster" - no claim about whether /readyz includes or excludes specific etcd sub-checks. Thanks for the test data showing /healthz catches isolation via the etcd sub-check.

it's  also documented in the "Considered Alternatives" section. See also the API health endpoints docs for reference on the available sub-checks.

docs/design/cp-isolation-detection.md
mpryc: Fixed. The section now correctly states that remaining nodes maintain etcd quorum and the cluster continues operating in degraded mode.

docs/design/cp-isolation-detection.md
mpryc: Fixed. The root cause now describes that /readyz runs multiple sub-checks but does not provide a signal that detects loss of connectivity to the rest of the cluster. Link to k8s sub-checks https://kubernetes.io/docs/reference/using-api/health-checks/

docs/design/cp-isolation-detection.md
mpryc: Added three "Considered Alternatives", please review if I am covering them correctly and if we want to consider more options. I think it's always good to have all of freshly brainstormed in case in the future one will want to modify behavior.

docs/design/cp-isolation-detection.md
mpryc: Dropped. The CP gate now enters the existing isConsideredHealthy() flow directly.

internal/apicheck/check.go
mpryc: Done. [Added rand.Shuffle.](https://github.com/medik8s/self-node-remediation/pull/325/changes#diff-bb88ec6331507bace73a5ee08b0c826b96195b08898305399806110bb80f563aR186)

internal/apicheck/check.go
```diff
@@ -32,16 +33,43 @@ import (
 
 const (
 	eventReasonPeerTimeoutAdjusted = "PeerTimeoutAdjusted"
+
+	peerReachabilityTimeout = 2 * time.Second
+	maxPeersToSample        = 3
 )
```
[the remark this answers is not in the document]
mpryc: intentional design decision. peerReachabilityTimeout (2s, TCP SYN check) is deliberately separate from config.PeerDialTimeout (gRPC health dial). They serve different purposes:
- peerReachabilityTimeout: lightweight "is the network up?" check - should always be fast
- PeerDialTimeout: full gRPC peer health connection - may be tuned higher for slow networks

Merging them would mean operators tuning gRPC timeouts unintentionally slow down isolation detection. The design doc uses the matching name peerReachabilityTimeout=2s to avoid confusion. No change needed imo. Please let me know if you agree ?

docs/design/cp-isolation-detection.md
mpryc: yes, single CP + multiple workers is now explicitly called out in the Edge Cases table. Two separate rows: "Single CP node, no peers (e.g. SNO)" and "Single CP node + multiple workers" - both state no false-positive self-reboot risk, with the mechanism explained (getAllPeerAddresses() includes worker peers, MinPeersForRemediation safety threshold).

mpryc:
Please review this PR and let me know if we all agree on the current design doc and implementation changes or this require further updates/changes.

mpryc:
Closing, because we have now: https://github.com/medik8s/self-node-remediation/pull/339/

PR #331, opened 2026-08-03
Use node `status.addresses` for kubelet health-check

### LINKED ISSUE #330: Support getting kubelet address from `status.addresses`
Related to #267 - in our cluster, the node name is not resolvable. However, the kubelet does have resolvable addresses stored in the node's `status.addresses` that could be used.

I propose the following changes:

- A new configuration option `kubeletPreferredAddressTypes` (similar to the `--kubelet-preferred-address-types` flag of the kube-apiserver), which takes a list of strings: `NodeName` (for current behaviour), `Hostname`, `InternalDNS`, `ExternalDNS`, `InternalIP`, `ExternalIP`. Defaults to `["NodeName"]` for backwards compatibility.
- Try all specified address types in order, if any of them succeeds, consider the kubelet healthy.

Will open a PR with these changes shortly.

kwohlfahrt:
#### Why we need this PR

To support clusters where the node's name is not resolvable, but the addresses in `status.addresses` can be used instead.

#### Changes made

- Add a new config options (`preferredAddressTypes`) that allows selecting addresses from `status.addresses` (and a special `NodeName` value, for the old behaviour, on by default)
- Get node addresses on startup from the node's `status.addresses`, use the enabled ones when trying the kubelet health-check
- Small tweak to `Makefile` to download the correct `protoc` on `aarch64` dev machines

Closes #330.

#### Test plan

Added new unit tests for control-plane manager kubelet liveness checks, and new e2e tests for daemonset rendering.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -406,8 +406,10 @@ bundle-push: ## Push the bundle image.
 
 .PHONY: protoc
 PROTOC = $(shell pwd)/bin/proto/bin/protoc
+# map Go arch names to the ones used by protobuf release artifacts
+PROTOC_ARCH = $(shell go env GOARCH | sed -e 's/amd64/x86_64/' -e 's/arm64/aarch_64/')
 protoc: protoc-gen-go protoc-gen-go-grpc ## Download protoc (protocol buffers tool needed for gRPC)
-	test -f ${PROTOC} || (cd $(shell pwd)/bin/proto && curl -sSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-linux-x86_64.zip && unzip protoc.zip && rm protoc.zip)
+	test -f ${PROTOC} || (cd $(shell pwd)/bin/proto && curl -sSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-linux-$(PROTOC_ARCH).zip && unzip -o protoc.zip && rm protoc.zip)
 
 .PHONY: protoc-gen-go
 PROTOC_GEN_GO = $(shell pwd)/bin/proto/bin/protoc-gen-go
--- a/api/v1alpha1/selfnoderemediationconfig_types.go
+++ b/api/v1alpha1/selfnoderemediationconfig_types.go
@@ -121,9 +121,11 @@ type SelfNodeRemediationConfigSpec struct {
 
 	// PreferredAddressTypes is a list of node address types, that self node remediation agents which run on control-plane nodes will use to try to access the kubelet if they can't contact their peers.
 	// Takes the values accepted on `node.status.addresses.type`, or the special value `"NodeName"`, in which case the node's `metadata.name` will be used.
+	// Addresses are currently only read on startup, so a pod restart is required if node addresses change.
 	// This is a part of self diagnostics which will decide whether the node should be remediated or not.
 	// If empty, it will be equivalent to ["NodeName"] (the previous behaviour).
 	// +optional
+	// +kubebuilder:validation:items:Enum=NodeName;Hostname;InternalDNS;ExternalDNS;InternalIP;ExternalIP
 	PreferredAddressTypes []string `json:"preferredAddressTypes,omitempty"`
 
 	// HostPort is used for internal communication between SNR agents.
--- a/api/v1alpha1/selfnoderemediationconfig_types.go
+++ b/api/v1alpha1/selfnoderemediationconfig_types.go
@@ -119,6 +119,13 @@ type SelfNodeRemediationConfigSpec struct {
 	// +optional
 	EndpointHealthCheckUrl string `json:"endpointHealthCheckUrl,omitempty"`
 
+	// PreferredAddressTypes is a list of node address types, that self node remediation agents which run on control-plane nodes will use to try to access the kubelet if they can't contact their peers.
+	// Takes the values accepted on `node.status.addresses.type`, or the special value `"NodeName"`, in which case the node's `metadata.name` will be used.
+	// This is a part of self diagnostics which will decide whether the node should be remediated or not.
+	// If empty, it will be equivalent to ["NodeName"] (the previous behaviour).
+	// +optional
+	PreferredAddressTypes []string `json:"preferredAddressTypes,omitempty"`
+
 	// HostPort is used for internal communication between SNR agents.
 	// +kubebuilder:default:=30001
 	// +kubebuilder:validation:Minimum=1
--- a/internal/controller/tests/config/selfnoderemediationconfig_controller_test.go
+++ b/internal/controller/tests/config/selfnoderemediationconfig_controller_test.go
@@ -110,6 +110,7 @@ var _ = Describe("SNR Config Test", func() {
 			Expect(container.Image).To(Equal(shared.DsDummyImageName))
 			envVars := getEnvVarMap(container.Env)
 			Expect(envVars["WATCHDOG_PATH"].Value).To(Equal(config.Spec.WatchdogFilePath))
+			Expect(envVars["PREFERRED_ADDRESS_TYPES"].Value).To(Equal(""))
 
 			Expect(len(ds.OwnerReferences)).To(Equal(1))
 			Expect(ds.OwnerReferences[0].Name).To(Equal(config.Name))
@@ -122,6 +123,23 @@ var _ = Describe("SNR Config Test", func() {
 			Expect(container.SecurityContext).ToNot(BeNil())
 			Expect(container.SecurityContext.Privileged).To(Equal(pointer.Bool(true)))
 			Expect(container.SecurityContext.ReadOnlyRootFilesystem).To(Equal(pointer.Bool(true)))
+		})
+		When("Configuration has customized address types", func() {
+			BeforeEach(func() {
+				config.Spec.PreferredAddressTypes = []string{"InternalDNS", "InternalIP"}
+			})
+			It("Daemonset should have comma-separated address types in env var", func() {
+				Eventually(func(g Gomega) {
+					ds = &appsv1.DaemonSet{}
+					g.Expect(k8sClient.Get(context.Background(), dsKey, ds)).Should(BeNil())
+
+					dsContainers := ds.Spec.Template.Spec.Containers
+					g.Expect(len(dsContainers)).To(BeNumerically("==", 1))
+					container := dsContainers[0]
+					envVars := getEnvVarMap(container.Env)
+					g.Expect(envVars["PREFERRED_ADDRESS_TYPES"].Value).To(Equal("InternalDNS,InternalIP"))
+				}, 10*time.Second, 250*time.Millisecond).Should(Succeed())
+			})
 		})
 		When("Configuration has customized tolerations", func() {
 			var expectedToleration corev1.Toleration
--- a/internal/controlplane/manager.go
+++ b/internal/controlplane/manager.go
@@ -5,8 +5,10 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -22,13 +24,16 @@ import (
 )
 
 const (
-	kubeletPort = "10250"
+	kubeletPort         = "10250"
+	nodeNameAddressType = "NodeName"
 )
 
 // Manager contains logic and info needed to fence and remediate controlplane nodes
 type Manager struct {
 	nodeName                     string
 	nodeRole                     peers.Role
+	preferredAddressTypes        []string
+	nodeAddresses                []corev1.NodeAddress
 	endpointHealthCheckUrl       string
 	wasEndpointAccessibleAtStart bool
 	client                       client.Client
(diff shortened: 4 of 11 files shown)


Makefile
```diff
@@ -406,8 +406,10 @@ bundle-push: ## Push the bundle image.
 
 .PHONY: protoc
 PROTOC = $(shell pwd)/bin/proto/bin/protoc
+# map Go arch names to the ones used by protobuf release artifacts
+PROTOC_ARCH = $(shell go env GOARCH | sed -e 's/amd64/x86_64/' -e 's/arm64/aarch_64/')
 protoc: protoc-gen-go protoc-gen-go-grpc ## Download protoc (protocol buffers tool needed for gRPC)
-	test -f ${PROTOC} || (cd $(shell pwd)/bin/proto && curl -sSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-linux-x86_64.zip && unzip protoc.zip && rm protoc.zip)
+	test -f ${PROTOC} || (cd $(shell pwd)/bin/proto && curl -sSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-linux-$(PROTOC_ARCH).zip && unzip -o protoc.zip && rm protoc.zip)
```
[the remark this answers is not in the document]
kwohlfahrt: The download will currently fail, so it's not a silent error. If anyone is developing on these architectures, they can fix it ;)

internal/controlplane/manager.go
```diff
@@ -166,7 +181,28 @@ func (manager *Manager) isEndpointAccessible() bool {
...
+				}
+			}
+		}
+	}
+
+	return false
+}
+
+func (manager *Manager) isKubeletServiceRunningOnAddress(address string) bool {
+	url := fmt.Sprintf("https://%s:%s/pods", address, manager.kubeletPort)
```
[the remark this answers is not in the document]
kwohlfahrt: Good bot, fixed!

internal/controller/tests/config/selfnoderemediationconfig_controller_test.go
```diff
@@ -123,6 +124,23 @@ var _ = Describe("SNR Config Test", func() {
...
+			It("Daemonset should have comma-separated address types in env var", func() {
+				Eventually(func(g Gomega) {
+					ds = &appsv1.DaemonSet{}
+					g.Expect(k8sClient.Get(context.Background(), dsKey, ds)).Should(BeNil())
+
+					dsContainers := ds.Spec.Template.Spec.Containers
+					Expect(len(dsContainers)).To(BeNumerically("==", 1))
+					container := dsContainers[0]
+					envVars := getEnvVarMap(container.Env)
+					Expect(envVars["PREFERRED_ADDRESS_TYPES"].Value).To(Equal("InternalDNS,InternalIP"))
```
razo7: These assertions (lines 137 and 140) use bare `Expect` instead of `g.Expect` inside the `Eventually(func(g Gomega) {...})` callback. When bare `Expect` fails, it panics immediately instead of allowing `Eventually` to retry.

The adjacent toleration test (line 153+) correctly uses `g.Expect` throughout — this block should match that pattern.

```suggestion
					g.Expect(len(dsContainers)).To(BeNumerically("==", 1))
					container := dsContainers[0]
					envVars := getEnvVarMap(container.Env)
					g.Expect(envVars["PREFERRED_ADDRESS_TYPES"].Value).To(Equal("InternalDNS,InternalIP"))
```

internal/controlplane/manager_test.go
```diff
@@ -0,0 +1,121 @@
...
+		})
+
+		It("should fall back to later address types", func() {
+			addresses := []corev1.NodeAddress{{Type: corev1.NodeInternalDNS, Address: unreachableHost}}
+			manager := newTestManager(kubeletHost, []string{"InternalDNS", "NodeName"}, addresses)
+			Expect(manager.isKubeletServiceRunning()).To(BeTrue())
+		})
+
+		It("should contact the kubelet via an IPv6 address", func() {
+			listener, err := net.Listen("tcp", "[::1]:0")
```
razo7, line 92: **suggestion**: This test will fail in CI environments without IPv6 loopback. Consider adding a skip guard:

```go
listener, err := net.Listen("tcp", "[::1]:0")
if err != nil {
    Skip("IPv6 loopback not available: " + err.Error())
}
```

internal/controlplane/manager.go
```diff
@@ -166,7 +182,28 @@ func (manager *Manager) isEndpointAccessible() bool {
...
+					}
+				}
+			}
+		}
+	}
+
+	return false
+}
+
+func (manager *Manager) isKubeletServiceRunningOnAddress(address string) bool {
```
razo7, line 208: **pre-existing / follow-up**: The `http.Client` created inside this function (line 213) has no `Timeout`, and the request uses no `context.Context` with deadline. If a target address accepts TCP but never responds TLS, this blocks indefinitely — and with the new multi-address iteration, a single hanging connection also blocks fallback to later address types.

Not a blocker for this PR (the no-timeout client is pre-existing on `main`), but worth a follow-up:

```go
httpClient := &http.Client{
    Transport: tr,
    Timeout:   10 * time.Second,
}
```

internal/controlplane/manager.go
```diff
@@ -37,9 +42,19 @@ type Manager struct {
 
 // NewManager inits a new Manager return nil if init fails
 func NewManager(nodeName string, myClient client.Client) *Manager {
+	var preferredAddressTypes []string
+	rawPreferredAddressTypes := os.Getenv("PREFERRED_ADDRESS_TYPES")
+	if rawPreferredAddressTypes != "" {
+		preferredAddressTypes = strings.Split(rawPreferredAddressTypes, ",")
```
razo7, line 48: **nit**: `strings.Split` doesn't trim whitespace. If someone manually patches the pod env var with `"InternalIP, NodeName"` (space after comma), `" NodeName"` won't match `"NodeName"`.

Via the normal CRD→template path this can't happen (the `join` template produces no spaces), but for defensive robustness:

```go
for i, t := range preferredAddressTypes {
    preferredAddressTypes[i] = strings.TrimSpace(t)
}
```

internal/controlplane/manager.go
```diff
@@ -125,6 +140,7 @@ func (manager *Manager) initializeManager() error {
 		return wrapWithInitError(err)
 	}
 	manager.setNodeRole(node)
+	manager.nodeAddresses = node.Status.Addresses
```
razo7, line 146: **note**: `nodeAddresses` is cached at init and never refreshed. If node addresses change during the pod's lifetime (IP rotation, secondary NIC), the kubelet check uses stale data.

In practice, this is fine as node addresses rarely change, and the SNR daemonset restarts on config changes. But it's worth documenting this in the CRD field description so users know a pod restart is needed to pick up address changes.

internal/controlplane/manager.go
```diff
@@ -166,7 +182,28 @@ func (manager *Manager) isEndpointAccessible() bool {
...
 
 func (manager *Manager) isKubeletServiceRunning() bool {
-	url := fmt.Sprintf("https://%s:%s/pods", manager.nodeName, kubeletPort)
+	for _, addressType := range manager.preferredAddressTypes {
+		if addressType == "NodeName" {
+			if manager.isKubeletServiceRunningOnAddress(manager.nodeName) {
+				return true
+			}
+		} else {
+			addressType := corev1.NodeAddressType(addressType)
```
razo7: **nit**: This shadows the outer `addressType` (string) with a same-named variable of type `corev1.NodeAddressType`. Functionally correct since it's scoped to the `else` block, but `go vet -shadow` would flag it. Consider renaming to `nodeAddrType` or similar.

razo7 (commented):
Nice contribution @kwohlfahrt! The approach makes sense and it models after kube-apiserver's `--kubelet-preferred-address-types` which is the right pattern, backward compatibility is preserved, and test coverage is solid.

Before applying my `ok-to-test` please address the threads below

internal/controlplane/manager.go
kwohlfahrt: I've added a note to the CRD field. It would be nice to automatically update these, but agree that this is rare and probably not worth the complexity until someone complains about it (especially for control-plane nodes, which is the only place this matters, because AFAIK `etcd` requires stable addresses for peers).

api/v1alpha1/selfnoderemediationconfig_types.go
```diff
@@ -119,6 +119,15 @@ type SelfNodeRemediationConfigSpec struct {
...
 	EndpointHealthCheckUrl string `json:"endpointHealthCheckUrl,omitempty"`
 
+	// PreferredAddressTypes is a list of node address types, that self node remediation agents which run on control-plane nodes will use to try to access the kubelet if they can't contact their peers.
+	// Takes the values accepted on `node.status.addresses.type`, or the special value `"NodeName"`, in which case the node's `metadata.name` will be used.
+	// Addresses are currently only read on startup, so a pod restart is required if node addresses change.
+	// This is a part of self diagnostics which will decide whether the node should be remediated or not.
+	// If empty, it will be equivalent to ["NodeName"] (the previous behaviour).
+	// +optional
+	// +kubebuilder:validation:items:Enum=NodeName;Hostname;InternalDNS;ExternalDNS;InternalIP;ExternalIP
+	PreferredAddressTypes []string `json:"preferredAddressTypes,omitempty"`
```
razo7, line 129: Your new comment is missing from both `bundle/manifests/...yaml` and `config/crd/bases/...yaml`, causing GitHub CI to fail.
Please run `make manifests` (or even `make test`, see also https://github.com/medik8s/self-node-remediation?tab=contributing-ov-file#development-workflow) and commit the regenerated CRD YAMLs so `kubectl explain` shows the full description.

internal/controlplane/manager_test.go
```diff
@@ -0,0 +1,124 @@
...
+			if err != nil {
+				Skip("IPv6 loopback not available: " + err.Error())
+			}
+			Expect(err).NotTo(HaveOccurred())
```
razo7: **nit**: This `Expect(err).NotTo(HaveOccurred())` is unreachable — `Skip()` on line 94 exits the test when `err != nil`, so by the time we reach line 96, `err` is guaranteed to be `nil`. Safe to remove.

razo7 (commented):
Small failing CI change and a small NIT to address 👍🏻

api/v1alpha1/selfnoderemediationconfig_types.go
kwohlfahrt: Whoops, yeah I did it when first adding the field, but forgot after updating. Fixed now!

razo7 (approved):
Test coverage is solid: 8 unit test cases covering env parsing, address type selection, fallback ordering, IPv6, and multiple addresses of the same type. The `.invalid` TLD for unreachable hosts is a nice touch (RFC 6761 guarantees fast failure without network access).

One note: the `InsecureSkipVerify: true` on the kubelet TLS config is pre-existing and not something to address here. It is common across K8s operators since kubelet typically uses self-signed certificates.

**Suggestion:** before merge, consider squashing the fixup commits from review iterations (e.g. "Fix IPv6 address templating", "Use g.Expect() in tests", "Trim spaces from parsed address types", "Fix shadowed variable", "Remove redundant test") into the commits they correct. The final history would read more clearly as 4 or 5 coherent commits: (1) core address type support with CRD changes, (2) aarch64 protoc fix, (3) controller and daemonset template changes, (4) unit tests, (5) manifest regeneration.

kwohlfahrt:
Commits squashed as requested. IIUC this is now waiting for re-approval by @razo7 and a second maintainer?

razo7:
> IIUC this is now waiting for re-approval by @razo7 and a second maintainer?

Correct. Can you have a look as well @mshitrit?

kwohlfahrt:
/retest

No code changes since the last test - `ci/prow/5.0-tls13-adherence` looks like a test infra/DNS issue.

kwohlfahrt:
/retest

There's definitely something wrong with the test environment:

```
 level=fatal msg=failed to fetch Cluster Infrastructure Variables: failed to fetch dependency of "Cluster Infrastructure Variables": failed to generate asset "Platform Provisioning Check": baseDomain: Invalid value: "ocp-ci.medik8s-ci.devcluster.openshift.com": the zone already has record sets for the domain of the cluster: [api.ci-op-gzr4v7xj-d71ed.ocp-ci.medik8s-ci.devcluster.openshift.com. (A), \052.apps.ci-op-gzr4v7xj-d71ed.ocp-ci.medik8s-ci.devcluster.openshift.com. (A)] 
```

razo7 (dismissed):
/retest

internal/controlplane/manager.go
```diff
@@ -37,9 +42,22 @@ type Manager struct {
 
 // NewManager inits a new Manager return nil if init fails
 func NewManager(nodeName string, myClient client.Client) *Manager {
+	var preferredAddressTypes []string
+	rawPreferredAddressTypes := os.Getenv("PREFERRED_ADDRESS_TYPES")
+	if rawPreferredAddressTypes != "" {
+		preferredAddressTypes = strings.Split(rawPreferredAddressTypes, ",")
+	} else {
+		preferredAddressTypes = []string{"NodeName"}
```
mshitrit, line 50: Nit: "NodeName" is quite significant here and we use it multiple times, might be nice to have it as a const

mshitrit (dismissed):
nice PR !
left one non blocking nit (feel free to un-hold and ignore or fix) 

/hold

mshitrit (approved):

razo7 (approved):
Thank you for the contribution and for addressing our comments @kwohlfahrt

PR #334, opened 2026-08-04
Defer annotation update until after watchdog is started

### LINKED ISSUE #332: Node annotation `self-node-remediation.medik8s.io/watchdog-timeout` always `0`
After deploying SNR, with both the `ipmi_watchdog` and `softdog` watchdogs, the node annotation `self-node-remediation.medik8s.io/watchdog-timeout` is always `0`, instead of the correct timeout (30s for IPMI watchdog, manually configured, and 60s default for softdog).

I think I know the cause, I'll try to open a PR with a fix for this shortly.

kwohlfahrt:
#### Why we need this PR

Currently, the watchdog timeout is read upon creating the watchdog, before starting it. This always returns a value of 0, because the timeout is only available after the watchdog's start.

#### Changes made

- Add a `Started()` method to the watchdog interface, that notifies when the watchdog is started
- Add a new runnable that waits for the watchdog to start, and then calls the annotation updater
- Fix a bug where watchdog initialization errors were not surfaced when software fallback is disabled

Fixes #332 

#### Test plan

Added new unit test. I'm open to suggestions for a more thorough e2e test, but am not familiar enough with the test setup to come up with a clean idea.


internal/watchdog/watchdog_test.go
```diff
@@ -81,6 +83,54 @@ var _ = Describe("Watchdog", func() {
...
 
+var _ = Describe("Watchdog start signalling", func() {
+
+	var ctx context.Context
+
+	BeforeEach(func() {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(context.Background())
+		DeferCleanup(cancel)
+	})
```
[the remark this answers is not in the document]
kwohlfahrt: This is fixed by using the Ginkgo test context instead.

cmd/main.go
```diff
@@ -350,20 +350,27 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
...
-	if err = utils.UpdateNodeAnnotations(wasWatchdogInitiated, watchdogTimeout, myNodeName, mgr); err != nil {
-		setupLog.Error(err, "failed to update node's annotation", "annotation", utils.IsRebootCapableAnnotation)
-		os.Exit(1)
+		if err = mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-wd.Started():
+			}
+			return utils.UpdateNodeAnnotations(wd.Status() == watchdog.Armed, wd.GetTimeout(), myNodeName, mgr)
```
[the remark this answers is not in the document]
kwohlfahrt: This should now be fixed - the `wd.Start()` method now surfaces an error in the case were hardware initialization fails and the software fallback is not enabled. This tears down the manager and crashes the agent.

internal/watchdog/synchronized.go
```diff
@@ -52,16 +54,23 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
 	timeout, startErr := swd.impl.start()
 	if startErr != nil {
 		//In case can't use software reboot return an error
-		if isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled(); err != nil || !isSoftwareRebootEnabled {
-			return errors.Wrapf(err, "failed to start watchdog, can't default to software reboot")
+		isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled()
+		if err != nil {
```
razo7: **[HIGH] `started` channel never closed on this error path + lost error context.**

When `IsSoftwareRebootEnabled()` returns an error, `return err` exits `Start()` without closing `swd.started`. The annotation runnable in `main.go` blocks on `<-wd.Started()` indefinitely until the manager's context is cancelled.

While the manager *will* cancel the context (because `Start()` returned an error), this creates a fragile implicit dependency on manager shutdown behavior. If the manager's shutdown ever changes, or if the error is swallowed, the annotation runnable leaks.

Also, `startErr` is silently discarded here — the caller has no idea the watchdog also failed to start. Consider:
```go
return fmt.Errorf("watchdog start failed (%v) and software reboot check errored: %w", startErr, err)
```

Broader fix: use `defer close(swd.started)` at the top of `Start()` (after the disarmed check), or a `sync.Once` to close it in all exit paths. Callers then check `Status()` to distinguish armed vs error.

internal/watchdog/synchronized.go
```diff
@@ -52,16 +54,23 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
...
+			return err
+		}
+
+		if !isSoftwareRebootEnabled {
```
razo7: **[HIGH] `started` channel never closed on this error path either.**

When software reboot is disabled and the watchdog fails to start, `return errors.Wrapf(startErr, ...)` exits without closing `swd.started`. Same annotation runnable hang risk as the path above.

Combined, there are **two paths** where `started` is never closed. Both rely on the manager's context cancellation as a safety net.

cmd/main.go
```diff
@@ -350,20 +350,27 @@ func initSelfNodeRemediationAgent(mgr manager.Manager) {
...
-		watchdogTimeout = wd.GetTimeout()
-	}
-
-	if err = utils.UpdateNodeAnnotations(wasWatchdogInitiated, watchdogTimeout, myNodeName, mgr); err != nil {
-		setupLog.Error(err, "failed to update node's annotation", "annotation", utils.IsRebootCapableAnnotation)
-		os.Exit(1)
+		if err = mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+			select {
+			case <-ctx.Done():
+				return nil
```
razo7: **[MEDIUM] `ctx.Done()` wins the select race → stale annotations persist.**

When context is cancelled before the watchdog starts, `return nil` means no annotation update occurs. The node retains whatever annotation it had from a previous DaemonSet pod run.

If the previous run set `is-reboot-capable=true` (which the old code always did when `wd != nil`), the stale annotation persists even though the node can no longer reboot. NHC would then attempt remediation on a node that cannot self-reboot.

Consider updating annotations to `is-reboot-capable=false` in the `ctx.Done()` path:
```go
case <-ctx.Done():
    return utils.UpdateNodeAnnotations(false, 0, myNodeName, mgr)
```

cmd/main.go
razo7: **[MEDIUM] TOCTOU between `Status()` and `GetTimeout()`.**

`wd.Status()` and `wd.GetTimeout()` each separately acquire and release the mutex. Between the two calls, a concurrent `Stop()` could change the watchdog state. Practical impact is negligible (Stop only happens during active remediation when the node is about to reboot), but a single atomic method like `StatusAndTimeout() (watchdogStatus, time.Duration)` would be more defensible for safety-critical code.

internal/watchdog/watchdog_test.go
```diff
@@ -81,6 +83,45 @@ var _ = Describe("Watchdog", func() {
 	})
```
razo7: **[MEDIUM] Missing test cases for the most dangerous error paths.**

1. **Watchdog fails + software reboot disabled**: `NewFake(false)` + `IS_SOFTWARE_REBOOT_ENABLED=false` → verify `Start()` returns error and `Started()` is NOT closed.
2. **`IsSoftwareRebootEnabled()` returns parse error**: set `IS_SOFTWARE_REBOOT_ENABLED="not-a-bool"` → verify `Start()` returns the parse error and `Started()` is NOT closed.

These are the exact paths that trigger the HIGH findings above, and they have zero coverage.

internal/watchdog/interface.go
```diff
@@ -9,6 +9,9 @@ import (
 type Watchdog interface {
 	// Start should be called by the manager and block on the given context
 	Start(ctx context.Context) error
+	// Returns a channel that is closed when the watchdog is started or software
```
razo7: The current wording ("closed when the watchdog is started or software fallback is activated") could mislead a future consumer into omitting a `ctx.Done()` fallback. Suggest:
```go
// Returns a channel that is closed when the watchdog is started or software
// fallback is activated. The channel may remain open if Start() returns a
// fatal error; callers must also select on their context to avoid blocking.
```

internal/watchdog/synchronized.go
```diff
@@ -52,16 +54,23 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
...
+			return errors.Wrapf(startErr, "failed to start watchdog, software reboot is disabled")
```
razo7: **[LOW/Style]** Unnecessary `else` after `return`. The `if !isSoftwareRebootEnabled` branch returns, so the `else` block can be un-indented. This matches the project's early-return style elsewhere in this file (e.g., line 53 `if swd.status != Disarmed`).

internal/watchdog/synchronized.go
```diff
@@ -52,16 +54,23 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
 	timeout, startErr := swd.impl.start()
 	if startErr != nil {
 		//In case can't use software reboot return an error
-		if isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled(); err != nil || !isSoftwareRebootEnabled {
-			return errors.Wrapf(err, "failed to start watchdog, can't default to software reboot")
+		isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled()
+		if err != nil {
+			return err
+		}
```
razo7, line 99: When `IsSoftwareRebootEnabled()` returns a parse error, `started` is never closed. The annotation runnable in `main.go` blocks on `<-wd.Started()` indefinitely and only unblocks when the manager cancels the context (because this `Start()` returned an error). At that point `ctx.Done()` wins the select, so the annotation is never set.

Closing `started` here lets the annotation runnable proceed and set `is-reboot-capable=false` — which is the correct outcome for a node that can neither use a watchdog nor determine its software reboot status.

Also worth wrapping `startErr` into this return so operators see both failures in the log:

```suggestion
		if err != nil {
			close(swd.started)
			return errors.Wrapf(err, "watchdog start failed (%v) and software reboot check also failed", startErr)
		}
```

internal/watchdog/synchronized.go
```diff
@@ -52,16 +54,23 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
 	timeout, startErr := swd.impl.start()
 	if startErr != nil {
 		//In case can't use software reboot return an error
-		if isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled(); err != nil || !isSoftwareRebootEnabled {
-			return errors.Wrapf(err, "failed to start watchdog, can't default to software reboot")
+		isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled()
+		if err != nil {
+			return err
```
razo7: This bare `return err` is also the root cause of the **CI failure** at `rebooter_test.go:37` (e.g., when running `make test` - https://github.com/medik8s/self-node-remediation/actions/runs/31007892926/job/95290682181?pr=334). That test asserts:
```go
Expect(err.Error()).To(ContainSubstring("failed to start watchdog, can't default to software reboot"))
```
but that error message no longer exists after the error handling was split into two branches.

The test context is named "Software reboot is disabled" but actually tests the **parse error** path, the env var is unset, so `ParseBool("")` fails. The old code hit the same `return` for both cases (`err != nil || !isSoftwareRebootEnabled`), masking the difference.

Suggested fix:
1. Update the existing test's expected substring to match the new error format
2. Rename the context to reflect what it actually tests (parse error, not disabled)
3. Add a new context that explicitly sets `IS_SOFTWARE_REBOOT_ENABLED=false` for the real disabled path:
```go
Context("Software reboot is disabled", func() {
    BeforeEach(func() {
        _ = os.Setenv(utils.IsSoftwareRebootEnabledEnvVar, "false")
    })
    AfterEach(func() {
        _ = os.Unsetenv(utils.IsSoftwareRebootEnabledEnvVar)
    })
    It("watchdog should not start", func() {
        wd := rebooter.wd
        err := wd.Start(context.TODO())
        Expect(err).To(HaveOccurred())
        Expect(err.Error()).To(ContainSubstring("failed to start watchdog, software reboot is disabled"))
        Expect(wd.Status()).To(Equal(watchdog.Disarmed))
    })
})
```

internal/watchdog/synchronized.go
razo7, line 102: Same issue as above when software reboot is disabled, `started` is never closed. The annotation runnable hangs until the manager cancels the context.

Closing `started` before returning lets the annotation runnable set `is-reboot-capable=false`, which correctly tells NHC this node cannot self-remediate.

```suggestion
		if !isSoftwareRebootEnabled {
			close(swd.started)
			return errors.Wrapf(startErr, "failed to start watchdog, software reboot is disabled")
```

internal/watchdog/watchdog_test.go
```diff
@@ -81,6 +83,45 @@ var _ = Describe("Watchdog", func() {
...
+				defer GinkgoRecover()
+				Expect(wd.Start(ctx)).To(Succeed())
+			}()
+
+			Eventually(wd.Started(), 1*time.Second).Should(BeClosed())
+			Expect(wd.Status()).To(Equal(watchdog.Malfunction))
+			Expect(wd.GetTimeout()).To(BeZero())
+		})
+	})
+})
```
razo7, line 165: Two test contexts are missing for the error paths that currently leave `started` open:

1. Watchdog fails + software reboot disabled → `Start()` returns error, `Started()` should still close
2. Watchdog fails + `IsSoftwareRebootEnabled` parse error → same expectation

These would fail against the current code (proving the bug) and pass after adding `close(swd.started)` to the error paths.

```suggestion
	})

	Context("watchdog start fails with software reboot disabled", func() {
		BeforeEach(func() {
			Expect(os.Setenv(utils.IsSoftwareRebootEnabledEnvVar, "false")).To(Succeed())
			DeferCleanup(os.Unsetenv, utils.IsSoftwareRebootEnabledEnvVar)
		})

		It("should close the Started channel so callers are not blocked", func(ctx SpecContext) {
			wd := watchdog.NewFake(false)

			go func() {
				defer GinkgoRecover()
				err := wd.Start(ctx)
				Expect(err).To(HaveOccurred())
			}()

			Eventually(wd.Started(), 1*time.Second).Should(BeClosed())
			Expect(wd.Status()).To(Equal(watchdog.Disarmed))
			Expect(wd.GetTimeout()).To(BeZero())
		})
	})

	Context("watchdog start fails with IsSoftwareRebootEnabled parse error", func() {
		BeforeEach(func() {
			Expect(os.Setenv(utils.IsSoftwareRebootEnabledEnvVar, "not-a-bool")).To(Succeed())
			DeferCleanup(os.Unsetenv, utils.IsSoftwareRebootEnabledEnvVar)
		})

		It("should close the Started channel so callers are not blocked", func(ctx SpecContext) {
			wd := watchdog.NewFake(false)

			go func() {
				defer GinkgoRecover()
				err := wd.Start(ctx)
				Expect(err).To(HaveOccurred())
			}()

			Eventually(wd.Started(), 1*time.Second).Should(BeClosed())
			Expect(wd.Status()).To(Equal(watchdog.Disarmed))
			Expect(wd.GetTimeout()).To(BeZero())
		})
	})
})
```

internal/watchdog/interface.go
```diff
@@ -9,6 +9,9 @@ import (
...
+	// fallback is activated
+	Started() <-chan struct{}
```
razo7, line 15: The doc says "closed when the watchdog is started or software fallback is activated" but doesn't mention what happens on fatal error paths. After applying the `close(swd.started)` fix, the channel closes in all exit paths — the doc should reflect that, and callers should check `Status()` to distinguish armed/malfunction/error:

```suggestion
	// Started returns a channel that is closed when Start() completes,
	// regardless of outcome. Callers should check Status() to determine
	// whether the watchdog armed successfully.
	Started() <-chan struct{}
```

razo7 (commented):
Nice catch on the root cause here, @kwohlfahrt — the `GetTimeout()` before `Start()` race is subtle and the `Started()` channel is a clean solution. Welcome to the project!

The fix is architecturally sound and the new tests cover the happy paths well. I left a few suggestions for closing `started` on the two remaining error paths, plus test cases that would prove those paths work.

Also worth noting: the old code on `main` had a pre-existing bug where `is-reboot-capable` was set to `true` after `NewLinux()` succeeded, even if `Start()` later failed to open the device. This PR improves that by having annotations that reflect the actual watchdog outcome.

internal/watchdog/interface.go
kwohlfahrt: Applied with a slight rewording to "initialization completes" instead of "Start() completes", because `Start()` blocks until the context finishes.

internal/watchdog/synchronized.go
kwohlfahrt: Factored out initialization into a helper function, so both the lock and channel close can be handled with a `defer`, to ensure they are called on all branches.

internal/watchdog/synchronized.go
kwohlfahrt: As above - `started` is closed unconditionally, so this is resolved.

cmd/main.go
razo7: NIT: The old code called `os.Exit(1)` on `UpdateNodeAnnotations` failure, which triggered a kubelet restart — an implicit retry. This new path returns the error to the manager instead. If the API server is transiently unavailable at startup, the node could remain without the `is-reboot-capable` annotation until the pod restarts.

This is consistent with controller-runtime conventions (let the manager handle errors), but worth noting in the PR description as a deliberate behavior change.

internal/watchdog/watchdog_test.go
```diff
@@ -81,6 +83,87 @@ var _ = Describe("Watchdog", func() {
...
+				err := wd.Start(ctx)
+				Expect(err).To(HaveOccurred())
+			}()
+
+			Eventually(wd.Started(), 1*time.Second).Should(BeClosed())
+			Expect(wd.Status()).To(Equal(watchdog.Disarmed))
+			Expect(wd.GetTimeout()).To(BeZero())
+		})
+	})
+})
```
razo7, line 165: **nit**: The post-startup assertions (`Eventually` + `Status` + `GetTimeout`) can be shared across all 4 test contexts with a 2-parameter helper:

```go
assertStartedClosed := func(wd watchdog.Watchdog, expectedStatus watchdogStatus, expectedTimeout time.Duration) {
    Eventually(wd.Started(), 1*time.Second).Should(BeClosed())
    Expect(wd.Status()).To(Equal(expectedStatus))
    Expect(wd.GetTimeout()).To(Equal(expectedTimeout))
}
```

Usage:
```go
assertStartedClosed(wd, watchdog.Armed, 1*time.Second)  // success
assertStartedClosed(wd, watchdog.Malfunction, 0)         // sw reboot enabled
assertStartedClosed(wd, watchdog.Disarmed, 0)            // sw reboot disabled
assertStartedClosed(wd, watchdog.Disarmed, 0)            // parse error
```

This also adds a `Status() == Armed` assertion to the success test, which is currently missing.

internal/watchdog/synchronized.go
```diff
@@ -33,44 +33,34 @@ type synchronizedWatchdog struct {
...
-		if isSoftwareRebootEnabled, err := utils.IsSoftwareRebootEnabled(); err != nil || !isSoftwareRebootEnabled {
-			return errors.Wrapf(err, "failed to start watchdog, can't default to software reboot")
-		} else {
-			swd.status = Malfunction
-			swd.log.Error(startErr, "error while starting watchdog, reverting to software reboot")
-			return nil
-		}
+	if swd.status != Armed {
+		return nil
 	}
```
razo7: NIT: After `init()` returns, the mutex is released (via `defer`). This `swd.status` read happens without the mutex held. Technically a data race if another goroutine calls `Stop()` concurrently, though in practice `Stop()` can't be called before the manager starts runnables.

Not introduced by this PR (same pattern existed on `main`), but if you're already refactoring `Start()` you could wrap this check:
```go
swd.mutex.Lock()
armed := swd.status == Armed
swd.mutex.Unlock()
if !armed {
    return nil
}
```

razo7 (commented):
I like the new changes and I think it is almost ready for approval, while we have few small NITs (one is not from this PR) that could be addressed here.

cmd/main.go
kwohlfahrt: I don't think there's a significant change here - now the `UpdateNodeAnnotations` happens inside a callback added via `mgr.Add(manager.RunnableFunc(...))`, while before it was just in `initSelfNodeRemediationAgent`.

IIUC if any callback added via `mgr.Add` returns an error, the manager will exit with an error and then `os.Exit(1)` is called in `main.go:234`. So it's only a timing change, the retry loop will still happen, just via a graceful shutdown and slightly later.

internal/watchdog/watchdog_test.go
kwohlfahrt: This doesn't work as given, because `watchdogStatus` is not public. Maybe it should be, because `Status()` is part of the public interface? For now, I've just added the missing test-case.

internal/watchdog/synchronized.go
kwohlfahrt: Handled by going via the `swd.Status()` function, which locks.

kwohlfahrt:
I suspect the test failure may have been caused by a conflict on the node object, so trying a `Patch()` instead of `Update()`. This should be safe semantically (we want to set the node state to our computed value unconditionally) and RBAC-wise (we already do `Patch()` on nodes in other places).

It could also have just been a random test flake, but this seems like an improvement that's useful anyway.

internal/reboot/rebooter_test.go
```diff
@@ -29,12 +29,28 @@ var _ = Describe("Rebooter tests", func() {
...
 
+		Context("Software reboot is not configured", func() {
+			It("watchdog should not start", func() {
+				wd := rebooter.wd
+				err := wd.Start(context.TODO())
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("software reboot check failed"))
+				Expect(wd.Status()).To(Equal(watchdog.Disarmed))
+			})
+		})
```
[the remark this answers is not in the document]
kwohlfahrt: I don't think this is the place for the fix. If there is a leaked env var, the test will fail and we can track down the leak, not work around it in the test setup.

internal/watchdog/synchronized.go
```diff
@@ -85,6 +75,39 @@ func (swd *synchronizedWatchdog) Start(ctx context.Context) error {
...
+		if err != nil {
+			return errors.Wrapf(err, "software reboot check failed while handling watchdog start failure: %v", startErr)
+		}
+
+		if !isSoftwareRebootEnabled {
+			return errors.Wrapf(startErr, "failed to start watchdog, software reboot is disabled")
+		} else {
+			swd.status = Malfunction
+			swd.log.Error(startErr, "error while starting watchdog, reverting to software reboot")
+			return nil
```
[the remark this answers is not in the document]
kwohlfahrt: Good bot, nice find!

Fixed, and test added. The added test adds quite a large vendored dependency to this PR, but not much code.

kwohlfahrt:
Fixed the race flagged by the review bot [here](https://github.com/medik8s/self-node-remediation/pull/334#discussion_r3850718506). The bulk of the change is a new vendored dependency.

kwohlfahrt:
/retest

Previous test failed due to hitting VPC limits.

internal/utils/annotations_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	Context("shutdown before the watchdog settles", func() {
+		It("should not touch the node's annotations", func(ctx SpecContext) {
+			wd := watchdog.NewFake(true)
+			updater := utils.NewAnnotationUpdater(wd, nodeName, k8sClient, k8sClient)
+
+			cancelledCtx, cancel := context.WithCancel(ctx)
+			cancel()
+			Expect(updater.Start(cancelledCtx)).To(Succeed())
+
```
razo7: NIT: we can remove this blank line

razo7 (dismissed):
Switching from `client.Update()` to `client.MergeFrom().Patch()` for node annotations eliminates conflicts with concurrent node status updates from the kubelet. Good improvement.

The vendor addition (6853 lines) is the transitive cost of pulling in `controller-runtime/pkg/client/fake` for the annotation shutdown test. All standard k8s.io dependencies, nothing third-party.

**Suggestion:** before merge, consider squashing the fixup commits from review iterations (e.g. "Fix test context handling", "Get status via locking helper function", "Test watchdog status is armed on success") into the commits they correct. The final history would read more clearly as 3 or 4 coherent commits: (1) core watchdog `Started()` channel and annotation updater, (2) merge patch for node annotations, (3) shutdown annotation handling with tests.

kwohlfahrt:
Commits squashed as requested (and imports sorted as per the failing test).

vendor/sigs.k8s.io/controller-runtime/pkg/client/fake/doc.go
```diff
@@ -0,0 +1,38 @@
...
+
+	client := NewClientBuilder().WithScheme(scheme).WithObjects(initObjs...).Build()
+
+You can invoke the methods defined in the Client interface.
+
+When in doubt, it's almost always better not to use this package and instead use
+envtest.Environment with a real client and API server.
+
+WARNING: ⚠️ Current Limitations / Known Issues with the fake Client ⚠️
+  - This client does not have a way to inject specific errors to test handled vs. unhandled errors.
```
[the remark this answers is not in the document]
kwohlfahrt: This, and all of the other comments from this batch are in vendored upstream code.

vendor/k8s.io/apimachinery/pkg/util/rand/rand.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return string(b)
+}
+
+// SafeEncodeString encodes s using the same characters as rand.String. This reduces the chances of bad words and
+// ensures that strings generated from hash functions appear consistent throughout the API.
+func SafeEncodeString(s string) string {
+	r := make([]byte, len(s))
+	for i, b := range []rune(s) {
+		r[i] = alphanums[(int(b) % len(alphanums))]
```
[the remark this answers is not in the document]
kwohlfahrt: This, and all of the other comments from this batch are in vendored upstream code.

vendor/k8s.io/apimachinery/pkg/util/rand/rand.go
```diff
@@ -0,0 +1,127 @@
...
+	defer rng.Unlock()
+	return rng.rand.Intn(max)
+}
+
+// IntnRange generates an integer in range [min,max).
+// By design this should panic if input is invalid, <= 0.
+func IntnRange(min, max int) int {
+	rng.Lock()
+	defer rng.Unlock()
+	return rng.rand.Intn(max-min) + min
```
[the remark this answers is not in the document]
kwohlfahrt: This, and all of the other comments from this batch are in vendored upstream code.

razo7 (changes requested):
(created by mistake)

razo7 (commented):
(created by mistake)

internal/utils/annotations.go
```diff
@@ -12,31 +11,55 @@ import (
...
+		case <-wd.Started():
+		}
+		return UpdateNodeAnnotations(wd.Status() == watchdog.Armed, wd.GetTimeout(), nodeName, reader, writer)
+	})
+}
+
+// UpdateNodeAnnotations updates the is-reboot-capable and watchdog timeout node annotations.
+// The reader should bypass any node-scoped cache, e.g. manager.GetAPIReader(),
+// so it can run during manager shutdown.
+func UpdateNodeAnnotations(watchdogInitiated bool, watchdogTimeout time.Duration, nodeName string, reader client.Reader, writer client.Writer) error {
```
razo7: NIT: During manager shutdown, if the annotation update is triggered (lines 34-38 in NewAnnotationUpdater), the `UpdateNodeAnnotations` call uses `context.Background()` with no deadline. If the API server is slow/unresponsive, this can hang indefinitely.

**Suggested fix**:

1. Add `ctx context.Context` parameter to `UpdateNodeAnnotations` function signature
2. In `NewAnnotationUpdater`, wrap the post-cancellation cleanup call with a timeout:

```go
case <-ctx.Done():
  select {
  case <-wd.Started():
  default:
    return nil
  }
  // Add timeout to prevent indefinite hang during shutdown
  cleanupCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
  defer cancel()
  return UpdateNodeAnnotations(wd.Status() == watchdog.Armed, wd.GetTimeout(), nodeName, reader, writer, cleanupCtx)
```

This ensures graceful shutdown completes even if the API server is slow.

razo7 (dismissed):
I have found a (new) small NIT with context, other than that LGTM

internal/utils/annotations.go
kwohlfahrt: I think this is a pre-existing issue, since it used `context.Background` before too. But happy to fix it - I now use the main context if we have it, or a 5-second timeout if we're already shutting down.

Also slightly refactored both annotation updates to go via `NewAnnotationUpdater` - the branch on `wd == nil` is moved to that function.

razo7 (approved):

mshitrit (approved):
Nice PR, +1 approved !

PR #335, opened 2026-08-24
WIP [RHWA-1371]: use pqc ready ubi-minimal for ocp5

weshayutin:
podman pull and podman build worked locally
Testing w/ prow ci


weshayutin:
This change is probably / hopefully not needed at all.  Digging through the RH / OpenShift network of requirements to confirm now :/

PR #339, opened 2026-09-03
Fix CP node isolation detection via existing peer health check

mshitrit:
#### Why we need this PR

SNR cannot remediate isolated control plane nodes because `/readyz?exclude=shutdown` returns 200 when the local kube-apiserver is reachable, even when the node is network-isolated from the rest of the cluster. The `/readyz` endpoint does not check etcd connectivity, so isolated CP nodes continue feeding the watchdog indefinitely instead of triggering remediation.

This is an alternative fix to PR #325 that reuses existing battle-tested peer health check logic instead of introducing new TCP dial infrastructure.

#### Changes made

- Added CP isolation detection gate after `/readyz` passes on control plane nodes
- Calls existing `getControlPlanePeersStatus()` to verify peers are reachable before resetting error count
- If no CP peers reachable, treats `/readyz` success as false positive and enters peer consultation flow
- Added `isControlPlane()` helper method to deduplicate control plane detection logic
- Worker node behavior unchanged
- Added integration test for [RHWA-1384](https://redhat.atlassian.net/browse/RHWA-1384) scenario (readyz passes but CP peers unreachable)
- Refactored CP isolation test suite to eliminate code duplication using shared parent context

[RHWA-1384](https://redhat.atlassian.net/browse/RHWA-1384)

#### Test plan

- Integration test validates watchdog triggers when `/readyz` passes but all CP peers are unreachable
- Test uses real API config (readyz succeeds) with no peerhealth server running (peers unreachable)
- Verifies remediation triggers within expected timeout (~90s)

### DIFF
--- a/internal/apicheck/check.go
+++ b/internal/apicheck/check.go
@@ -115,6 +115,8 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 		// /readyz does not check etcd connectivity. Verify CP peer health before
 		// concluding the node is healthy.
 		if c.isControlPlane() {
+			// Directly check peer reachability to detect isolation.
+			// Skip errorCount threshold since readyz succeeded (not an API error).
 			canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
 
 			// If CP peers say we're unhealthy, trigger remediation immediately
@@ -130,6 +132,8 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 			// BUT: only treat as isolation if peers actually exist
 			peersExist := len(c.config.Peers.GetPeersAddresses(peers.ControlPlane)) > 0
 			if !canBeReached && peersExist {
+				// Note: If 2/3 CPs are dead (not isolated), this remediates the survivor.
+				// Acceptable since cluster already lacks quorum and is non-functional.
 				c.config.Log.Info("CP node: readyz passed but no CP peers reachable, treating as failure")
 				if isHealthy := c.isConsideredHealthy(); !isHealthy {
 					c.config.Log.Info("CP isolation detected despite passing readyz, triggering reboot")
--- a/internal/apicheck/check.go
+++ b/internal/apicheck/check.go
@@ -110,7 +110,40 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 			return
 		}
 
-		// reset error count after a successful API call
+		// On control plane nodes, a passing readyz check may be a false positive:
+		// the local API server responds 200 even when network-isolated because
+		// /readyz does not check etcd connectivity. Verify CP peer health before
+		// concluding the node is healthy.
+		if c.isControlPlane() {
+			canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
+
+			// If CP peers say we're unhealthy, trigger remediation immediately
+			if cpUnhealthy {
+				c.config.Log.Info("CP peers report this node as unhealthy, triggering reboot")
+				if err := c.config.Rebooter.Reboot(); err != nil {
+					c.config.Log.Error(err, "failed to trigger reboot")
+				}
+				return
+			}
+
+			// If no CP peers are reachable, readyz was a false positive
+			// BUT: only treat as isolation if peers actually exist
+			peersExist := len(c.config.Peers.GetPeersAddresses(peers.ControlPlane)) > 0
+			if !canBeReached && peersExist {
+				c.config.Log.Info("CP node: readyz passed but no CP peers reachable, treating as failure")
+				if isHealthy := c.isConsideredHealthy(); !isHealthy {
+					c.config.Log.Info("CP isolation detected despite passing readyz, triggering reboot")
+					if err := c.config.Rebooter.Reboot(); err != nil {
+						c.config.Log.Error(err, "failed to trigger reboot")
+					}
+				} else {
+					c.config.Log.Info("not all peers confirmed isolation, waiting for threshold")
+				}
+				return
+			}
+		}
+
+		// reset error count after a fully successful check
 		c.errorCount = 0
 
 	}, c.config.CheckInterval)
@@ -118,12 +151,16 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 	return nil
 }
 
+// isControlPlane returns true if this node is a control plane node.
+func (c *ApiConnectivityCheck) isControlPlane() bool {
+	return c.controlPlaneManager != nil && c.controlPlaneManager.IsControlPlane()
+}
+
 // isConsideredHealthy keeps track of the number of errors reported, and when a certain amount of error occur within a certain
 // time, ask peers if this node is healthy. Returns if the node is considered to be healthy or not.
 func (c *ApiConnectivityCheck) isConsideredHealthy() bool {
 	workerPeersResponse := c.getWorkerPeersResponse()
-	isWorkerNode := c.controlPlaneManager == nil || !c.controlPlaneManager.IsControlPlane()
-	if isWorkerNode {
+	if !c.isControlPlane() {
 		return workerPeersResponse.IsHealthy
 	}
 	canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
--- a/internal/controller/tests/controller/selfnoderemediation_controller_test.go
+++ b/internal/controller/tests/controller/selfnoderemediation_controller_test.go
@@ -3,6 +3,7 @@ package testcontroler
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -509,16 +511,20 @@ var _ = Describe("SNR Controller", func() {
 		})
 	})
 
-	// Regression test for https://github.com/medik8s/self-node-remediation/issues/251:
-	// A control-plane node whose worker-peer path yields no result (isolated from workers)
-	// must still remediate when peer CP nodes have a SNR CR for it (consider it unhealthy).
-	Context("Control plane node - API server down - CP peer reports it unhealthy", func() {
+	// Tests for control plane node isolation detection scenarios
+	Context("Control plane node isolation scenarios", func() {
 		const cpPeerHealthPort = 9001
 
-		var cpDummyDog watchdog.FakeWatchdog
+		var (
+			cpDummyDog watchdog.FakeWatchdog
+			cpPeers    *peerspkg.Peers
+			cpManager  *controlplane.Manager
+			certReader *certificates.MemoryCertStorage
+		)
 
+		// Shared setup for all CP isolation tests
 		BeforeEach(func() {
-			// 1. Label both nodes as control-plane so the Peers selector works correctly.
+			// 1. Label both nodes as control-plane so the Peers selector works correctly
 			for _, nodeName := range []string{shared.UnhealthyNodeName, shared.PeerNodeName} {
 				node := &corev1.Node{}
 				Expect(k8sClient.Client.Get(context.Background(), client.ObjectKey{Name: nodeName}, node)).To(Succeed())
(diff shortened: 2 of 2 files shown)


internal/apicheck/check.go
```diff
@@ -110,20 +110,57 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
+				}
+				return
+			}
+
+			// If no CP peers are reachable, readyz was a false positive
+			// BUT: only treat as isolation if peers actually exist
+			peersExist := len(c.config.Peers.GetPeersAddresses(peers.ControlPlane)) > 0
+			if !canBeReached && peersExist {
+				c.config.Log.Info("CP node: readyz passed but no CP peers reachable, treating as failure")
+				if isHealthy := c.isConsideredHealthy(); !isHealthy {
```
mpryc, line 134: The `isConsideredHealthy()` calls again `getControlPlanePeersStatus()` - see below line 166, which we had already on line 118, can this be optimized?

internal/apicheck/check.go
```diff
@@ -110,20 +110,57 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
+				if err := c.config.Rebooter.Reboot(); err != nil {
+					c.config.Log.Error(err, "failed to trigger reboot")
+				}
+				return
+			}
+
+			// If no CP peers are reachable, readyz was a false positive
+			// BUT: only treat as isolation if peers actually exist
+			peersExist := len(c.config.Peers.GetPeersAddresses(peers.ControlPlane)) > 0
+			if !canBeReached && peersExist {
```
mpryc, line 132: Should we increase error count  before entering `isConsideredHealthy()` ?

internal/apicheck/check.go
```diff
@@ -110,20 +110,57 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
-		// reset error count after a successful API call
+		// On control plane nodes, a passing readyz check may be a false positive:
+		// the local API server responds 200 even when network-isolated because
+		// /readyz does not check etcd connectivity. Verify CP peer health before
+		// concluding the node is healthy.
+		if c.isControlPlane() {
+			canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
+
+			// If CP peers say we're unhealthy, trigger remediation immediately
+			if cpUnhealthy {
```
mpryc, line 121: This is changing the existing path of remediation and bypasses all the threasholds, is this intentional?

internal/apicheck/check.go
```diff
@@ -110,20 +110,57 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
 			return
 		}
 
-		// reset error count after a successful API call
+		// On control plane nodes, a passing readyz check may be a false positive:
+		// the local API server responds 200 even when network-isolated because
+		// /readyz does not check etcd connectivity. Verify CP peer health before
+		// concluding the node is healthy.
+		if c.isControlPlane() {
+			canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
```
mpryc, line 118: So the call here and than call in the `peersExist` calls `GetPeersAddresses()` again. The list may change in between?

internal/apicheck/check.go
mshitrit: Well it can.
My first thought was passing `canBeReached` and `cpUnhealthy` to `isConsideredHealthy`.
However this fills like a an un-elegant solution so I prefered the overhead of executing `getControlPlanePeersStatus` - which I think is negligible. 
Let me know if you had something else in mind or feels strongly to go forward with this optimization.

internal/apicheck/check.go
mshitrit: I don't think so, it'll be increased in `getWorkerPeersResponse` which is triggered by `isConsideredHealthy`, so increasing it before will cause a double increase.

I do agree though that the way it's currently increased is confusing

internal/apicheck/check.go
mshitrit: Yep. the error threshold is used when there is an API Server error, in this case we get 200 ok (but may still need to remediate) so there is not point counting API Server errors.

internal/apicheck/check.go
mshitrit: Good question, made me think.

In theory yes.
In practice I think it's very unlikely (cp number will need to change in the `apiServerTimeout` update window between both calls) and even if it does, I don't think it'll cause a significant impact.
I couldn't come up with a scenario that will create an issue, let me know what you think.

mpryc: Ack, unlikely in practice. Fine as-is.

internal/apicheck/check.go
mpryc: Got it, thanks for clarifying.

internal/apicheck/check.go
mpryc: Fair enough, overhead is negligible.

internal/apicheck/check.go
mpryc: Makes sense since it's not an API error. Can you add a brief comment in the code explaining why this path skips the error threshold?

mshitrit: SGTM, Done :+1:

weshayutin:
@mshitrit please spend a little time here to bring it out of draft and start running tests please. 
thanks 0/

mpryc (dismissed):

razo7 (commented):
none

internal/apicheck/check.go
```diff
@@ -110,20 +110,59 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
+		// concluding the node is healthy.
+		if c.isControlPlane() {
+			// Directly check peer reachability to detect isolation.
+			// Skip errorCount threshold since readyz succeeded (not an API error).
+			canBeReached, cpUnhealthy := c.getControlPlanePeersStatus()
+
+			// If CP peers say we're unhealthy, trigger remediation immediately
+			if cpUnhealthy {
+				c.config.Log.Info("CP peers report this node as unhealthy, triggering reboot")
+				if err := c.config.Rebooter.Reboot(); err != nil {
```
razo7, line 125: Immediate reboot bypasses `MaxErrorsThreshold`. Line 295 uses first-responder pattern (ANY peer triggers reboot). Risk: single compromised peer or transient unavailability triggers premature remediation without confirmation cycles. Existing `isConsideredHealthy()` (lines 163-175) already handles `cpUnhealthy`, so we can route through it for consistency:

```go
if cpUnhealthy || (!canBeReached && peersExist) {
    if isHealthy := c.isConsideredHealthy(); !isHealthy {
        c.config.Log.Info("CP isolation/unhealthy confirmed, triggering reboot")
        if err := c.config.Rebooter.Reboot(); err != nil {
            c.config.Log.Error(err, "failed to trigger reboot")
        }
    }
    return
}
```

internal/apicheck/check.go
```diff
@@ -110,20 +110,59 @@ func (c *ApiConnectivityCheck) Start(ctx context.Context) error {
...
+					c.config.Log.Error(err, "failed to trigger reboot")
+				}
+				return
+			}
+
+			// If no CP peers are reachable, readyz was a false positive
+			// BUT: only treat as isolation if peers actually exist
+			peersExist := len(c.config.Peers.GetPeersAddresses(peers.ControlPlane)) > 0
+			if !canBeReached && peersExist {
```
razo7, line 134: Quorum loss: if 2/3 CP nodes fail, surviving node triggers remediation → 0/3 CP available. Can't distinguish "I'm isolated" from "they're all dead" (both = `!canBeReached && peersExist`).

**Option 1 - Integrate etcd health:**
```go
if !canBeReached && peersExist {
    etcdHealthy, err := checkEtcdQuorum()
    if err != nil || !etcdHealthy {
        c.config.Log.Info("etcd quorum issue, deferring isolation check")
        return
    }
    // etcd healthy but peers unreachable = true isolation
    if isHealthy := c.isConsideredHealthy(); !isHealthy { ... }
}
```

**Option 2 - Accept + document:** Code comment + operator docs explaining limitation.

razo7 (commented):
PR correctly addresses [RHWA-1384](https://redhat.atlassian.net/browse/RHWA-1384) by reusing battle-tested `getControlPlanePeersStatus()` mechanism. Strong mTLS validation provides good defense-in-depth (TLS 1.3, client cert verification).

**E2E Validation:** [RHWA-1384](https://redhat.atlassian.net/browse/RHWA-1384) is critical failure mode (isolated CP never remediates). Integration test validates code path but doesn't replicate production environment where `/readyz` returns 200 despite etcd partition (see [openshift/kubernetes#2174](https://github.com/openshift/kubernetes/pull/2174) from last year that deliberately removes etcd from `/readyz`). I would like to see E2E validation on real cluster with network partition, to avoid regression or at least a manual confirmation that 1384 is not repeatable anymore, before merge to confirm fix addresses root cause.

internal/apicheck/check.go
razo7, line 134: **NIT:** `peersExist=false` skips isolation check (correct for SNO). Add comment explaining design:

```go
if !peersExist {
    // SNR not designed for SNO - can't distinguish
    // isolation from single-node operation
}
```

internal/apicheck/check.go
mshitrit: I think this comment is similar to [that](https://github.com/medik8s/self-node-remediation/pull/339#discussion_r3936795540) one

internal/apicheck/check.go
mshitrit: Here are my thoughts:

- I don't think it really matters to distinguish the use cases (worst case we remediate a cp of an already unworking cluster)
- I don't think checkEtcdQuorum will work here or get us any additional info, it's too late since we already established the node is isolated

I think adding a comment explaining this makes sense aligning with option 2

mshitrit: Not sure I follow.
Why SNO, maybe the cp is just isolated no ?

razo7: >I don't think it really matters to distinguish the use cases (worst case we remediate a cp of an already unworking cluster)

I think we introduce more harm by remediating the "surviving CP node" if we don't make this distinction. I think this is an edge case that we should ak

>I think adding a comment explaining this makes sense aligning with option 2

SGTM, but it should be propagated to the ticket as a known limitation for our release notes and maybe upstream at https://www.medik8s.io/faq

razo7: The code distinguishes two scenarios:

1. **`peersExist=false`** → No CP peers exist in cluster (SNO/single-node deployment) → skip isolation check ✅ correct
2. **`!canBeReached && peersExist=true`** → CP peers exist but unreachable (true isolation) → trigger remediation

This NIT is for scenario 1: when there are literally no other CP nodes configured (e.g., SNO), the isolation check is skipped because SNR requires peers to function. The code is correct; just suggesting a comment explaining *why* we skip when `peersExist=false`.

For scenario 2 (isolation), the check proceeds normally via line 134's condition.

internal/apicheck/check.go
razo7: Thanks for pointing to the earlier discussion. You're right that the threshold bypass was already discussed.
My remaining concern is the **first-responder pattern** at line 295:

```go
return (healthyResponses + unhealthyResponses + apiErrorsResponses) > 0,
       unhealthyResponses > 0  // ANY peer unhealthy triggers cpUnhealthy
```

Combined with the immediate reboot path, this means:
- Single peer reporting "unhealthy" → immediate remediation
- No consensus/majority requirement
- No confirmation cycles

**Question:** Is this first-responder design intentional? The concern is:
1. (could happen) Transient peer unavailability (network blip, pod restart) → single unhealthy response → premature reboot
2. (less likely to happen) While mTLS mitigates unauthorized peers, a compromised peer pod could trigger unwanted reboots

For worker nodes, the threshold mechanism provides confirmation cycles. For CP nodes on this path, we get immediate action on first peer signal. Was this trade-off considered, or should we route through `isConsideredHealthy()` for consistency or something else?

internal/apicheck/check.go
mshitrit: My point was that `peersExist=false` does not necessarily means SNO.
It just means that we didn't find peers, maybe it's SNO and maybe CP is isolated or maybe peers have other issues.

internal/apicheck/check.go
mshitrit: Good question ! , but I don't think it's relevant for this PR.

For CP node both [first-responder pattern](https://github.com/medik8s/self-node-remediation/blob/main/internal/apicheck/check.go#L256) and immediately [acting](https://github.com/medik8s/self-node-remediation/blob/main/internal/apicheck/check.go#L101-L106) on it preceded this PR. 
This PR just following the current convention. 

Having said that I'm thinking that a fellow CP peer explicitly saying a node is unhealthy might be a strong indication as well as the fact that we had this setup running for quite awhile, but TBH those are just educated guess on my part.

I do agree though that there might be an improvement potential here, so I think it's fine to continue looking into it.

internal/apicheck/check.go
razo7: Agreed the first-responder pattern pre-exists, but the immediate reboot path (lines 123-127) is new:

```go
if cpUnhealthy {
    c.config.Log.Info("CP peers report this node as unhealthy, triggering reboot")
    if err := c.config.Rebooter.Reboot(); err != nil { ... }
    return  // NEW
}
```

Opened **#344** to track consensus/threshold improvements separately. This PR works as designed given current conventions.

internal/apicheck/check.go
razo7: Agreed the first-responder pattern pre-exists, but the immediate reboot path (lines 123-127) is new:
```
if cpUnhealthy {
    c.config.Log.Info("CP peers report this node as unhealthy, triggering reboot")
    if err := c.config.Rebooter.Reboot(); err != nil { ... }
    return  // NEW
}
```
Opened https://github.com/medik8s/self-node-remediation/issues/344 to track consensus/threshold improvements separately. This PR works as designed given current conventions.

razo7 (approved):

internal/apicheck/check.go
mshitrit: >but the immediate reboot path (lines 123-127) is new

I don't think it's an accurate statement.

It's new only for the use case where there isn't an API-Server Error (the code snippet you've shown), but
immediate reboot for CP [already existed](https://github.com/medik8s/self-node-remediation/blob/main/internal/apicheck/check.go#L101-L106) for API-Server error use case (note that CP node `isConsideredHealthy` already bypass threashhold).

mpryc (approved):

mshitrit:
/override ci/prow/5.0-openshift-e2e

Test already [passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/medik8s_self-node-remediation/339/pull-ci-medik8s-self-node-remediation-main-5.0-openshift-e2e/2095444125498216448) in current iteration (delta was comment update) and currently failing due to prow issues

PR #300, opened 2026-05-11
Add telco best practices certsuite compliance

slintes:
## Summary

Addresses certsuite test failures to qualify SNR for the "Meets Best Practices" label in RHEC (Red Hat Ecosystem Catalog).

Related: [RHWA-449](https://redhat.atlassian.net/browse/RHWA-449) / [RHWA-989](https://redhat.atlassian.net/browse/RHWA-989)

## Problem

Certsuite testing flagged missing lifecycle hooks, startup probe, and terminationMessagePolicy — preventing the operator from receiving the "Meets Best Practices" certification label required for telco customers and RHEC visibility.

### Manager Container (`config/manager/manager.yaml`)

**Added:**
- `startupProbe`: httpGet on `/healthz:8081` with 12 attempts × 5s = 60s budget
  - Protects slow-starting containers from premature liveness probe termination
- `lifecycle.postStart`: no-op `/bin/true` (certsuite compliance requirement)
- `lifecycle.preStop`: `sleep 5` for endpoint propagation delay
  - Prevents traffic routing to terminating pods during endpoint update propagation
- `terminationMessagePolicy: FallbackToLogsOnError`
  - Captures last log output as termination message on crash

**Modified:**
- Removed `initialDelaySeconds: 15` from livenessProbe (superseded by startupProbe)
- Bumped `terminationGracePeriodSeconds` from 10 to 15 (accommodate 5s preStop + actual shutdown)

### Bundle

Regenerated bundle (`make bundle-reset`) with all changes.

## Certsuite Tests Addressed

✅ **Fixed in this PR:**
1. `lifecycle-container-poststart` — Added postStart hook
2. `lifecycle-container-prestop` — Added preStop hook
3. `lifecycle-startup-probe` — Added startup probe
4. `observability-termination-policy` — Set FallbackToLogsOnError

## Testing

- ✅ Bundle validation passes (`make bundle-validate`)
- ✅ CSV includes all new fields in manager container deployment

## Notes

- postStart hook is a no-op (`/bin/true`) for compliance only; container doesn't require initialization before serving
- Same pattern as NHC operator (PR medik8s/node-healthcheck-operator#406)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Enhanced pod health monitoring with new startup probes to better detect container initialization issues early
  * Improved graceful shutdown behavior with lifecycle management hooks (postStart and preStop)
[...]

PR #341, opened 2026-09-07
Add production-readiness settings to DaemonSet agent

razo7:
## Why we need this PR

The DaemonSet agent runs on every node but lacks the production-readiness settings already present on the controller-manager: no health probes (kubelet cannot detect or restart a hung agent), `imagePullPolicy: Always` (agent pods fail to start when the registry is unreachable), and crash output is lost (`terminationMessagePolicy: File`). This PR adds the same settings established for the controller-manager in PR #300, adapted for the DaemonSet's design.

Related: [RHWA-1616](https://redhat.atlassian.net/browse/RHWA-1616)

## Changes made

Applied production-readiness settings to `install/self-node-remediation-deamonset.yaml`:

| Change | Detail |
|--------|--------|
| `imagePullPolicy` | `Always` to `IfNotPresent` |
| `livenessProbe` | httpGet `/healthz:8081`, period 20s |
| `readinessProbe` | httpGet `/readyz:8081`, initialDelay 5s, period 10s |
| `startupProbe` | httpGet `/healthz:8081`, failureThreshold 12, period 5s (60s budget) |
| `terminationMessagePolicy` | `File` to `FallbackToLogsOnError` |

The agent registers `/healthz` and `/readyz` endpoints via controller-runtime's health probe server on port 8081 (`cmd/main.go:281-288`), used by both manager and agent modes.

**Note:** Lifecycle hooks (`preStop`, `postStart`) are intentionally omitted. The DS agent is not behind a Service, so `preStop` sleep has no endpoints to drain. `postStart /bin/true` is a no-op with no operational value. Omitting both avoids adding complexity that serves no purpose for the agent's architecture.

## Test plan

- [x] DaemonSet pods start and operate normally with new settings
- [x] Probes respond correctly on port 8081
- [x] Peer health checking works with `IfNotPresent` pull policy
- [x] Remediation flow (watchdog + software reboot) unaffected

### DIFF
--- a/install/self-node-remediation-deamonset.yaml
+++ b/install/self-node-remediation-deamonset.yaml
@@ -39,6 +39,7 @@ spec:
       containers:
       - args:
         - --is-manager=false
+        - --health-probe-bind-address=:8081
         command:
         - /manager
         env:
--- a/install/self-node-remediation-deamonset.yaml
+++ b/install/self-node-remediation-deamonset.yaml
@@ -75,7 +75,24 @@ spec:
           - name: MIN_PEERS_FOR_REMEDIATION
             value: "{{.MinPeersForRemediation}}"
         image: {{.Image}}
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
+          periodSeconds: 20
+        readinessProbe:
+          httpGet:
+            path: /readyz
+            port: 8081
+          initialDelaySeconds: 5
+          periodSeconds: 10
+        startupProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
+          failureThreshold: 12
+          periodSeconds: 5
         volumeMounts:
           - name: devices
             mountPath: /dev
@@ -92,7 +109,7 @@ spec:
             cpu: 20m
             memory: 60Mi
         terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
+        terminationMessagePolicy: FallbackToLogsOnError
       dnsPolicy: ClusterFirst
       restartPolicy: Always
       schedulerName: default-scheduler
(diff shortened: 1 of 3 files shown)


jmontleon (approved):
LGTM

install/self-node-remediation-deamonset.yaml
```diff
@@ -75,7 +75,24 @@ spec:
           - name: MIN_PEERS_FOR_REMEDIATION
             value: "{{.MinPeersForRemediation}}"
         image: {{.Image}}
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
```
mshitrit, line 83: The probes here check port 8081, but that value only matches the agent's actual listening port by coincidence — it's the Go flag's coded default, not something pinned in this manifest. 

The controller-manager avoids this by passing --health-probe-bind-address explicitly in its own args; this DaemonSet doesn't. If that default ever changes for an unrelated reason, every
agent pod's probes would silently break fleet-wide with nothing catching it. 

Worth pinning it here too:

```
  args:
  - --is-manager=false
  - --health-probe-bind-address=:8081
  command:
  - /manager
```

mshitrit (dismissed):
/hold
Nice PR !
I left one comment, but I'm considering it a blocker.
Feel free to either update or unhold

install/self-node-remediation-deamonset.yaml
razo7: >Worth pinning it here too:

SGTM

mshitrit (approved):

install/self-node-remediation-deamonset.yaml
```diff
@@ -39,6 +39,7 @@ spec:
       containers:
       - args:
         - --is-manager=false
+        - --health-probe-bind-address=:8081
```
mpryc, line 42: Just informational note: The DaemonSet does not use hostNetwork, so port 8081 for health probes is pod-scoped - no conflict with the manager Deployment binding the same port on the same node.

mpryc (approved):
/lgtm

PR #349, opened 2026-09-17
:bug: Fix Makefile and Dockerfile to work on Mac hosts

pranavgaikwad:
Minor fixes to make container builds work on Mac hosts through the hack/local-run.sh scripts we added recently that emulate the GitHub action runs.


eemcmullan (approved):

Makefile
```diff
@@ -189,6 +189,19 @@ test: go-verify envtest generate fix-imports manifests fmt vet ## Run tests.
...
+
+.PHONY: test-linux
+test-linux: ## Run the full test target in a Linux container (for macOS hosts).
+	$(CONTAINER_TOOL) run --rm \
+		-v "$(CURDIR):/workspace$(if $(filter podman,$(CONTAINER_TOOL)),:Z,)" \
+		-v /workspace/bin \
+		-v /workspace/testbin \
+		-w /workspace -e GOTOOLCHAIN=auto -e TEST_OPS \
+		$(TEST_LINUX_IMAGE) \
+		bash -ec 'apt-get update && apt-get install -y --no-install-recommends unzip && make test'
```
mpryc, line 203: do we need to run this on every test run? May be slowing down quite a bit tests.

Makefile
```diff
@@ -407,9 +420,15 @@ bundle-push: ## Push the bundle image.
...
-PROTOC_ARCH = $(shell go env GOARCH | sed -e 's/amd64/x86_64/' -e 's/arm64/aarch_64/')
+PROTOC_OS = $(shell go env GOHOSTOS | sed -e 's/darwin/osx/')
+PROTOC_ARCH = $(shell go env GOHOSTARCH | sed -e 's/amd64/x86_64/' -e 's/arm64/aarch_64/')
+ifeq ($(PROTOC_OS),osx)
+# Protobuf 3.16.0 only provides an x86_64 macOS binary (requires Rosetta on Apple Silicon).
+PROTOC_ARCH = x86_64
+endif
 protoc: protoc-gen-go protoc-gen-go-grpc ## Download protoc (protocol buffers tool needed for gRPC)
-	test -f ${PROTOC} || (cd $(shell pwd)/bin/proto && curl -sSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-linux-$(PROTOC_ARCH).zip && unzip -o protoc.zip && rm protoc.zip)
+	$(PROTOC) --version >/dev/null 2>&1 || (cd $(shell pwd)/bin/proto && curl -fsSLo protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protoc-3.16.0-$(PROTOC_OS)-$(PROTOC_ARCH).zip && unzip -o protoc.zip && rm protoc.zip)
```
mpryc, line 430: Not for this PR, but would consider changing as follow up to a much never version of protobuf which ships native aarch_64 macOS binaries - this would eliminate Rosetta work-around from this pr?

Makefile
pranavgaikwad: @mpryc I don't know if I understand, this is a separate target that folks are supposed to run locally to run the unit tests on a Mac host. This is running in an ephemeral container

Makefile
pranavgaikwad: @mpryc sounds good

Makefile
```diff
@@ -189,6 +189,19 @@ test: go-verify envtest generate fix-imports manifests fmt vet ## Run tests.
 		KUBEBUILDER_CONTROLPLANE_STOP_TIMEOUT="60s"\
 		go test ./api/... ./internal/... -coverprofile cover.out -v ${TEST_OPS}
 
+CONTAINER_TOOL ?= podman
+TEST_LINUX_IMAGE ?= docker.io/library/golang:$(shell go list -m -f '{{.GoVersion}}')-bookworm
+
+.PHONY: test-linux
+test-linux: ## Run the full test target in a Linux container (for macOS hosts).
```
mshitrit, line 196: Nice PR !

Nit: maybe update the name/description of `check` [job](https://github.com/medik8s/self-node-remediation/pull/349/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L526) 
ATM the description is a bit confusing - basically saying it does what `test-linux` is doing while in fact it already assumes a running container.

I think there is a good chance it can be removed all together (even better if so 😄 )

mshitrit (approved):
/hold

left a non blocking nit comment. 
Feel free to update or ignore

PR #354, opened 2026-09-18
Add PeerTopologyKey: failure-domain-aware peer evaluation

maxnmr:
#### Why we need this PR

When a whole failure domain (a site, a zone, a rack) is partitioned from the rest of the cluster, the SNR agents inside that domain keep each other alive. Each of them answers `ApiError` ("I can't reach the API server either") to the others, and today that answer counts as proof that the node is not isolated: it resets the `MaxTimeForNoPeersResponse` timer and feeds the "more than 50% of peers can't reach the API → control plane failure → healthy" rule.

As a result, **no node of a partitioned domain ever self-remediates as long as it has at least one peer in the same domain**, while NHC on the majority side fences those nodes by time (`safeTimeToAssumeNodeRebootedSeconds`), force-deletes their pods and VolumeAttachments, and reschedules workloads that are in fact still running on the isolated side.

Reproduced on OpenShift 4.22.12 with SNR 0.13.1 (OLM, Red Hat operator catalog), two sites of two workers each plus a third site holding only a control plane node. Agent log of a worker of the isolated site, every 25 s for the whole duration of the partition:

```
Aggregate peer health responses  {"healthyResponses": 0, "unhealthyResponses": 0, "apiErrorsResponses": 1}
Ignoring no peers response error, time is below threshold for no peers response
  {"time without peers response (seconds)": 0.00002, "threshold (seconds)": 30}
peers did not confirm that we are unhealthy, ignoring error
```

The single `apiErrorsResponses` is the other worker of the same site. The agent cannot distinguish "my domain is partitioned" from "the control plane is down" with this information: a peer that shares its network fate cannot tell it anything about the rest of the cluster.

#### Changes made

New optional field `SelfNodeRemediationConfig.Spec.PeerTopologyKey` (`PEER_TOPOLOGY_KEY` env on the agent): the name of a node label identifying the failure domain, e.g. `topology.kubernetes.io/zone`.

When set, an `ApiError` answer coming from a peer **located in the same failure domain as the node** is not taken as evidence:

- it does not reset `timeOfLastPeerResponse`,
- it does not count toward the "control plane failure" majority, whose denominator becomes the number of peers **outside** the node's domain.

`Healthy` and `Unhealthy` answers are never filtered, whatever the domain of the peer: a same-domain peer that can see the API, or that sees a SelfNodeRemediation CR for us, still carries real information.

Backward compatibility — the behavior is strictly unchanged when:

- the field is empty (default): none of the new code paths is entered,
- the node does not carry the label,
- no peer exists outside of the node's own domain (single-domain clusters): the legacy logic is used as is.

A peer whose node lacks the label is treated as being outside of the node's domain, so a partially labeled cluster degrades toward the current behavior, never toward more reboots.

The value is validated by the existing `SelfNodeRemediationConfig` webhook (`validation.IsQualifiedName`), since it is rendered into the agent DaemonSet and used to read node labels.

Each evaluation round works on a single snapshot of the peers (addresses, per-peer domain classification, number of peers outside the domain), taken under one lock, so that a peer refresh happening in the middle of a round cannot make the majority denominator and the answer classification disagree.

Files:

- `api/v1alpha1/selfnoderemediationconfig_types.go`: new optional field, CRD regenerated
- `internal/controller/selfnoderemediationconfig_controller.go`, `install/self-node-remediation-deamonset.yaml`, `cmd/main.go`: plumbing to the agent env
- `internal/peers/peers.go`: keeps the failure domain of each peer next to its IP; `GetPeersSnapshot` returns a consistent view of the peers of a role
- `internal/apicheck/check.go`: peer answers are tagged with their origin; same-domain `ApiError` ignored when the feature is active; `PeersProvider` interface so the decision logic can be unit tested with fake peers
- `internal/webhook/v1alpha1/selfnoderemediationconfig_webhook.go`: `peerTopologyKey` must be a valid label key
- `internal/apicheck/check_test.go`, `internal/peers/peers_test.go`, `internal/webhook/v1alpha1/selfnoderemediationconfig_webhook_test.go`: 11 + 6 new specs and 5 new test cases
- `bundle/manifests/...selfnoderemediationconfigs.yaml`: bundle CRD regenerated (`make bundle`)

No existing issue, the problem is described above.

#### Test plan

`make test-imports`, `make verify-vendor`, `make verify-bundle` and `make test` (unit tests with envtest, formatting, code generation) all pass locally.

New unit tests cover: feature disabled, single-domain cluster (fallback), partitioned domain (same-domain `ApiError` ignored → isolated), control plane outage (other-domain `ApiError` counted → majority), same-domain `Healthy`/`Unhealthy` still trusted, unlabeled peer treated as outside the domain, no-peers-response timer not reset by a same-domain `ApiError`, domains surviving a control plane peers refresh, webhook accepting/rejecting the key.

Manual E2E on OpenShift 4.22.12 (SNR built from this branch, deployed through the 0.13.1 OLM CSV with the image and the CRD swapped), 3 control planes and 4 workers spread over two sites (2 workers each) plus an arbiter site, i6300ESB watchdog device with a 30 s timeout, `isSoftwareRebootEnabled: false`, NHC `minHealthy: 50%` / `duration: 300s`. Site isolation done with `tc` filters on the NIC so that the pods of the isolated site can only reach each other (an `nftables` filter on the host does **not** cut pod traffic, which is switched by OVS).

| Test | `peerTopologyKey` | Scenario | Result |
|---|---|---|---|
| 1 | unset | single worker, NIC down | reboots in 1 min 30 — unchanged |
| 2 | unset | one site isolated | **no reboot**, NHC fences the still-running nodes a few minutes later (`duration` + `safeTimeToAssumeNodeRebootedSeconds`) — the bug |
| 3 | set | one site isolated, API of the site's own control plane node unusable | **both workers of the site reboot within ~1 min**, before NHC acts |
| 4 | set | API servers unreachable from all workers, network otherwise fine | **no reboot**, `HealthyBecauseMostPeersCantAccessAPIServer` on every cycle |

Test 3, agent of a worker of the isolated site:

```
Error count exceeds threshold, trying to ask other nodes if I'm healthy
Failure domain awareness is active, api errors from peers of the same domain will be ignored {"peersOutsideMyDomain": 2, "allPeers": 3}
got response from peer {"IP": "<other worker of the same site>", "status": 3}
Ignoring api errors reported by peers of my own failure domain {"ignoredApiErrorsResponses": 1}
Aggregate peer health responses {"healthyResponses": 0, "unhealthyResponses": 0, "apiErrorsResponses": 0}
Failed to get health status peers. Assuming unhealthy {"reason": "UnHealthyBecauseNodeIsIsolated"}
we are unhealthy, triggering a reboot
watchdog feeding has stopped, waiting for reboot to commence
```

Test 4, same agent, all workers cut from the API servers:

```
Failure domain awareness is active, api errors from peers of the same domain will be ignored {"peersOutsideMyDomain": 2, "allPeers": 3}
Ignoring api errors reported by peers of my own failure domain {"ignoredApiErrorsResponses": 1}
Aggregate peer health responses {"healthyResponses": 0, "unhealthyResponses": 0, "apiErrorsResponses": 2}
More than 50% of the nodes couldn't access the api-server, assuming this is a control plane failure
```

Related observation, out of scope here (and the reason why test 3 also cuts the API of the site's own control plane node): without that, the agent's `/readyz` check may never fail for the whole partition, because its HTTP connection to the `kubernetes` service had landed on the API server of the isolated site, which keeps answering 200 on `/readyz` without etcd quorum. The peer logic is only entered after a failed API check, so the agent never evaluated anything. The control plane code path already handles this false positive for control plane nodes; workers have no equivalent. Worth a separate issue.


maxnmr:
Thanks for the review. Both points addressed in the updated commit:

- **Consistent snapshot**: each evaluation round now takes a single `peers.Snapshot` (addresses, per-peer domain classification, number of peers outside the domain) under one lock, and reuses it for the whole round. The two separate reads are gone.
- **Validation and rendering**: `peerTopologyKey` is validated by the existing `SelfNodeRemediationConfig` webhook with `validation.IsQualifiedName` on create and update, which rules out anything that could break the DaemonSet template. Covered by webhook tests, and checked on a live cluster (an invalid value is rejected by the admission webhook).

Also added doc comments on the touched functions.

Re-ran the unit suite (`make test`) and the manual E2E scenarios of the test plan with the updated build: same results.

mshitrit:
Hi @maxnmr , thanks for contributing !

Before diving into the implementation details, I have a couple of design-level questions I'd like to discuss:

1. Does remediation help a node in a partitioned zone?

If I understand correctly, the underlying assumption is that a node which detects it's in an isolated failure domain would benefit from self-remediation (reboot). I'm not sure that holds for the primary use case this feature targets.

When an entire zone is partitioned from the API server and other zones, the root cause is typically infrastructure-level (switch failure, BGP misconfiguration, router issue) — not something on the node itself. If the node reboots, it comes back up into the same partitioned network: kubelet still can't reach the API server, the SNR agent re-detects isolation, and we're back where we started — except now the workloads on that node went through an unnecessary disruption.

This is different from the normal SNR flow, where NHC detects a single node as NotReady and creates an SNR CR. There, the root cause is usually node-local (hung kubelet, stuck runtime, kernel deadlock), and rebooting actually resolves it. Zone-level partitions are, by definition, not node-local problems.

2. Blast radius of zone-wide self-remediation

Even if we agree there are cases where rebooting helps, the current implementation would cause all nodes in the partitioned zone to self-remediate roughly simultaneously. Each node independently reaches the same conclusion ("I'm isolated → reboot"), with no cross-node coordination or throttling.

This is concerning because:
- A transient partition (e.g., a brief router flap) would trigger mass unnecessary reboots across the entire zone
- Stateful workloads (PVs, databases) get disrupted simultaneously on every node in the zone
- The thundering herd of nodes restarting at once could destabilize the cluster further when they rejoin
- There's no topology-aware MaxUnhealthy equivalent — NHC's maxUnhealthy governs SNR CR creation, but here the nodes bypass that and self-remediate directly

If we agree that  use case 1 is valid, we could mitigate this with something like a configurable cap on simultaneous zone-level remediations 

I'd love to hear your thoughts on these before we go deeper into the code. Thanks!

maxnmr:
Thanks for looking at this — both questions are fair, let me take them in turn.

### 1. The reboot is fencing, not repair

I agree that rebooting doesn't fix a switch, and that the node comes back into the same partition. But repairing the node isn't the goal — making the healthy side's assumption true is. From the API doc of `safeTimeToAssumeNodeRebootedSeconds`:

> the healthy self node remediation agents will assume the unhealthy node has been rebooted, and it is safe to recover affected workloads. This is extremely important as starting replacement Pods while they are still running on the failed node will likely lead to data corruption and violation of run-once semantics.

In a partitioned domain holding >= 2 workers, that assumption is silently false today, and nothing else in the sequence changes: NHC still creates the SNR CRs, the healthy agents still time out, apply the out-of-service taint, force-delete the pods and VolumeAttachments, and the replacements still start — while the original pods keep running on the isolated side, for as long as the partition lasts. So the workload disruption you're describing happens either way; the only variable is whether run-once semantics hold while it happens. That's the guarantee SNR exists to provide, and it is the one thing the majority side cannot enforce on its own — it can only assume.

It's also worth noting that SNR already reboots nodes for causes that aren't node-local: a single isolated worker (switch port down, NIC, cable) goes through `UnHealthyBecauseNodeIsIsolated` and reboots today. Whether a node self-fences currently depends on how many workers happen to sit behind the same switch — one per domain and it reboots, two and nobody does, with the same root cause and the same consequences for the workloads. This PR removes that inconsistency rather than adding a new class of remediation.

### 2. Blast radius

**The node does not reboot-loop.** I tested this explicitly for this discussion: a single worker cut at the infrastructure level (port shut, so the partition survives a reboot), SNR/NHC from the released operator, 20 minutes.

| | |
|---|---|
| T+0 | port shut |
| T+1m29 | watchdog reset |
| T+1m39 | node back up, kubelet logging `Unable to register node with API server` |
| T+5m35 | the only container started while isolated: the *static* pod `kube-rbac-proxy-crio` |
| T+1m39 -> T+20m | **no second reset** — a single boot in `journalctl --list-boots` |
| T+20m | port re-enabled; node Ready 19 s later, no taint, no leftover SNR CR; the agent DaemonSet starts and opens `/dev/watchdog` |

Without API access kubelet can only start static pods; the agent is a DaemonSet, so nothing re-arms the watchdog. The node stays up and idle — precisely the fenced state — and rejoins cleanly when the partition heals.

**The trigger is narrower than "a zone-wide event".** The path is only entered after the node has lost the API for `maxApiErrorThreshold * apiCheckInterval` (45 s by default) **and** received no answer at all from any peer outside its own domain for `MaxTimeForNoPeersResponse` (30 s) — about 1 min 45 end to end in my tests. A brief router flap doesn't reach that. And a domain that still has a working control plane never enters the path, because its API checks succeed: this only fires on the side that lost the control plane, i.e. exactly the side whose workloads the other side is taking over. A full control-plane outage with the network intact is still classified healthy (the ">50% of peers can't reach the API server" rule is kept, with out-of-domain peers as the denominator — test 4 in the description).

**On a cap.** My concern with "at most N nodes per domain at a time" is that a partially fenced domain is worse than either extreme: the healthy side assumes **every** fenced node rebooted after `safeTimeToAssumeNodeRebootedSeconds`, so the nodes deliberately held back are exactly the run-once violations we're trying to prevent. What I could add without breaking that contract is a bounded random delay before the watchdog is stopped (configurable, default 0), so reboots inside a domain spread out but all still complete before `safeTimeToAssumeNodeRebootedSeconds`. Happy to implement that — or the cap, if you prefer it after all.

The feature also stays opt-in: with `peerTopologyKey` unset, which is the default, none of the new code paths is entered.

Happy to move this to an issue if you'd rather have the design discussion there.

slintes:
> But repairing the node isn't the goal — making the healthy side's assumption true is.

This can't be repeated often enough...

I have a question though: the peer check was introduced in order to prevent a mass reboot when the API is down. This PR fixes the problem to assume that the API is down when the affected nodes are in a separate and isolated failure domain.
What happens though when the API endpoint is in the same failure domain and really is down? We need to prevent a mass reboot in this domain in this case.

maxnmr:
Good catch, you're right — and I could reproduce it.

With the current commit, a failure domain that holds the control plane and loses the API server together with its links to the other domains ignores the api errors of its own peers, sees nobody outside, and reboots. Yet in that situation nobody can take over: an API server needs etcd quorum, and the members are on the partitioned side. The reboot protects nothing, it is exactly the mass reboot the peer check is there to prevent.

### Fix (new commit)

Same-domain api errors are now only ignored when a **strict majority of the control plane nodes is known to be outside of the node's failure domain**. Only then can an API server with quorum run on the other side of the partition and recover the workloads, so only then is self-fencing the right call. Otherwise the legacy evaluation applies unchanged.

- the distribution is built from the control plane **nodes** and their topology label, not from the agent pods, so it stays correct when the agent DaemonSet does not run on control plane nodes;
- control plane nodes without the label are not counted as outside, so a partially labeled cluster still degrades toward the legacy behavior, never toward more reboots;
- an agent running on a control plane node counts its own node in its domain;
- an even split counts as "no majority outside".

In the usual stretched layout, with one control plane node per zone (two data centers plus an arbiter site), every worker domain holds a minority of the control plane: the check never kicks in and the behavior is exactly the one of the first commit.

### Results

3 control plane nodes + 4 workers. For your scenario I used a test label as `peerTopologyKey` to put the 3 control plane nodes and 2 workers in domain A, and the 2 other workers in domain B; then, at the same second on the 4 workers, I cut the API server (control plane nodes and API VIP, port 6443) and the traffic between A and B, both ways — A's workers still see each other. Same cluster, same procedure, only the image changes.

| Scenario | Current commit | With the guard |
|---|---|---|
| Domain A holds the whole control plane, loses the API server and is cut from B (your case) | **A's workers reboot** (`Failure domain awareness is active` → `Ignoring api errors reported by peers of my own failure domain` → `UnHealthyBecauseNodeIsIsolated`) | **no reboot in A over 8 minutes**: every round logs `Failure domain awareness is not applied: my failure domain holds at least half of the control plane nodes`, the api error of the same-domain peer is counted again → `HealthyBecauseNoPeersResponseNotReachedTimeout` |
| A site holding 1 of the 3 control plane nodes is isolated, with the zone label (the case this PR is for) | reboot | **reboot** of both workers of the site after 1 min 22 (`active … "controlPlaneNodesInOtherDomains": 2`) |
| API server unreachable from all workers, network otherwise fine | no reboot | **no reboot**, `HealthyBecauseMostPeersCantAccessAPIServer` on every round, all 4 workers |

### What remains

In the first scenario, the two workers of domain B, which hold no control plane node, still reboot. From their side, "the partition left a working control plane on the other side" and "the other side lost its control plane too" look exactly the same; with the control plane majority outside, the correct default is to assume that someone takes over and to fence. So the remaining case is a double failure — the domain holding the control plane majority loses its API server *and* gets partitioned from a minority domain at the same time — where the minority domain reboots without need. No run-once violation there, only unnecessary disruption, and I don't see how a node could tell the two apart without seeing across the partition. I can add a sentence about it to the field documentation.

Unit tests: 6 new specs for the evaluation plus table tests for the counting; `make test`, `test-imports`, `verify-vendor` and `verify-bundle` pass. I pushed the change as a separate commit to keep it easy to review, and will squash once the approach is agreed.

internal/peers/peers.go
```diff
@@ -54,9 +75,17 @@ func New(myNodeName string, peerUpdateInterval time.Duration, reader client.Read
 		apiServerTimeout:           apiServerTimeout,
 		workerPeersAddresses:       []v1.PodIP{},
 		controlPlanePeersAddresses: []v1.PodIP{},
+		topologyDomains:            map[string]string{},
```
mshitrit, line 80: IIUC this should be protected somehow (probably a mutex lock)  as it might be concurrently updated

maxnmr: Thanks — good call to look at it. The map itself is fine: `topologyDomains` and `controlPlaneNodeDomains` are only written from `updatePeers`, which holds `p.mutex` for its whole body, and only read in `GetPeersSnapshot` / `GetControlPlaneDomains`, which take it as well.

But the neighbouring field was not: `myTopologyDomain` was assigned in `Start` without the lock, while the API connectivity check reads it under the lock from another goroutine. In practice the first read happens long after `Start`, but that's timing, not synchronization.

Fixed in a040ffe8: the assignment now takes the mutex, the struct documents which fields it guards, and a new test (`TestTopologyStateIsGuardedByTheMutex`) reads the failure domain state concurrently with `Start`. Without the fix, `go test -race` reports the race (write in `Start`, read in `GetPeersSnapshot`); with it, 20 runs under `-race` are clean, as are the `peers` and `apicheck` packages.

PR #355, opened 2026-09-18
update the github workflow for SNR

weshayutin:
essentially the same workflow as NHCO but this one of couse installs SNR from src and pulls NHCO


mshitrit:
ATM we need to figure out how to work around SNR - Kind limitation.

@pranavgaikwad  put it quite well:
>So the core issue is that in kind we share the kernel's softdog with all the nodes because they run on same host. In SBR this is a blocker because preflight checks test watchdog device and workers get "Device busy" errors since everybody is kind of opening the saem device even though handles are different.

Ongoing discussion over [here](https://redhat-internal.slack.com/archives/C03M5GKJNBA/p1790084703189469)

internal/reboot/rebooter.go
```diff
@@ -68,15 +84,25 @@ func (r *watchdogRebooter) Reboot() error {
...
 func (r *watchdogRebooter) softwareReboot() error {
 	r.log.Info("about to try software reboot")
-	// privileged:true required to run this
-	rebootCmd := exec.Command("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/bash", "-c", "echo b > /proc/sysrq-trigger")
-
-	if err := rebootCmd.Run(); err != nil {
-		r.log.Error(err, "failed to run reboot command")
-		// TODO retry because of this?
+	cmds := [][]string{
+		{"systemctl", "reboot", "--force", "--force"},
```
mpryc: I see it's similar to https://github.com/mpryc/storage-based-remediation/blob/94ef27af1909754f8ffeb06d5dde83cd2dd5c7f6/cmd/sbr-agent/main.go#L2297 


But the way they are being invoked it's slightly different, shoud we consider moving this to medik8s/common and use it from there? Can be separate PR, so we don't block the SNR CI.

.github/workflows/release.yaml
razo7: Any reason for adding this file?
It looks like a duplicate of https://github.com/medik8s/self-node-remediation/blob/main/.github/workflows/release.yml 🤔

mpryc (approved):

jmontleon (approved):
/lgtm

